PR #11, opened 2020-03-11
:bug: Fix clusterctl-settings.json file

kashifest:
This small PR updates clusterctl-settings.json with correct name of the provider i.e. metal3 and also corrects the version of release.

### DIFF
--- a/clusterctl-settings.json
+++ b/clusterctl-settings.json
@@ -1,7 +1,7 @@
 {
-  "name": "infrastructure-baremetal",
+  "name": "infrastructure-metal3",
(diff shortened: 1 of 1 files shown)


maelk:
@kashifest This went too fast in. The nextVersion is incorrect, it should be v0.4.0

kashifest:
@maelk  Fixing it, actually nextVersion can be anything , so it is not incorrect or v0.4.0 is correct in that sense, but yes it should not make sense to give the release version as the next version. A PR is coming to change it.

PR #14, opened 2020-03-12
:bug: Modify deploy-bmo-cr in Makefile

kashifest:
`deploy-bmo-cr` in Makefile deploys bmo CRDs and CRs from examples/_out, while it is not generated before `make generate-examples` or `make deploy`. This PR modifies the `deploy-bmo-cr` to deploy the bmo CRD and CR from examples directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,8 +280,8 @@ install:
 
 #Deploy the BaremetalHost CRDs and CRs (for testing purposes only)
 deploy-bmo-cr:
-	kubectl apply -f ./examples/_out/metal3crds.yaml
-	kubectl apply -f ./examples/_out/metal3plane.yaml
+	kubectl apply -f ./examples/metal3crds/metal3.io_baremetalhosts.yaml
+	kubectl apply -f ./examples/metal3plane/hosts.yaml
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,14 @@ The architecture with the components involved is documented [here](docs/architec
 
 ## Deployment and examples
 
+### Generate example configurations
+
+Generates example configurations (CRs and provider components)
+
+```sh
+    make generate-examples
```
maelk: do we need this here ? it is automatically triggered as part of `make deploy` or `make install`

kashifest: If I follow the instructions as it is now from the current flow in README, I hit a bump. because the first instruction is, `make deploy-bmo-cr`. And the CR is not there. Then I realized , i didn't generate the examples. For someone who is new, he would not know such a phony exists and if he simple starts from README, since it says deployment and examples in the heading, he will hit the issue too. In my opinion, this should be visible here.

maelk: hum, the problem is rather that the generate-examples dependency was removed from the deploy-bmo-cr target. I don't recall why. We probably should fix it there so that it would work out of the box, or deploy directly from the examples folder, not the _out folder

kashifest: True. I have updated the PR

PR #20, opened 2020-03-16
:bug: Remove make install from Makefile

kashifest:
Current way of deploying CAPM3 locally is `make run` which in turn calls `make install` and applies the CRDs only. This is no longer sufficient with webhooks picture.  So the current way of deploying CAPM3 has bugs. One solution is to remove `install` phony all-together from Makefile because this is not needed anymore. This PR removes it which fixes the bug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -271,12 +271,8 @@ set-manifest-pull-policy:
 ## --------------------------------------
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
-run: generate fmt vet install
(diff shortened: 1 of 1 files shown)


maelk:
looks good, but you also need to update the readme probably

kashifest:
> looks good, but you also need to update the readme probably

Can you pinpoint? I didn't find any mention of  `install` anywhere else

maelk:
you're right, my bad, sorry. 
/lgtm

PR #24, opened 2020-03-19
:sparkles: Set/Unset PauseAnnotation in BMH through CAPM3 machine + Tests

kashifest:
This PR adds the functionality to set/unset pause annotation in BMH through CAPM3 machine reconcilation. The test codes are also added.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -53,11 +53,12 @@ const (
 	ProviderName = "baremetal"
 	// HostAnnotation is the key for an annotation that should go on a Machine to
 	// reference what BareMetalHost it corresponds to.
-	HostAnnotation     = "metal3.io/BareMetalHost"
-	requeueAfter       = time.Second * 30
-	bmRoleControlPlane = "control-plane"
-	bmRoleNode         = "node"
-	userDataFinalizer  = "metal3machine.infrastructure.cluster.x-k8s.io/userData"
+	HostAnnotation      = "metal3.io/BareMetalHost"
+	requeueAfter        = time.Second * 30
+	bmRoleControlPlane  = "control-plane"
+	bmRoleNode          = "node"
+	userDataFinalizer   = "metal3machine.infrastructure.cluster.x-k8s.io/userData"
+	pausedAnnotationKey = "metal3.io/capm3"
 )
 
 // MachineManagerInterface is an interface for a ClusterManager
@@ -73,6 +74,8 @@ type MachineManagerInterface interface {
 	HasAnnotation() bool
 	SetNodeProviderID(context.Context, string, string, ClientGetter) error
 	SetProviderID(string)
+	SetPauseAnnotation(context.Context) error
+	RemovePauseAnnotation(context.Context) error
 }
 
 // MachineManager is responsible for performing machine reconciliation
@@ -148,6 +151,68 @@ func (m *MachineManager) role() string {
 		return bmRoleControlPlane
 	}
 	return bmRoleNode
+}
+
+// RemovePauseAnnotation checks and/or Removes the pause annotations on associated bmh
+func (m *MachineManager) RemovePauseAnnotation(ctx context.Context) error {
+	// look for associated BMH
+	host, err := m.getHost(ctx)
+	if err != nil {
+		m.setError("Failed to get a BaremetalHost for the Metal3Machine",
+			capierrors.CreateMachineError,
+		)
+		return err
+	}
+
+	if host == nil {
+		return nil
+	}
+
+	annotations := host.GetAnnotations()
+
+	if annotations != nil {
+		if _, ok := annotations[bmh.PausedAnnotation]; ok {
+			if m.Cluster.Name == host.Labels[capi.ClusterLabelName] && annotations[bmh.PausedAnnotation] == pausedAnnotationKey {
+				// Removing BMH Paused Annotation Since Owner Cluster is not paused
+				delete(host.Annotations, bmh.PausedAnnotation)
+			} else if m.Cluster.Name == host.Labels[capi.ClusterLabelName] && annotations[bmh.PausedAnnotation] != pausedAnnotationKey {
+				m.Log.Info("BMH is paused by user. Not removing Pause Annotation")
+				return nil
+			}
+		}
+	}
+	return m.updateObject(ctx, host)
+}
+
+// SetPauseAnnotation sets the pause annotations on associated bmh
+func (m *MachineManager) SetPauseAnnotation(ctx context.Context) error {
+	// look for associated BMH
+	host, err := m.getHost(ctx)
+	if err != nil {
+		m.setError("Failed to get a BaremetalHost for the Metal3Machine",
+			capierrors.CreateMachineError,
+		)
+		return err
+	}
+	if host == nil {
+		return nil
+	}
+
+	annotations := host.GetAnnotations()
+
+	if annotations != nil {
+		if _, ok := annotations[bmh.PausedAnnotation]; ok {
+			m.Log.Info("BaremetalHost is already paused")
+			return nil
+		} else {
+			m.Log.Info("Adding PausedAnnotation in BareMetalHost")
+			host.Annotations[bmh.PausedAnnotation] = pausedAnnotationKey
+		}
+	} else {
+		host.Annotations = make(map[string]string)
+		host.Annotations[bmh.PausedAnnotation] = pausedAnnotationKey
+	}
+	return m.updateObject(ctx, host)
 }
 
 // GetBaremetalHostID return the provider identifier for this machine
@@ -464,6 +529,11 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 
 		if host.Labels != nil && host.Labels[capi.ClusterLabelName] == m.Machine.Spec.ClusterName {
 			delete(host.Labels, capi.ClusterLabelName)
+		}
+
+		m.Log.Info("Removing Paused Annotation (if any)")
+		if host.Annotations != nil && host.Annotations[bmh.PausedAnnotation] == pausedAnnotationKey {
+			delete(host.Annotations, bmh.PausedAnnotation)
 		}
 
 		// Delete created secret, if data was set without DataSecretName or if
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -163,9 +163,54 @@ func bmmObjectMetaWithValidAnnotations() *metav1.ObjectMeta {
 		Name:            "mybmmachine",
 		Namespace:       "myns",
 		OwnerReferences: []metav1.OwnerReference{},
+		Labels: map[string]string{
+			capi.ClusterLabelName: clusterName,
+		},
 		Annotations: map[string]string{
 			HostAnnotation: "myns/myhost",
 		},
+	}
+}
+
+func bmhObjectMetaWithValidCAPM3PausedAnnotations() *metav1.ObjectMeta {
+	return &metav1.ObjectMeta{
+		Name:            "myhost",
+		Namespace:       "myns",
+		OwnerReferences: []metav1.OwnerReference{},
+		Labels: map[string]string{
+			capi.ClusterLabelName: clusterName,
+		},
+		Annotations: map[string]string{
+			bmh.PausedAnnotation: pausedAnnotationKey,
+		},
+	}
+}
+
+func bmhObjectMetaWithValidEmptyPausedAnnotations() *metav1.ObjectMeta {
+	return &metav1.ObjectMeta{
+		Name:            "myhost",
+		Namespace:       "myns",
+		OwnerReferences: []metav1.OwnerReference{},
+		Annotations: map[string]string{
+			bmh.PausedAnnotation: "",
+		},
+	}
+}
+
+func bmhObjectMetaEmptyAnnotations() *metav1.ObjectMeta {
+	return &metav1.ObjectMeta{
+		Name:            "myhost",
+		Namespace:       "myns",
+		OwnerReferences: []metav1.OwnerReference{},
+		Annotations:     map[string]string{},
+	}
+}
+
+func bmhObjectMetaNoAnnotations() *metav1.ObjectMeta {
+	return &metav1.ObjectMeta{
+		Name:            "myhost",
+		Namespace:       "myns",
+		OwnerReferences: []metav1.OwnerReference{},
 	}
 }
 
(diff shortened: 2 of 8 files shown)


kashifest:
Holding the PR to give some time for review 
/hold

controllers/metal3machine_controller_test.go
```diff
@@ -439,4 +440,56 @@ var _ = Describe("Metal3Machine manager", func() {
...
+			r := Metal3MachineReconciler{
+				Client: c,
+			}
+			obj := handler.MapObject{
+				Object: tc.Cluster,
+			}
+			reqs := r.ClusterToMetal3Machines(obj)
+
+			if tc.ExpectRequest {
+				Expect(len(reqs)).To(Equal(1), "Expected 1 request, found %d", len(reqs))
```
maelk: you should add a test case with multiple machines

controllers/metal3machine_controller.go
```diff
@@ -248,6 +261,34 @@ func (r *Metal3MachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	capiMachineList := &capi.MachineList{}
+	if err := r.Client.List(context.TODO(), capiMachineList, client.InNamespace(c.Namespace), client.MatchingLabels(labels)); err != nil {
+		r.Log.Error(err, "failed to list Metal3Machines")
+		return nil
+	}
+	for _, m := range capiMachineList.Items {
+		if m.Spec.InfrastructureRef.Name == "" {
+			continue
+		}
+		name := client.ObjectKey{Namespace: m.Namespace, Name: m.Name}
```
maelk: the namespace should come from the InfrastructureRef spec if provided

controllers/metal3machine_controller.go
```diff
@@ -151,6 +145,19 @@ func (r *Metal3MachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+		// set pause annotation on associated bmh (if any)
+		err := machineMgr.SetPauseAnnotation(ctx)
+		if err != nil {
+			machineLog.Info("failed to set pause annotation on associated bmh")
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil
+	}
+
```
maelk: you're missing the logic to remove the Pause annotation if the cluster has been unpaused

controllers/metal3machine_controller.go
```diff
@@ -151,6 +145,19 @@ func (r *Metal3MachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 		return ctrl.Result{}, errors.Wrapf(err, "failed to create helper for managing the machineMgr")
 	}
 
+	// Return early if the BMMachine or Cluster is paused.
+	if util.IsPaused(cluster, capm3Machine) {
```
maelk: this means that the BMH will be paused also if the pause annotation is set on the m3machine. Is that desired behaviour ?

baremetal/metal3machine_manager_test.go
```diff
@@ -563,6 +594,96 @@ var _ = Describe("Metal3Machine manager", func() {
...
+					},
+				},
+			},
+			BMMachine: newMetal3Machine("mybmmachine", nil, bmmSpec(), nil,
+				bmmObjectMetaWithValidAnnotations()),
+			ExpectPresent: true,
+			ExpectError:   false,
+		}),
+	)
+
```
maelk: you're missing the test cases to remove the pause annotation

baremetal/metal3machine_manager_test.go
```diff
@@ -563,6 +594,96 @@ var _ = Describe("Metal3Machine manager", func() {
...
+					Namespace: tc.Host.Namespace,
+				},
+				&savedHost,
+			)
+			Expect(err).NotTo(HaveOccurred())
+			if tc.ExpectPresent {
+				Expect(savedHost.Annotations[bmh.PausedAnnotation]).NotTo(BeNil())
+			} else {
+				Expect(savedHost.Annotations[bmh.PausedAnnotation]).To(BeNil())
+				return
```
maelk: this return is unneeded (it's the end of the function anyways)

baremetal/metal3machine_manager.go
```diff
@@ -149,6 +151,43 @@ func (m *MachineManager) role() string {
...
+		m.Log.Info("BaremetalHost not associated")
+		return nil
+	}
+
+	annotations := host.GetAnnotations()
+
+	if annotations != nil {
+		if _, ok := annotations[bmh.PausedAnnotation]; ok {
+			if host.Annotations[bmh.PausedAnnotation] != pausedAnnotationKey {
+				m.Log.Info("BaremetalHost is already paused by user")
```
maelk: If you don't set it yourself, what if the user removes its pause annotation ? What will be the flow ?

baremetal/metal3machine_manager.go
```diff
@@ -149,6 +151,43 @@ func (m *MachineManager) role() string {
...
+			} else {
+				m.Log.Info("BaremetalHost is already paused by capM3")
```
maelk: maybe this does not need to be logged, because it will jam the logs, every 30 sec. for all bmhs...

baremetal/metal3machine_manager.go
maelk: maybe this does not need to be logged, because it will jam the logs, every 30 sec. for all bmhs...

baremetal/metal3machine_manager.go
```diff
@@ -149,6 +151,43 @@ func (m *MachineManager) role() string {
...
+	// look for associated BMH
+	host, err := m.getHost(ctx)
+	if err != nil {
+		m.setError("Failed to get a BaremetalHost for the Metal3Machine",
+			capierrors.CreateMachineError,
+		)
+		return err
+	}
+	if host == nil {
+		m.Log.Info("BaremetalHost not associated")
```
maelk: maybe this does not need to be logged, because it will jam the logs, every 30 sec. for all bmhs...

baremetal/metal3machine_manager.go
```diff
@@ -149,6 +151,43 @@ func (m *MachineManager) role() string {
...
+			if host.Annotations[bmh.PausedAnnotation] != pausedAnnotationKey {
+				m.Log.Info("BaremetalHost is already paused by user")
+			} else {
+				m.Log.Info("BaremetalHost is already paused by capM3")
+			}
+			return nil
+		} else {
+			m.Log.Info("Adding PausedAnnotation in BareMetalHost")
+			host.Annotations[bmh.PausedAnnotation] = pausedAnnotationKey
+			return m.updateObject(ctx, host)
```
maelk: you could take this return and the next one out of the condition, since we return already when the object should not be updated

controllers/metal3machine_controller.go
kashifest: I was also thinking the same. Do we have a logic for that in case of Metal3Machine?

controllers/metal3machine_controller.go
kashifest: I suppose it is? If the CAPM3Machine is paused, shouldn't the associated BMH be? What is your opinion?

baremetal/metal3machine_manager_test.go
kashifest: Do you mean deletion? It is added later ...

baremetal/metal3machine_manager.go
kashifest: A change in the bmh will reconcile CAPM3 machine and then  if the cluster/CAPM3 machine is still paused, the annotation will be added by CAPM3

baremetal/metal3machine_manager.go
kashifest: ok

baremetal/metal3machine_manager.go
kashifest: I would rather keep this one.

baremetal/metal3machine_manager.go
kashifest: ok.

baremetal/metal3machine_manager.go
kashifest: Done

baremetal/metal3machine_manager_test.go
kashifest: Done

controllers/metal3machine_controller_test.go
kashifest: I was following the existing workflow, tests for `BareMetalHost To Metal3Machines` also does not have any test with multiple machines. I can add it anyhow.

baremetal/metal3machine_manager.go
maelk: yes, but in between the BMH will have been reconciled, at least once. That could be very problematic

controllers/metal3machine_controller_test.go
maelk: yes, that would be better

controllers/metal3machine_controller.go
maelk: we don't need it for metal3machine. If the cluster is unpaused, the next reconciliation loop will be completed

baremetal/metal3machine_manager.go
kashifest: Hmmm, if it has to be done through CAPM3, then what else can we do?

controllers/metal3machine_controller.go
maelk: I don't think that m3machine and bmh should be tied on that. you might want to pause the m3machine reconciliation, but keep BMO reconciling the BMH to get the possible state changes for example

baremetal/metal3machine_manager_test.go
maelk: not deletion, no. the removal of the pause annotation when the cluster is unpaused

kashifest: Yeah, but I don't have the logic yet since I was unsure, how can I have tests :D

baremetal/metal3machine_manager.go
```diff
@@ -52,11 +52,12 @@ const (
 	ProviderName = "baremetal"
```
Jaakko-Os: should the provider name reflect the overall naming change?
Meaning:
ProviderName = "metal3"

baremetal/metal3machine_manager.go
maelk: Maybe modify the annotation to add the CAPM3 part to the user input ? so that when the cluster is not paused anymore, you are still able to restore the pause field of the user.

controllers/metal3machine_controller.go
kashifest: Changed the code to remove the tie.

baremetal/metal3machine_manager.go
kashifest: Please check the removal logic and comment if its ok

controllers/metal3machine_controller.go
kashifest: I have added the logic.Please review

controllers/metal3machine_controller.go
kashifest: I am not sure if I understood you correctly. I am following existing logic: https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/controllers/metal3machine_controller.go#L281

maelk: now you are assuming that the machine and Metal3Machine are in the same namespace. if the namespace is given in m.Spec.InfrastructureRef.Namespace, you should use that one, and default to the machine's namespace only if it is not given

kashifest:
> The removal logic is good. There is still a corner case. If the user has set the Pause annotation, you do not change it. The user could then remove it, and you could have some reconciliation in that short time before the metal3machine reconciliation triggered by the change kicks in and sets it again. A solution could be that if there is already a pause annotation, you concatenate it with yours into a single string. And remove yours from that string when done.

I am not sure if this is applicable anymore .... Have you checked the change in the metal3machine reconciliation function  where BMH pause is not dependent on metal3machine pause ... it only depends on the cluster ....

maelk:
That corner case was not related to whether you base that annotation on the metal3machine annotation or not. The problem comes from the fact that if you do not modify the pause annotation value, whatever set the annotation in the first place could remove it without knowing you were relying on it. If you modify it, it is at least visible. But you need a way to put it back as it was when you don't need it anymore.

kashifest:
> The removal logic is good. There is still a corner case. If the user has set the Pause annotation, you do not change it. The user could then remove it, and you could have some reconciliation in that short time before the metal3machine reconciliation triggered by the change kicks in and sets it again. A solution could be that if there is already a pause annotation, you concatenate it with yours into a single string. And remove yours from that string when done.

@maelk here you mentioned there is a short time before  the metal3machine reconciliation triggered by the change kicks in and sets it again, but this is not possible anymore, and moreover, the code does not remove the pause annotation if it set by the user. It only removes the pause annotation if it set by CAPM3 , so "whatever set the annotation in the first place could remove it without knowing you were relying on it." is also  not fully possible. Only , the entity which is setting the annotation is removing its annotation. If the user removes the pause annotation which is set by CAPM3, that is possible and then that is a more general problem in my opinion, since user can actually remove any annotation. If that is the case you are referring, I would rather not complicate the setting/un-setting the annotation with concatenated keys. We could later probably add a test case together to see what problem that might cause.

controllers/metal3machine_controller.go
kashifest: Is it possible to have machine and metal3lmachine in different namespaces ?

maelk:
yes, that's what I was referring to. I am not sure about postponing it. On the one hand, this is a very small corner case, that probably won't happen, but on the other hand, it could be a race condition later on, that would result into a misbehaving node.

controllers/metal3machine_controller_test.go
kashifest: Added

controllers/metal3machine_controller.go
kashifest: Done

baremetal/metal3machine_manager_test.go
kashifest: Added

baremetal/metal3machine_manager.go
kashifest: This needs to be tackled in a separate PR. I will look into it

controllers/metal3machine_controller.go
```diff
@@ -151,6 +145,26 @@ func (r *Metal3MachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+			machineLog.Info("failed to set pause annotation on associated bmh")
+			return ctrl.Result{}, nil
+		}
+	} else {
+		err := machineMgr.CheckPauseAnnotation(ctx)
+		if err != nil {
+			machineLog.Info("failed to check pause annotation on associated bmh")
+			return ctrl.Result{}, nil
+		}
+	}
```
maelk: If your cluster is paused, this will never be executed, due to the return. you need to change the order with L 149

kashifest: True, good catch, changed the code.

baremetal/metal3machine_manager.go
```diff
@@ -53,11 +53,12 @@ const (
...
-	requeueAfter       = time.Second * 30
-	bmRoleControlPlane = "control-plane"
-	bmRoleNode         = "node"
-	userDataFinalizer  = "metal3machine.infrastructure.cluster.x-k8s.io/userData"
+	HostAnnotation      = "metal3.io/BareMetalHost"
+	requeueAfter        = time.Second * 30
+	bmRoleControlPlane  = "control-plane"
+	bmRoleNode          = "node"
+	userDataFinalizer   = "metal3machine.infrastructure.cluster.x-k8s.io/userData"
+	pausedAnnotationKey = "metal3.io/capM3"
```
maelk: nit, maybe capm3, to be consistent with the other places where we use it ?

baremetal/metal3machine_manager.go
```diff
@@ -150,6 +153,68 @@ func (m *MachineManager) role() string {
 	return bmRoleNode
 }
 
+// CheckPauseAnnotation checks and/or Removes the pause annotations on associated bmh
+func (m *MachineManager) CheckPauseAnnotation(ctx context.Context) error {
```
maelk: the naming of the function might be a bit misleading, when its purpose is to remove the annotation conditionally.

baremetal/metal3machine_manager.go
```diff
@@ -150,6 +153,68 @@ func (m *MachineManager) role() string {
...
+			if m.Cluster.Name == host.Labels[capi.ClusterLabelName] && annotations[bmh.PausedAnnotation] == pausedAnnotationKey {
+				// Removing BMH Paused Annotation Since Owner Cluster is not paused
+				delete(host.Annotations, bmh.PausedAnnotation)
+			} else if m.Cluster.Name == host.Labels[capi.ClusterLabelName] && annotations[bmh.PausedAnnotation] != pausedAnnotationKey {
+				m.Log.Info("BMH is paused by user. Not removing Pause Annotation")
+				return nil
+			}
+		}
+	}
+	return m.updateObject(ctx, host)
```
maelk: We should probably handle this kind of function differently. I'll create an issue so that we call the updateObject on the BMH only once at the end of the reconciliation. Otherwise we call it multiple time and make ourselves prone to conflicts because BMO will reconcile in the meantime.

baremetal/metal3machine_manager_test.go
```diff
@@ -563,6 +600,168 @@ var _ = Describe("Metal3Machine manager", func() {
...
+						APIVersion: capm3.GroupVersion.String(),
+					},
+				},
+			},
+			BMMachine: newMetal3Machine("mybmmachine", nil, bmmSpec(), nil,
+				bmmObjectMetaWithValidAnnotations()),
+			ExpectPresent: true,
+			ExpectError:   false,
+		}),
+	)
```
maelk: you should add an entry without annotation, just to make sure nothing crashes if the annotation was removed unexpectedly.

controllers/metal3machine_controller.go
```diff
@@ -151,6 +145,30 @@ func (r *Metal3MachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+
+	// set pause annotation on associated bmh (if any)
+	if cluster.Spec.Paused {
+		err := machineMgr.SetPauseAnnotation(ctx)
+		if err != nil {
+			machineLog.Info("failed to set pause annotation on associated bmh")
+			return ctrl.Result{}, nil
+		}
+	}
```
maelk: this would also never run if the cluster is paused, because of the return on L 160. It needs to move up too

controllers/metal3machine_controller.go
```diff
@@ -278,7 +330,7 @@ func (r *Metal3MachineReconciler) Metal3ClusterToMetal3Machines(o handler.MapObj
 		if m.Spec.InfrastructureRef.Name == "" {
 			continue
 		}
-		name := client.ObjectKey{Namespace: m.Namespace, Name: m.Name}
+		name := client.ObjectKey{Namespace: m.Spec.InfrastructureRef.Namespace, Name: m.Name}
```
maelk: This might be empty, you need to default it to m.Namespace if m.Spec.InfrastructureRef.Namespace == ""

baremetal/metal3machine_manager.go
kashifest: ok. I'll change it.

baremetal/metal3machine_manager.go
kashifest: Its not removing in case it is set by the user, thats why it is check, but I can rename it.

baremetal/metal3machine_manager_test.go
kashifest: good point. I'll add it.

controllers/metal3machine_controller.go
kashifest: true.

controllers/metal3machine_controller.go
kashifest: empty-case is checked in line 330. I don't that what you suggested is necessary

maelk: Name and namespace are different. You might have empty namespace but given name

kashifest: Is it possible? Didn't know that? OK, I will default it

controllers/metal3machine_controller.go
```diff
@@ -248,6 +270,37 @@ func (r *Metal3MachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	capiMachineList := &capi.MachineList{}
+	if err := r.Client.List(context.TODO(), capiMachineList, client.InNamespace(c.Namespace), client.MatchingLabels(labels)); err != nil {
+		r.Log.Error(err, "failed to list Metal3Machines")
+		return nil
+	}
+	for _, m := range capiMachineList.Items {
+		if m.Spec.InfrastructureRef.Name == "" {
+			continue
+		}
+		name := client.ObjectKey{Namespace: m.Namespace, Name: m.Name}
```
maelk: It should not be m.Name, it should be m.Spec.InfrastructureRef.Name

controllers/metal3machine_controller.go
```diff
@@ -248,6 +270,37 @@ func (r *Metal3MachineReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if m.Spec.InfrastructureRef.Namespace != "" {
+			name = client.ObjectKey{Namespace: m.Spec.InfrastructureRef.Namespace, Name: m.Name}
```
maelk: Same here

controllers/metal3machine_controller.go
```diff
@@ -278,7 +331,7 @@ func (r *Metal3MachineReconciler) Metal3ClusterToMetal3Machines(o handler.MapObj
 		if m.Spec.InfrastructureRef.Name == "" {
 			continue
 		}
-		name := client.ObjectKey{Namespace: m.Namespace, Name: m.Name}
+		name := client.ObjectKey{Namespace: m.Spec.InfrastructureRef.Namespace, Name: m.Name}
```
maelk: This also needs defaulting, and it should not be m.Name but m.Spec.InfrastructureRef.Name

kashifest: This is a bug as you reported, fixing it in a separate PR

controllers/metal3machine_controller.go
kashifest: Done

controllers/metal3machine_controller.go
kashifest: Done

PR #32, opened 2020-03-23
🐛 Fix Namespace handling

### LINKED ISSUE #29: Fix the namespaces assumptions and handling
Multiple assumptions were taken when developing regarding to namespaces to which the objects belong, and the code does not use the same assumptions consistently. For example, when it comes to the userData secret, if the BMH is in a different namespace, another secret is created. However, in other places, it is assumed that the objects are in the same namespace, and that namespace is referred to either by one or the other object's namespace

We should define a clear way to handle the object namespacing, making it uniform in the code, and verifying that the RBAC access are set in a way that allows the controllers to access them to keep everything working fine.

We should consider how CAPI does it, i.e. all objects for a cluster in the same namespace, and the impact it has in Metal3 case. For example, if some BMH would be shared between clusters, those clusters would need to be in the same namespace. That is if BMO is included in the provider-components.

**Environment:**

- Cluster-api version: v1alpha3
- CAPM3 version: v1alpha3 and v1alpha4
- Minikube version: NA
- environment (metal3-dev-env or other): NA
- Kubernetes version: (use `kubectl version`): NA

/kind bug

maelk:
This PR fixes the use of namespaces when referring to objects. It was inconsistent and sometimes wrong. It includes :

- Use ConfigRef namespace if available for userdata secret
- refer to userdata namespace instead of host namespace

Fixes #29

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -59,6 +59,9 @@ const (
 	bmRoleNode          = "node"
 	userDataFinalizer   = "metal3machine.infrastructure.cluster.x-k8s.io/userData"
 	pausedAnnotationKey = "metal3.io/capm3"
+
+	// metal3SecretType defines the type of secret created by metal3
+	metal3SecretType corev1.SecretType = "infrastructure.cluster.x-k8s.io/secret"
 )
 
 // MachineManagerInterface is an interface for a ClusterManager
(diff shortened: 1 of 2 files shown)


fmuyassarov:
@maelk  Would you mind to rebase your commit please?

maelk:
/hold cancel
fixed

fmuyassarov:
I believe you have to wait until Kashif's metal3-dev-env fix is in place.

jan-est:
This looks good to me

PR #41, opened 2020-03-24
:bug: Fix Metal3ClusterToMetal3Machines function in metal3Machine Controller

### LINKED ISSUE #40: The Metal3ClusterToMetal3Machines function returns a machine
The Metal3ClusterToMetal3Machines returns references to a machine, instead of a metal3machine. This is a remaining bit from v1alpha1.

It should return a reference to a Metal3Machine, hence m.Spec.InfrastructureRef.Name if set.

/kind bug

kashifest:
Metal3ClusterToMetal3Machines function was wrongly returning Machine object instead of metal3Machine. This PR fixes this bug and the relevant tests. 

Fixes #40

### DIFF
--- a/controllers/metal3machine_controller_test.go
+++ b/controllers/metal3machine_controller_test.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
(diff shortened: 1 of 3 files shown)


controllers/metal3machine_controller_test.go
```diff
@@ -308,75 +307,90 @@ var _ = Describe("Metal3Machine manager", func() {
...
-		}
-		Expect(len(out)).To(Equal(2), "Expected 2 Metal3 machines to reconcile but got %d", len(out))
+	type TestCaseMetal3ClusterToBMM struct {
+		Cluster       *capi.Cluster
+		M3Cluster     *infrav1.Metal3Cluster
+		Machine0      *capi.Machine
+		Machine1      *capi.Machine
+		Machine2      *capi.Machine
+		M3Machine0    *infrav1.Metal3Machine
+		ExpectRequest int
```
Jaakko-Os: I would understand 'ExpectedRequest' if it was 'bool'.
A comment would clarify the usage now when the type is 'int'. Below values 0,1,2 seem to be used. The meaning?

kashifest: The number of requests returned by the function.

kashifest: Changed it to bool.

controllers/metal3machine_controller_test.go
```diff
@@ -308,75 +307,90 @@ var _ = Describe("Metal3Machine manager", func() {
...
-			machineNames[i] = out[i].Name
-		}
-		Expect(len(out)).To(Equal(2), "Expected 2 Metal3 machines to reconcile but got %d", len(out))
+	type TestCaseMetal3ClusterToBMM struct {
+		Cluster       *capi.Cluster
+		M3Cluster     *infrav1.Metal3Cluster
+		Machine0      *capi.Machine
+		Machine1      *capi.Machine
+		Machine2      *capi.Machine
+		M3Machine0    *infrav1.Metal3Machine
```
maelk: you don't need this M3Machine0. The test is not fetching it

kashifest: removed. thanks.

Issue #48, opened 2020-03-26
Comparison including version should only use Kind and group


maelk:
**What steps did you take and what happened:**
In some places we compare objects group version and kind, when we should only compare group and kind to keep things working across upgrades. For example https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/controllers/metal3machine_controller.go#L350 . 

This needs to be fixed in both v0.3.1 and v0.4.0

/kind bug

maelk:
This also needs fixing in both versions : https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L770

maelk:
@jan-est did you fix it for the release-0.3 branch ?

maelk:
merged
/close

PR #49, opened 2020-03-26
 🐛 PR fixes issue #48

jan-est:
This PR fix comparison problem issued in [48](https://github.com/metal3-io/cluster-api-provider-metal3/issues/48) by compare only group and kind in metal3machine_controller and in metal3machine_manager

### DIFF
--- a/controllers/metal3machine_controller.go
+++ b/controllers/metal3machine_controller.go
@@ -347,7 +347,7 @@ func (r *Metal3MachineReconciler) BareMetalHostToMetal3Machines(obj handler.MapO
 	if host, ok := obj.Object.(*bmh.BareMetalHost); ok {
 		if host.Spec.ConsumerRef != nil &&
 			host.Spec.ConsumerRef.Kind == "Metal3Machine" &&
-			host.Spec.ConsumerRef.APIVersion == capm3.GroupVersion.String() {
+			host.Spec.ConsumerRef.GroupVersionKind().Group == capm3.GroupVersion.Group {
 			return []ctrl.Request{
 				ctrl.Request{
 					NamespacedName: types.NamespacedName{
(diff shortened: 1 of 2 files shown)


controllers/metal3machine_controller_test.go
```diff
@@ -431,7 +431,7 @@ var _ = Describe("Metal3Machine manager", func() {
 						},
 					},
 				},
-				ExpectRequest: true,
+				ExpectRequest: false,
```
maelk: that should still be true IMO

controllers/metal3machine_controller_test.go
```diff
@@ -427,7 +427,7 @@ var _ = Describe("Metal3Machine manager", func() {
 							Name:       "someothermachine",
 							Namespace:  "myns",
 							Kind:       "Metal3Machine",
-							APIVersion: infrav1.GroupVersion.String(),
+							APIVersion: infrav1.GroupVersion.Group,
```
maelk: I don't think this is the correct way to do it. The test, if not passing shows that there is a problem in the matching function

controllers/metal3machine_controller.go
```diff
@@ -345,9 +346,10 @@ func (r *Metal3MachineReconciler) Metal3ClusterToMetal3Machines(o handler.MapObj
 // BareMetalHost and that BareMetalHost references a Metal3Machine.
 func (r *Metal3MachineReconciler) BareMetalHostToMetal3Machines(obj handler.MapObject) []ctrl.Request {
 	if host, ok := obj.Object.(*bmh.BareMetalHost); ok {
+		capm3GV := schema.ParseGroupResource(capm3.GroupVersion.String())
```
maelk: isn't capm3.GroupVersion already capm3GV ?

jan-est: You are right! I fixed it!

Issue #46, opened 2020-03-25
UpdateObject function should operate on a deepcopy of the object


maelk:
**What steps did you take and what happened:**
the UpdateObject (and maybe createObject too) functions in metal3machine_manager.go (branch release-0.3) should operate on deepcopies of the object. Using the object directly translates into a lost status.

- Cluster-api version: 0.3.2
- CAPBM version: 0.3.0

/kind bug

maelk:
This should be fixed both in master and in branch release-0.3

Jaakko-Os:
How was the 'lost status' detected?
Is there a test case that reproduces this problem?

maelk:
it was detected in the metadata implementation phase. There is no test case that tests this, but you could add one on the updateObject function. Calling the function with an object with the status , after the call the object should retain its status.

PR #50, opened 2020-03-26
:bug: UpdateObject function should operate on a deepcopy of the object

Jaakko-Os:
Fixes: #46

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -207,10 +207,9 @@ func (m *MachineManager) SetPauseAnnotation(ctx context.Context) error {
 		if _, ok := annotations[bmh.PausedAnnotation]; ok {
 			m.Log.Info("BaremetalHost is already paused")
 			return nil
-		} else {
-			m.Log.Info("Adding PausedAnnotation in BareMetalHost")
-			host.Annotations[bmh.PausedAnnotation] = pausedAnnotationKey
 		}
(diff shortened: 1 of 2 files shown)


maelk:
It should be here https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L1071 and here https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L1079

baremetal/metal3machine_manager.go
```diff
@@ -207,10 +207,9 @@ func (m *MachineManager) SetPauseAnnotation(ctx context.Context) error {
 		if _, ok := annotations[bmh.PausedAnnotation]; ok {
 			m.Log.Info("BaremetalHost is already paused")
 			return nil
-		} else {
-			m.Log.Info("Adding PausedAnnotation in BareMetalHost")
-			host.Annotations[bmh.PausedAnnotation] = pausedAnnotationKey
 		}
+		m.Log.Info("Adding PausedAnnotation in BareMetalHost")
+		host.Annotations[bmh.PausedAnnotation] = pausedAnnotationKey
```
maelk: why changing this ? Kashif did it this way so that we would not override the user's pause annotation

Jaakko-Os: lint complaint fix, the 'else' has no effect since the 'if' has a 'return'

maelk: ah ok then fine by me

PR #51, opened 2020-03-26
:bug: Multiple updates operation on the BMH #33

### LINKED ISSUE #33: Multiple updates operation on the BMH
Currently, CAPM3 can execute multiple update operations on the BMH. This is problematic because the update triggers a reconcile of the BMH while we still process the BMH, regularly ending in a conflict, triggering a requeue.

We should postpone the update operations as much as possible to do only one update per reconciliation of the Metal3machine maximum.

/kind bug

mikkosest:
Currently, CAPM3 can execute multiple update operations on the BMH. This is problematic because the update triggers a reconcile of the BMH while we still process the BMH, regularly ending in a conflict, triggering a requeue.
 
Fixes #33

### DIFF
--- a/controllers/metal3machine_controller_test.go
+++ b/controllers/metal3machine_controller_test.go
@@ -85,11 +85,13 @@ func setReconcileNormalExpectations(ctrl *gomock.Controller,
 		if tc.AssociateFails {
 			m.EXPECT().Associate(context.TODO()).Return(errors.New("Failed"))
 			m.EXPECT().GetBaremetalHostID(context.TODO()).MaxTimes(0)
-			m.EXPECT().Update(context.TODO()).MaxTimes(0)
 			return m
 		} else {
 			m.EXPECT().Associate(context.TODO()).Return(nil)
 		}
+		m.EXPECT().Update(context.TODO()).MaxTimes(0)
+	} else {
+		m.EXPECT().Update(context.TODO())
 	}
 
 	// if node is now associated, if getting the ID fails, we do not go further
@@ -97,7 +99,6 @@ func setReconcileNormalExpectations(ctrl *gomock.Controller,
 		m.EXPECT().GetBaremetalHostID(context.TODO()).Return(nil,
 			errors.New("Failed"),
 		)
-		m.EXPECT().Update(context.TODO()).MaxTimes(0)
 		m.EXPECT().SetProviderID("abc").MaxTimes(0)
 		return m
 	}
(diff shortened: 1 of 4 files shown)


controllers/metal3machine_controller_test.go
```diff
@@ -74,7 +74,6 @@ func setReconcileNormalExpectations(ctrl *gomock.Controller,
 	if tc.BootstrapNotReady {
 		m.EXPECT().HasAnnotation().MaxTimes(0)
 		m.EXPECT().GetBaremetalHostID(context.TODO()).MaxTimes(0)
-		m.EXPECT().Update(context.TODO()).MaxTimes(0)
```
maelk: why removing it from here ? if bootstrap is not ready, Update should not be called (neither Associate btw).

controllers/metal3machine_controller_test.go
```diff
@@ -85,19 +84,20 @@ func setReconcileNormalExpectations(ctrl *gomock.Controller,
 		if tc.AssociateFails {
 			m.EXPECT().Associate(context.TODO()).Return(errors.New("Failed"))
 			m.EXPECT().GetBaremetalHostID(context.TODO()).MaxTimes(0)
-			m.EXPECT().Update(context.TODO()).MaxTimes(0)
 			return m
 		} else {
 			m.EXPECT().Associate(context.TODO()).Return(nil)
+			m.EXPECT().Update(context.TODO()).MaxTimes(0)
```
maelk: In both cases Update should not be called, so you can add `m.EXPECT().Update(context.TODO()).MaxTimes(0)` outside of the if/else block

controllers/metal3machine_controller_test.go
```diff
@@ -85,19 +84,20 @@ func setReconcileNormalExpectations(ctrl *gomock.Controller,
...
 		}
+	} else {
+		m.EXPECT().Update(context.TODO()).MaxTimes(2)
```
maelk: why max times 2 ? if it should be called, remove the MaxTimes, it will ensure that it is called at least once.

mikkosest: Fixed

controllers/metal3machine_controller_test.go
mikkosest: Fixed

controllers/metal3machine_controller_test.go
mikkosest: Fixed

maelk:
don't forget to submit the same for the branch release-0.3

PR #54, opened 2020-03-27
:bug: Multiple updates operation on the BMH #33

### LINKED ISSUE #33: Multiple updates operation on the BMH

mikkosest:
Currently, CAPM3 can execute multiple update operations on the BMH. This is problematic because the update triggers a reconcile of the BMH while we still process the BMH, regularly ending in a conflict, triggering a requeue.

Fixes #33

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -259,6 +259,11 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 		return err
 	}
 
+	err = m.updateObject(ctx, host)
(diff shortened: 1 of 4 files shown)


mikkosest:
//cc @maelk

PR #58, opened 2020-03-30
add tests for updateObj and createObj

Jaakko-Os:
https://github.com/metal3-io/cluster-api-provider-metal3/issues/46

### DIFF
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -2302,6 +2302,71 @@ var _ = Describe("Metal3Machine manager", func() {
 		}),
 	)
 
+	DescribeTable("Test updateObj function, original object shall remain unchanged",
+		func(Host *bmh.BareMetalHost) {
+			c := fakeclient.NewFakeClientWithScheme(setupSchemeMm(), Host)
+			machineMgr, err := NewMachineManager(c, nil, nil, nil, nil, klogr.New())
+			Expect(err).NotTo(HaveOccurred())
+
+			tcHost := bmh.BareMetalHost{}
+			err = c.Get(context.TODO(),
+				client.ObjectKey{
+					Name:      Host.Name,
+					Namespace: Host.Namespace,
+				},
+				&tcHost,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = machineMgr.updateObject(context.TODO(), Host)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = c.Get(context.TODO(),
+				client.ObjectKey{
+					Name:      Host.Name,
+					Namespace: Host.Namespace,
+				},
+				Host,
+			)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(tcHost.Status).To(Equal(Host.Status))
+			Expect(tcHost.ResourceVersion).NotTo(Equal(Host.ResourceVersion))
+		},
+		Entry("updateObj host",
+			newBareMetalHost("myhost-obj", nil, bmh.StateProvisioned, bmhPowerStatus(), true, true),
+		),
+		Entry("updateObj host next",
+			newBareMetalHost("myhost", nil, bmh.StateNone, bmhStatus(), false, false),
+		),
+	)
+
+	DescribeTable("Test createObj function",
+		func(BootstrapSecret *corev1.Secret, label bool) {
+			objects := []runtime.Object{}
+			c := fakeclient.NewFakeClientWithScheme(setupSchemeMm(), objects...)
+			machineMgr, err := NewMachineManager(c, nil, nil, nil, nil, klogr.New())
+			Expect(err).NotTo(HaveOccurred())
+
+			err = machineMgr.createObject(context.TODO(), BootstrapSecret)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = c.Get(context.TODO(),
+				client.ObjectKey{
+					Name:      BootstrapSecret.Name,
+					Namespace: BootstrapSecret.Namespace,
+				},
+				BootstrapSecret,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			baseSecret := newBMCSecret("mycredentials", label)
+			Expect(baseSecret.TypeMeta).To(Equal(BootstrapSecret.TypeMeta))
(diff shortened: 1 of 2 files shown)


baremetal/metal3machine_manager_test.go
```diff
@@ -2302,6 +2302,93 @@ var _ = Describe("Metal3Machine manager", func() {
 		}),
 	)
 
+	type testCaseUpdateObj struct {
+		Host      *bmh.BareMetalHost
+		BMMachine *capm3.Metal3Machine
```
maelk: why do we need the BMMachine ?

baremetal/metal3machine_manager_test.go
```diff
@@ -2302,6 +2302,93 @@ var _ = Describe("Metal3Machine manager", func() {
...
+				bmmObjectMetaWithValidAnnotations(),
+			),
+			Host: newBareMetalHost("myhost", nil, bmh.StateNone, bmhStatus(), false, false),
+		}),
+	)
+
+	type testCaseCreateObj struct {
+		Machine         *capi.Machine
+		Host            *bmh.BareMetalHost
+		BMMachine       *capm3.Metal3Machine
```
maelk: why do we need Machine, Host and BMMachine ?

baremetal/metal3machine_manager_test.go
Jaakko-Os: removed

baremetal/metal3machine_manager_test.go
Jaakko-Os: removed

baremetal/metal3machine_manager_test.go
```diff
@@ -2302,6 +2302,70 @@ var _ = Describe("Metal3Machine manager", func() {
 		}),
 	)
 
+	type testCaseUpdateObj struct {
+		Host *bmh.BareMetalHost
+	}
```
maelk: if you have only one object in your struct, you don't need to create a struct, you can directly pass the BareMetalHost instead of a testCaseUpdateObj

baremetal/metal3machine_manager_test.go
```diff
@@ -2302,6 +2302,70 @@ var _ = Describe("Metal3Machine manager", func() {
...
+			Host: newBareMetalHost("myhost", nil, bmh.StateProvisioned, bmhPowerStatus(), true, true),
+		}),
+		Entry("updateObj host next", testCaseUpdateObj{
+			Host: newBareMetalHost("myhost", nil, bmh.StateNone, bmhStatus(), false, false),
+		}),
+	)
+
+	type testCaseCreateObj struct {
+		BootstrapSecret *corev1.Secret
+	}
```
maelk: same here

baremetal/metal3machine_manager_test.go
Jaakko-Os: done

baremetal/metal3machine_manager_test.go
Jaakko-Os: done

baremetal/metal3machine_manager_test.go
```diff
@@ -2302,6 +2302,57 @@ var _ = Describe("Metal3Machine manager", func() {
...
+			c := fakeclient.NewFakeClientWithScheme(setupSchemeMm(), Host)
+			machineMgr, err := NewMachineManager(c, nil, nil, nil, nil, klogr.New())
+
+			Expect(err).NotTo(HaveOccurred())
+
+			tcHost := Host.DeepCopyObject()
+			err = machineMgr.updateObject(context.TODO(), Host)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(tcHost).To(Equal(Host))
```
kashifest: Shouldn't you get the host object from the fakeclient rather than making a Deepcopy and comparing against it? How are you sure that updateObject() function has not changed the host???

baremetal/metal3machine_manager_test.go
```diff
@@ -2302,6 +2302,57 @@ var _ = Describe("Metal3Machine manager", func() {
...
+			c := fakeclient.NewFakeClientWithScheme(setupSchemeMm(), objects...)
+			machineMgr, err := NewMachineManager(c, nil, nil, nil, nil, klogr.New())
+
+			Expect(err).NotTo(HaveOccurred())
+
+			tcSecret := BootstrapSecret.DeepCopyObject()
+			err = machineMgr.createObject(context.TODO(), BootstrapSecret)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(tcSecret).To(Equal(BootstrapSecret))
```
kashifest: Same concern here  as above. The secret should be fetched from the fakeclient IMO.

baremetal/metal3machine_manager_test.go
Jaakko-Os: My goal is to compare the entire original host object '**tcHost**' to the host object '**Host**' given to the updateObj function. With DeepCopyObj I make sure that the entire object is copied to 'tcHost'.
In this case a simple assignment would likely work but deepcopy handles possible lists, maps etc correctly without user attention.

baremetal/metal3machine_manager_test.go
Jaakko-Os: I don't see your point. Test case checks basically that createObject doesn't tamper the BootstrapSecret in any way.

baremetal/metal3machine_manager_test.go
kashifest: My point is if the goal is to check the behavior of updateObj function, then shouldn't the object be fetched from the client and see how if the object has bee changed?

Jaakko-Os: My understanding is that this is achieved via the machineMgr.
Please correct me if I'm wrong.

kashifest: If you go through the machine manager test file, you will see in few other test cases the hosts are being fetched through the client after a function call (i.e. https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager_test.go#L632 ). That is the proper way to compare the changes in objects before and after an operation IMO.

Jaakko-Os: Thanks for the example. I got it, done.

baremetal/metal3machine_manager_test.go
Jaakko-Os: I see, done.

PR #68, opened 2020-04-08
✨ Make BMO as part of CAPM3 deployment

furkatgofurov7:
This PR deploys BMO as part of capm3-controller manager. It is needed for fully functional pivoting.

Design proposal is given in [PR#91](https://github.com/metal3-io/metal3-docs/pull/91)
Depends on [PR#522](https://github.com/metal3-io/baremetal-operator/pull/522/)
Related to: 
- [baremetal-operator/pull/548](https://github.com/metal3-io/baremetal-operator/pull/548)
- [metal3-dev-env/pull/336](https://github.com/metal3-io/metal3-dev-env/pull/336)


maelk:
the problem with the CI is coming from the fact that you added a sed in "set-manifest-image" make target, but the variables are empty, so it messes the patch applied for the kustomization

furkatgofurov7:
> the problem with the CI is coming from the fact that you added a sed in "set-manifest-image" make target, but the variables are empty, so it messes the patch applied for the kustomization



> the problem with the CI is coming from the fact that you added a sed in "set-manifest-image" make target, but the variables are empty, so it messes the patch applied for the kustomization

True, thanks for the review, updated

furkatgofurov7:
Sorry, I triggered already passed test without refreshing the page.

maelk:
As soon as you remove the hold, this will go in

PR #69, opened 2020-04-09
⚠️ Introduce templating for metadata and network data

maelk:
This PR implements https://github.com/metal3-io/metal3-docs/pull/79 . It introduces Metal3DataTemplate and Metal3Data objects and their reconciliation logic.


controllers/metal3machine_controller.go
```diff
@@ -365,6 +404,24 @@ func (r *Metal3MachineReconciler) BareMetalHostToMetal3Machines(obj handler.MapO
...
+// Metal3DataToMetal3Machines will return a reconcile request for a Metal3Machine if the event is for a
+// Metal3Data and that Metal3Data references a Metal3Machine.
+func (r *Metal3MachineReconciler) Metal3DataToMetal3Machines(obj handler.MapObject) []ctrl.Request {
+	if m3d, ok := obj.Object.(*capm3.Metal3Data); ok {
+		if m3d.Spec.Metal3Machine != nil {
+			return []ctrl.Request{
+				ctrl.Request{
+					NamespacedName: types.NamespacedName{
+						Name:      m3d.Spec.Metal3Machine.Name,
+						Namespace: m3d.Namespace,
```
kashifest, line 416: why not checking if the m3d.Spec.Metal3Machine.Namespace exists or not?

controllers/metal3machine_controller.go
```diff
@@ -182,43 +187,62 @@ func (r *Metal3MachineReconciler) reconcileNormal(ctx context.Context,
...
 	if machineMgr.IsProvisioned() {
-		err := machineMgr.Update(ctx)
-		return ctrl.Result{}, err
+		errType := capierrors.UpdateMachineError
+		return checkMachineError(machineMgr, machineMgr.Update(ctx),
+			"Failed to update the Metal3Machine", errType,
+		)
 	}
 
+	errType := capierrors.CreateMachineError
```
kashifest, line 198: I would rather move the declaration close to where it is used, i.e. line# 205

controllers/metal3datatemplate_controller.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (r *Metal3DataTemplateReconciler) reconcileNormal(ctx context.Context,
+	metadataMgr baremetal.DataTemplateManagerInterface,
+) (ctrl.Result, error) {
+	// If the Metal3DataTemplate doesn't have finalizer, add it.
+	metadataMgr.SetFinalizer()
+
+	err := metadataMgr.RecreateStatusConditionally(ctx)
+	if err != nil {
+		return checkMetadataError(err, "Failed to recreate the status")
```
kashifest, line 123: what is meant by recreate status?

controllers/metal3datatemplate_controller.go
```diff
@@ -0,0 +1,176 @@
...
+	defer func() {
+		err := helper.Patch(ctx, capm3Metadata)
+		if err != nil {
+			metadataLog.Info("failed to Patch capm3Metadata")
+		}
+	}()
+
+	// Fetch the Cluster.
+	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, capm3Metadata.ObjectMeta)
+	if capm3Metadata.ObjectMeta.DeletionTimestamp.IsZero() {
```
kashifest, line 79: this condition is also checked at line 107 for reconciling delete. I am wondering the need to check this multiple times. Also in case there is deletion timestamp set, will it then ignore the err if for example when the cluster label was missing?

controllers/metal3machine_controller_test.go
```diff
@@ -455,6 +455,85 @@ var _ = Describe("Metal3Machine manager", func() {
...
+
+			} else {
+				Expect(len(reqs)).To(Equal(0), "Expected 0 request, found %d", len(reqs))
+
+			}
+		},
+		Entry("No Metal3Machine in Spec",
+			TestCaseM3DToBMM{
+				Data: &infrav1.Metal3Data{
+					ObjectMeta: metav1.ObjectMeta{
```
Jaakko-Os, line 493: ObjectMeta struct could be defined separately and re-used. Less repetition and shorter TC Entry statements.

controllers/metal3datatemplate_controller_test.go
```diff
@@ -0,0 +1,414 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "abc",
+					Namespace: "def",
+				},
+			},
+		}),
+		Entry("Cluster not found", testCaseReconcile{
+			m3dt: &infrav1.Metal3DataTemplate{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "abc",
```
Jaakko-Os, line 134: Avoid repetition

maelk:
The code coverage did not change much, it slightly increased. In this PR, it is at :

- api : around 30% in average
- baremetal: 85%, 290 test cases
- controller: 86%, 85 test cases

controllers/metal3datatemplate_controller.go
maelk: that is if the object was moved, and the status is gone, it will repopulate it

controllers/metal3machine_controller.go
maelk: done, thanks

controllers/metal3machine_controller.go
maelk: done, thanks

controllers/metal3datatemplate_controller.go
maelk: yes, the idea is that in a case of deletion, we ignore a failure on getting the cluster. Otherwise we could be blocking here if the cluster was removed already.

controllers/metal3datatemplate_controller_test.go
maelk: done, thanks

controllers/metal3machine_controller_test.go
maelk: done, thanks

controllers/metal3data_controller_test.go
```diff
@@ -0,0 +1,301 @@
...
+						Name:      "abc",
+						Namespace: "def",
+					},
+				},
+			}),
+			Entry("Cluster not found", testCaseReconcile{
+				m3d: &infrav1.Metal3Data{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "abc",
+						Namespace: "def",
```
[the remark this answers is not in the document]
maelk: done, thanks

controllers/metal3data_controller_test.go
```diff
@@ -118,17 +125,14 @@ var _ = Describe("Metal3Data manager", func() {
...
 			}),
 			Entry("Cluster not found", testCaseReconcile{
 				m3d: &infrav1.Metal3Data{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "abc",
-						Namespace: "def",
+						Namespace: "myns",
 						Labels: map[string]string{
 							capi.ClusterLabelName: "abc",
 						},
```
Jaakko-Os, line 138: these could be defined like 'ObjectMeta: testObjectMetaWithLabel' to avoid repetition

maelk: done, thanks

PR #76, opened 2020-05-11
✨ Use Claim model for Metadata and IPAM controllers

maelk:
This PR introduces Metal3IPClaim and Metal3DataClaim objects to use a claim-based allocation model for the metadata and IPAM controllers


maelk:
This is on top of #75

maelk:
it was fixed in the following commit @Jaakko-Os

PR #80, opened 2020-05-26
🐛 Use patch helper to update BMH

### LINKED ISSUE #79: Infinite reconciliation loop when waiting for provisioning
When a node is being provisioned, the controller keeps reconciling the metal3machine about every second. 

It should not reconcile the metal3machine until the BMH is updated, and probably provisioned.

It looks like the call to updateObject function in https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L651 updates the BMH object that in turn triggers a reconciliation of the metal3machine. Calls to the updateObject function for BMH should happen only when there has been a change on the BMH object. Some refactoring and the use of patch helper from CAPI would be very much needed to prevent this waste of resources.

**Environment:**

- Cluster-api version: 0.3.6
- CAPM3 version: 0.4.X
- environment (metal3-dev-env or other): metal3-dev-env

/kind bug

maelk:
Using the updateObject function was triggering an infinite loop,
as the metal3machine were reconciled when BMH were updated and
the reconciliation was always updating the BMH

Fixes #79

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -37,12 +37,14 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/selection"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/pointer"
 	capi "sigs.k8s.io/cluster-api/api/v1alpha3"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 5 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -208,13 +210,13 @@ func (m *MachineManager) SetPauseAnnotation(ctx context.Context) error {
...
 	}
-	return updateObject(m.client, ctx, host)
+	return helper.Patch(ctx, host)
 }
 
 // GetBaremetalHostID return the provider identifier for this machine
 func (m *MachineManager) GetBaremetalHostID(ctx context.Context) (*string, error) {
 	// look for associated BMH
-	host, err := m.getHost(ctx)
+	host, _, err := m.getHost(ctx)
```
furkatgofurov7, line 219: is blank identifier replacing `helper` ?

baremetal/metal3machine_manager.go
```diff
@@ -684,8 +702,13 @@ func (m *MachineManager) exists(ctx context.Context) (bool, error) {
 // getHost gets the associated host by looking for an annotation on the machine
 // that contains a reference to the host. Returns nil if not found. Assumes the
 // host is in the same namespace as the machine.
-func (m *MachineManager) getHost(ctx context.Context) (*bmh.BareMetalHost, error) {
-	return getHost(ctx, m.Metal3Machine, m.client, m.Log)
+func (m *MachineManager) getHost(ctx context.Context) (*bmh.BareMetalHost, *patch.Helper, error) {
+	host, err := getHost(ctx, m.Metal3Machine, m.client, m.Log)
```
furkatgofurov7, line 706: why `m.Log` is needed here?

baremetal/metal3machine_manager.go
maelk: yes, because helper is unused here

baremetal/metal3machine_manager.go
maelk: because getHost function uses it to log some messages

PR #81, opened 2020-05-26
✨ Add ChecksumType field in image

maelk:
This PR adds support for the new BMH field ChecksumType in the image, and updates all the dependencies.


furkatgofurov7:
@maelk, is md5 a default `ChecksumType` in CAPM3? are sha256, sha512 can be used as well?

maelk:
yes, metal3-dev-env uses md5sum to compute the checksum. Other are also supported.

furkatgofurov7:
Looks good to me. But someone who has more expertise in repo shall review.

PR #83, opened 2020-05-27
🏃 refactor code to replace bmm by m3m

maelk:
This PR does a bit of code refactoring

### DIFF
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -57,17 +57,17 @@ const (
 var ProviderID = "metal3://12345ID6789"
 var CloudInitData = []byte("metal3:cloudInitData1010101test__hello")
 
-func bmmSpec() *capm3.Metal3MachineSpec {
+func m3mSpec() *capm3.Metal3MachineSpec {
 	return &capm3.Metal3MachineSpec{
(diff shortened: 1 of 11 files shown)


baremetal/metal3machine_manager_test.go
```diff
@@ -78,26 +78,26 @@ func bmmSpecAll() *capm3.Metal3MachineSpec {
...
 		},
 	}
 }
 
 func consumerRef() *corev1.ObjectReference {
 	return &corev1.ObjectReference{
-		Name:       "mybmmachine",
+		Name:       "mym3machine",
 		Namespace:  "myns",
 		Kind:       "BMMachine",
```
fmuyassarov, line 102: Should the value of ` Kind:` be kept as is?

jan-est: Should it be M3Machine?

maelk: thanks, I replaced only bmm, so I updated the patch to change BMM also

maelk:
Another XL PR, but it is only a search and replace, so should be quite straightforward to review. Sorry for that

furkatgofurov7:
lgtm

PR #84, opened 2020-05-28
🐛 Use patch helper to update BMH

### LINKED ISSUE #79: Infinite reconciliation loop when waiting for provisioning

maelk:
Using the updateObject function was triggering an infinite loop,
as the metal3machine were reconciled when BMH were updated and
the reconciliation was always updating the BMH

Fixes #79

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -27,24 +27,25 @@ import (
 	// comment for go-lint
 	"github.com/go-logr/logr"
 
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/utils/pointer"
-
 	bmh "github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1"
 	capm3 "github.com/metal3-io/cluster-api-provider-metal3/api/v1alpha3"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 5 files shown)


baremetal/utils.go
```diff
@@ -44,3 +53,26 @@ type NotFoundError struct {
...
 }
+
+func patchIfFound(ctx context.Context, helper *patch.Helper, host runtime.Object) error {
+	err := helper.Patch(ctx, host)
+	if err != nil {
+		notFound := true
+		if aggr, ok := err.(kerrors.Aggregate); ok {
+			for _, kerr := range aggr.Errors() {
+				if !apierrors.IsNotFound(kerr) {
+					notFound = false
```
kashifest, line 64: why not returning nil if `apierrors.IsNotFound(kerr)`

baremetal/metal3machine_manager.go
```diff
@@ -576,8 +585,19 @@ func (m *MachineManager) exists(ctx context.Context) (bool, error) {
 // getHost gets the associated host by looking for an annotation on the machine
 // that contains a reference to the host. Returns nil if not found. Assumes the
 // host is in the same namespace as the machine.
-func (m *MachineManager) getHost(ctx context.Context) (*bmh.BareMetalHost, error) {
-	annotations := m.Metal3Machine.ObjectMeta.GetAnnotations()
+func (m *MachineManager) getHost(ctx context.Context) (*bmh.BareMetalHost, *patch.Helper, error) {
```
kashifest, line 588: a bit confused why do we need two functions with same name, if the idea is to get helper also, may be the naming could be changed accordingly?

baremetal/utils.go
maelk: because it is already deleted, the end goal is achieved

baremetal/metal3machine_manager.go
maelk: one of them takes the manager, the other is a helper, that can be used in other places. Having the same name is not a problem, and IMO it clearly shows the flow, you call the gethost of the manager, that under the hood calls the helper of the same name

baremetal/utils.go
kashifest: not clear, you are anyhow returning nil when `if notFound`

maelk: oh, I see sorry, I did not get your question right. The problem is that you are getting an aggregate of errors, and I want to return nil if all errors are NotFound. If one of them is not NotFound, I want to return the normal error (or requeue). so we need to check them all, and notFound is true only if all of them are NotFound errors, in which case we return nil

PR #85, opened 2020-06-01
🏃 Remove code generation from unit test

### LINKED ISSUE #13: Speed up unit tests
**User Story**

Most of the time of the unit test jobs is spent on installing the tools and generating the code that is automatically generated. We should bake in all tools and not regenerate the code as it should be included in the repository.

/kind feature

digambar15:
generate-go script is taking lot of time in unit tests and everytime it generates the code which is not required for unit tests, so removing generate-go from unit.

Fixes #13

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,11 +175,6 @@ generate-go: $(CONTROLLER_GEN) $(MOCKGEN) $(CONVERSION_GEN) $(KUBEBUILDER) $(KUS
 		paths=./api/... \
 		object:headerFile=./hack/boilerplate/boilerplate.generatego.txt
 
-	$(CONVERSION_GEN) \
-		--input-dirs=./api/v1alpha2 \
-		--output-file-base=zz_generated.conversion \
-		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
-
 	$(MOCKGEN) \
 	  -destination=./baremetal/mocks/zz_generated.metal3cluster_manager.go \
 	  -source=./baremetal/metal3cluster_manager.go \
@@ -193,6 +188,37 @@ generate-go: $(CONTROLLER_GEN) $(MOCKGEN) $(CONVERSION_GEN) $(KUBEBUILDER) $(KUS
 		-package=baremetal_mocks \
 		-copyright_file=./hack/boilerplate/boilerplate.generatego.txt \
 		MachineManagerInterface
+
+	$(MOCKGEN) \
+	  -destination=./baremetal/mocks/zz_generated.metal3datatemplate_manager.go \
+	  -source=./baremetal/metal3datatemplate_manager.go \
+		-package=baremetal_mocks \
+		-copyright_file=./hack/boilerplate/boilerplate.generatego.txt \
+		DataTemplateManagerInterface
+
+	$(MOCKGEN) \
+	  -destination=./baremetal/mocks/zz_generated.metal3data_manager.go \
+	  -source=./baremetal/metal3data_manager.go \
+		-package=baremetal_mocks \
+		-copyright_file=./hack/boilerplate/boilerplate.generatego.txt \
+		DataManagerInterface
+
+	$(MOCKGEN) \
+	  -destination=./baremetal/mocks/zz_generated.manager_factory.go \
+	  -source=./baremetal/manager_factory.go \
+		-package=baremetal_mocks \
+		-copyright_file=./hack/boilerplate/boilerplate.generatego.txt \
+		ManagerFactoryInterface
+
+	$(CONVERSION_GEN) \
+		--input-dirs=./api/v1alpha2 \
+		--output-file-base=zz_generated.conversion \
+		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
+
+	$(CONVERSION_GEN) \
+		--input-dirs=./api/v1alpha3 \
+		--output-file-base=zz_generated.conversion \
+		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
 
 .PHONY: generate-manifests
 generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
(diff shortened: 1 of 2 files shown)


digambar15:
> Thank you for the PR. The code that you are removing from line 166 to 239 is actually needed, it does the code generation from kube-builder, mockgen etc. Removing the generate target as a dependency on line 85 is the correct direction. However, we still need the code generation, for the user to run it manually for example, and since we are removing it from the unit tests, we should add a job that takes care of verifying that the generated code is up to date.
> So from this PR, you would need to not delete the line 166 to 239, and add a script in the hack folder that calls `make generate` and makes a diff between the current code and the generated one. It would succeed only if the diff is empty. We would create a CI job for that then.
> In addition, the Dockerfile.unit in the hack folder can be improved to install the tools and more go dependencies to accelerate tremendously the tests

Yes, I pushed latest commit. Working on second part (Installation of tools)

Makefile
```diff
@@ -189,37 +194,6 @@ generate-go: $(CONTROLLER_GEN) $(MOCKGEN) $(CONVERSION_GEN) $(KUBEBUILDER) $(KUS
...
-	$(CONVERSION_GEN) \
-		--input-dirs=./api/v1alpha2 \
-		--output-file-base=zz_generated.conversion \
-		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
-
-	$(CONVERSION_GEN) \
-		--input-dirs=./api/v1alpha3 \
-		--output-file-base=zz_generated.conversion \
-		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
-
```
maelk, line 222: this should not be modified

Makefile
```diff
@@ -175,6 +175,11 @@ generate-go: $(CONTROLLER_GEN) $(MOCKGEN) $(CONVERSION_GEN) $(KUBEBUILDER) $(KUS
 		paths=./api/... \
 		object:headerFile=./hack/boilerplate/boilerplate.generatego.txt
 
+	$(CONVERSION_GEN) \
+		--input-dirs=./api/v1alpha2 \
+		--output-file-base=zz_generated.conversion \
+		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
+
```
maelk, line 182: why moving this ?

Makefile
```diff
@@ -234,7 +208,7 @@ generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
 		rbac:roleName=manager-role
 
 .PHONY: generate-examples
-generate-examples: $(KUSTOMIZE) clean-examples ## Generate examples configurations to run a cluster.
```
maelk, line 237: this is actually needed, and running that target is actually fast if kustomize is already installed this way

hack/codegen.sh
```diff
@@ -0,0 +1,22 @@
...
+
+IS_CONTAINER=${IS_CONTAINER:-false}
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+  export XDG_CACHE_HOME=/tmp/.cache
+  mkdir /tmp/unit
+  cp -r ./* /tmp/unit
+  cd /tmp/unit
+  make generate
```
maelk, line 13: that's a good starting point. Here we need to compare the existing (in /go/src/github.com/metal3-io/cluster-api-provider-metal3) to what was generated and fail if there is a difference

Makefile
digambar15: Done!

Makefile
digambar15: Fixed

Makefile
digambar15: Done!

hack/codegen.sh
digambar15: Handled in codegen.sh

digambar15:
Don't review, will be working on one fix

digambar15:
Resolved all the changes, tested also. It works as expected

maelk:
that looks good to me, but we need to have the CI updated for this to be merged. Can you update the configuration of prow in github.com/metal3-io/project-infra ?

hack/codegen.sh
```diff
@@ -0,0 +1,27 @@
...
+set -eux
+
+IS_CONTAINER=${IS_CONTAINER:-false}
+ARTIFACTS=${ARTIFACTS:-/tmp}
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+  eval "$(go env)"
+  cd "${GOPATH}"/src/github.com/metal3-io/cluster-api-provider-metal3
+  export XDG_CACHE_HOME=/tmp/.cache
```
maelk, line 12: this needs to be before line 10

hack/codegen.sh
```diff
@@ -0,0 +1,27 @@
...
+  INPUT_FILES="api/v1alpha2/zz_generated.*.go api/v1alpha3/zz_generated.*.go api/v1alpha4/zz_generated.*.go baremetal/mocks/zz_generated.*.go"
+  cksum "$INPUT_FILES" > "$ARTIFACTS/lint.cksums.before"
```
maelk, line 14: putting quotes around $INPUT_FILES make it behave as a single string. you need to remove the quotes and put a shellcheck exception before that

hack/codegen.sh
```diff
@@ -0,0 +1,27 @@
...
+  export VERBOSE="--verbose"
+  make generate
+  cksum "$INPUT_FILES" > "$ARTIFACTS/lint.cksums.after"
```
maelk, line 17: same here

hack/codegen.sh
digambar15: Done!

hack/codegen.sh
digambar15: Done!

hack/codegen.sh
digambar15: Done!

PR #96, opened 2020-06-09
✨ PR prevents the use of BMH annotated as unhealthy

jan-est:
This PR adds the new `UnhealthyAnnotation = "capi.metal3.io/unhealthy"` and PR prevents the unhealthy BMH from being selected as the host for metal3machine. With new annotation users can mark BMH as unhealthy.

### DIFF
--- a/api/v1alpha4/common_types.go
+++ b/api/v1alpha4/common_types.go
@@ -20,6 +20,11 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 )
 
+const (
+	// UnhealthyAnnotation is the annotation that sets unhealthy status of BMH
+	UnhealthyAnnotation = "capi.metal3.io/unhealthy"
+)
+
 // APIEndpoint represents a reachable Kubernetes API endpoint.
 type APIEndpoint struct {
 	// Host is the hostname on which the API server is serving.
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -808,13 +808,18 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
 		default:
 			continue
 		}
-		// continue if BaremetalHost is paused
+
+		// continue if BaremetalHost is paused or marked with UnhealthyAnnotation
 		annotations := host.GetAnnotations()
 		if annotations != nil {
 			if _, ok := annotations[bmh.PausedAnnotation]; ok {
 				continue
 			}
+			if _, ok := annotations[capm3.UnhealthyAnnotation]; ok {
+				continue
+			}
 		}
+
 		if labelSelector.Matches(labels.Set(host.ObjectMeta.Labels)) {
 			m.Log.Info("Host matched hostSelector for Metal3Machine", "host", host.Name)
 			availableHosts = append(availableHosts, &hosts.Items[i])
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -496,6 +496,13 @@ var _ = Describe("Metal3Machine manager", func() {
 				Labels:    map[string]string{"key1": "value1"},
 			},
 		}
+		hostWithUnhealthyAnnotation := bmh.BareMetalHost{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:        "hostWithUnhealthyAnnotation",
+				Namespace:   "myns",
+				Annotations: map[string]string{capm3.UnhealthyAnnotation: "unhealthy"},
+			},
+		}
 
 		m3mconfig, infrastructureRef := newConfig("", map[string]string{},
 			[]capm3.HostSelectorRequirement{},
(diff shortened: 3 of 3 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -808,13 +808,23 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
...
 		// continue if BaremetalHost is paused
 		annotations := host.GetAnnotations()
 		if annotations != nil {
 			if _, ok := annotations[bmh.PausedAnnotation]; ok {
 				continue
 			}
 		}
+
+		// continue if BaremetalHost is unhealthy
+		annotations := host.GetAnnotations()
```
kashifest, line 821: annotations is already delared in line 813 isn't it? We don't need this line I suppose, we can even take the whole `if _, ok := annotations[bmh.UnhealthyAnnotation]; ok`  inside the if defined in line 814

jan-est: Good point, better to move it as part of the previous annotation check.

maelk:
this should have a go.mod update to use the new version of BMO with the annotation definition

baremetal/metal3machine_manager.go
```diff
@@ -808,13 +808,18 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
...
 		}
-		// continue if BaremetalHost is paused
+
+		// continue if BaremetalHost is paused or BaremetalHost is unhealthy
 		annotations := host.GetAnnotations()
 		if annotations != nil {
 			if _, ok := annotations[bmh.PausedAnnotation]; ok {
 				continue
 			}
+			if _, ok := annotations[bmh.UnhealthyAnnotation]; ok {
```
maelk, line 818: that will not be bmh anymore

jan-est: @maelk thanks. this is fixed.

baremetal/metal3machine_manager.go
```diff
@@ -808,13 +808,18 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
 		default:
 			continue
 		}
-		// continue if BaremetalHost is paused
+
+		// continue if BaremetalHost is paused or BaremetalHost is unhealthy
```
fmuyassarov, line 812: Maybe the comment also should be updated? since you are checking `[capm3.UnhealthyAnnotation]`

fmuyassarov: but otherwise lgtm

api/v1alpha4/common_types.go
```diff
@@ -20,6 +20,11 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 )
 
+const (
+	// UnhealthyAnnotation is the annotation that sets unhealthy status of BMH
+	UnhealthyAnnotation = "capm3.metal3.io/unhealthy"
```
maelk, line 25: capm3 and metal3 are a bit redundant here. what about "capi.metal3.io/unhealthy" ?

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/golang/mock v1.4.3
 	github.com/google/gofuzz v1.1.0
 	github.com/mailru/easyjson v0.7.1 // indirect
-	github.com/metal3-io/baremetal-operator v0.0.0-20200602195426-56927565ba1d
+	github.com/metal3-io/baremetal-operator v0.0.0-20200609125308-fbb69f6cce0c
```
maelk, line 13: you do not need this anymore

baremetal/metal3machine_manager.go
```diff
@@ -808,13 +808,18 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
...
 		}
-		// continue if BaremetalHost is paused
+
+		// continue if BaremetalHost is paused or marked with UnhealthyAnnotation
 		annotations := host.GetAnnotations()
 		if annotations != nil {
 			if _, ok := annotations[bmh.PausedAnnotation]; ok {
 				continue
 			}
+			if _, ok := annotations[capm3.UnhealthyAnnotation]; ok {
```
maelk, line 818: Can you add a unit test for that ?

baremetal/metal3machine_manager_test.go
```diff
@@ -565,6 +572,14 @@ var _ = Describe("Metal3Machine manager", func() {
 					ExpectedHostName: host2.Name,
 				},
 			),
+			Entry("Ignore hostWithUnhealthyAnnotation and pick host2, which lacks a ConsumerRer",
```
kashifest, line 575: consumerRef ??

jan-est: @kashifest this is testing the selection of the host. Unhealthy is ignored and one that does not yet have consumer is selected for machine.

fmuyassarov: I think @kashifest was referring to a typo.

kashifest: I understand that. I meant it should be "ConsumerRef" not "ConsumerRer" :D

PR #97, opened 2020-06-11
🐛 Fix bug when cluster is not found

maelk:
And remove the restrictions on the events triggering the reconciliation
frommetal3dataclaim to catch update events when the deletion timestamp is
set. It also improves slightly the makefile

### DIFF
--- a/controllers/metal3datatemplate_controller.go
+++ b/controllers/metal3datatemplate_controller.go
@@ -29,11 +29,8 @@ import (
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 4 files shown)


controllers/metal3datatemplate_controller.go
```diff
@@ -110,14 +107,16 @@ func (r *Metal3DataTemplateReconciler) Reconcile(req ctrl.Request) (_ ctrl.Resul
...
 
-	metadataLog = metadataLog.WithValues("cluster", cluster.Name)
-	if err := metadataMgr.SetClusterOwnerRef(cluster); err != nil {
-		return ctrl.Result{}, err
-	}
-	// Return early if the Metadata or Cluster is paused.
-	if util.IsPaused(cluster, capm3DataTemplate) {
-		metadataLog.Info("reconciliation is paused for this object")
-		return ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil
+	if capm3DataTemplate.Spec.ClusterName != "" && cluster.Name != "" {
```
furkatgofurov7, line 110: Why are we not checking this `cluster != nil` condition here?

maelk: because the cluster is already defined, just with empty values, so cluster would never be nil

PR #99, opened 2020-06-15
🏃 Update test scripts

maelk:
Adds golint.sh and codegen.sh and make gofmt.sh fail if files
are changed

### DIFF
--- a/hack/codegen.sh
+++ b/hack/codegen.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+
+set -eux
+
+IS_CONTAINER="${IS_CONTAINER:-false}"
+ARTIFACTS="${ARTIFACTS:-/tmp}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+  export XDG_CACHE_HOME=/tmp/.cache
+  eval "$(go env)"
+  cd "${GOPATH}"/src/github.com/metal3-io/cluster-api-provider-metal3
+  INPUT_FILES="\
+  config/certmanager/*.yaml
+  config/crd/*.yaml
+  config/crd/bases/*.yaml
+  config/crd/patches/*.yaml
+  config/default/*.yaml
+  config/manager/*.yaml
+  config/rbac/*.yaml
+  config/webhook/*.yaml
+  config/*.yaml
+  api/v1alpha2/zz_generated.*.go
+  api/v1alpha3/zz_generated.*.go
+  baremetal/mocks/zz_generated.*.go"
+
+  # shellcheck disable=SC2086
+  cksum $INPUT_FILES > "$ARTIFACTS/lint.cksums.before"
+  export VERBOSE="--verbose"
+  make generate
+  # shellcheck disable=SC2086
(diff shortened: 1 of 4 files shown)


hack/codegen.sh
```diff
@@ -0,0 +1,42 @@
...
+set -eux
+
+IS_CONTAINER=${IS_CONTAINER:-false}
+ARTIFACTS=${ARTIFACTS:-/tmp}
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+  export XDG_CACHE_HOME=/tmp/.cache
+  eval "$(go env)"
+  cd "${GOPATH}"/src/github.com/metal3-io/cluster-api-provider-metal3
```
fmuyassarov, line 12: Don't you need to export `GOPATH`?

hack/codegen.sh
```diff
@@ -0,0 +1,42 @@
...
+  INPUT_FILES="\
+  config/certmanager/*.yaml
+  config/crd/*.yaml
+  config/crd/bases/*.yaml
+  config/crd/patches/*.yaml
+  config/default/*.yaml
+  config/manager/*.yaml
+  config/rbac/*.yaml
+  config/webhook/*.yaml
+  config/*.yaml
```
fmuyassarov, line 22: I don't see `/bmo`. I'm not sure if it also needs to be added or not?

hack/codegen.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+set -eux
+
+IS_CONTAINER=${IS_CONTAINER:-false}
+ARTIFACTS=${ARTIFACTS:-/tmp}
```
fmuyassarov, line 6: Double quotes around curly brackets?

hack/golint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+set -eux
+
+IS_CONTAINER=${IS_CONTAINER:-false}
```
fmuyassarov, line 5: Double quotes around curly brackets?

hack/codegen.sh
maelk: it's defined as part of line 11

hack/codegen.sh
maelk: this is v1a3, there is no BMO, and no ipam

hack/codegen.sh
maelk: I can add that indeed

hack/golint.sh
maelk: ok

hack/codegen.sh
furkatgofurov7: Do we need double quotes here? Check [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/036ea1866c391e8d22213cfc7c0ebba597292b1d/hack/codegen.sh#L5)

hack/golint.sh
furkatgofurov7: Check [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/036ea1866c391e8d22213cfc7c0ebba597292b1d/hack/golint.sh#L5) as well

PR #103, opened 2020-06-18
🏃 Update v1a3 CRDs to add image checksum type and disk format

maelk:
This PR adds the image checksum type and disk format in v1a3 CRDs, as it was added in release-0.3 branch.
This also limits the disk formats accepted to match BMO's.


maelk:
This requires https://github.com/metal3-io/metal3-dev-env/pull/346 to be able to run the integration tests

PR #104, opened 2020-06-22
Download libs of mod files

digambar15:
Download all libs from mod files as part of Dockerfile.unit.


hack/Dockerfile.unit
```diff
@@ -1,6 +1,7 @@
 FROM registry.hub.docker.com/library/golang:1.13
 
-WORKDIR /usr/local/kubebuilder
 COPY /hack/tools /usr/local/kubebuilder/
+WORKDIR /usr/local/kubebuilder
 RUN ./install_kustomize.sh && \
     ./install_kubebuilder.sh
+RUN go mod download
```
maelk, line 7: you are not copying the go.mod and go.sum files from the root folder here, so the go mod download is only downloading the modules that are in /hack/tools/go.mod, and not the go.mod ones. So you need to add a copy of the top level go.mod and go.sum and then run go mod download with thoses, outside of the kubebuilder folder to avoid conflicts

digambar15: Done!

hack/Dockerfile.unit
```diff
@@ -1,6 +1,10 @@
 FROM registry.hub.docker.com/library/golang:1.13
 
+COPY . /usr/local/kubebuilder/
```
maelk, line 3: I think you should not place this in /usr/local/kubebuilder. maybe in /CAPM3 and you delete that folder afterwards ?

hack/Dockerfile.unit
```diff
@@ -1,6 +1,10 @@
...
 WORKDIR /usr/local/kubebuilder
-COPY /hack/tools /usr/local/kubebuilder/
+RUN go mod download -json
+
+WORKDIR /usr/local/kubebuilder/hack/tools
```
maelk, line 7: you should set the workdir once only. for other needs, just use cd in your RUN commands

hack/Dockerfile.unit
digambar15: okay

hack/Dockerfile.unit
digambar15: okay

hack/Dockerfile.unit
```diff
@@ -1,6 +1,10 @@
 FROM registry.hub.docker.com/library/golang:1.13
 
-WORKDIR /usr/local/kubebuilder
-COPY /hack/tools /usr/local/kubebuilder/
+COPY .  /CAPM3
+RUN cd /CAPM3 && go mod download -json
+RUN cp -r /CAPM3/hack/tools /usr/local/kubebuilder/
+RUN cd .. && rm -rf CAPM3
```
maelk, line 6: please try to limit the number of layers that you will be generating. All this can be merged into a single RUN.

hack/Dockerfile.unit
```diff
@@ -1,6 +1,10 @@
...
+WORKDIR /usr/local/kubebuilder/
 RUN ./install_kustomize.sh && \
     ./install_kubebuilder.sh
+RUN go mod download -json
```
maelk, line 10: same here

hack/Dockerfile.unit
digambar15: okay

hack/Dockerfile.unit
digambar15: will optimize it

maelk:
/remove-lifecycle stale
This is still valid. We just need to figure out what is best for the CI. Maybe @russellb  could have some inputs there. The choice is between downloading all modules again (time + badnwidth requirements) vs size of the image.

maelk:
/remove-lifecycle stale
Now that the prow cluster will be moved to Nordix, this will need to be reconsidered. 
/cc @fmuyassarov 
/cc @kashifest

PR #105, opened 2020-06-28
🐛 Add an annotation with BMH status when pausing it

maelk:
In order to properly pivot the BMH, we need to set two annotations
on the BMH, one to pause the reconciliation, the other with the
marshalled status. This commit adds the second annotation as part
of the pausing workflow.


dhellmann:
This approach looks good. If there is a conflict writing the host here, it will retry and get the new status. If the BMO is reconciling the host at the same time it will get a conflict, try again, and see the pause annotation.

/approve

PR #108, opened 2020-07-07
🏃 Update Baremetal Operator for the new unmanaged state

maelk:
This adds support for a new state defined in BMO : https://github.com/metal3-io/baremetal-operator/pull/571


dhellmann:
This isn't sufficient by itself. We can't deprovision the host without credentials, so if we have an unmanaged host linked to a machine then we won't be able to delete the machine. To fix that, we also need to update https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L568 to include the new state as one for which we no longer need to wait before deleting the machine.

PR #109, opened 2020-07-07
🏃 Update Baremetal Operator for the new unmanaged state

maelk:
This PR adds support for the new state defined in https://github.com/metal3-io/baremetal-operator/pull/571


dhellmann:
See the comment I left on #108 about the additional work needed.

PR #113, opened 2020-07-14
🐛 Force deletion of BMH attributes when deleting m3m

maelk:

When deleting the Metal3machine, the userData used to provision might
already be deleted and absent from the status of the metal3machine,
causing the metal3machine not to be deleted since the BMH still has the
userData set. This patch forces the deletion of the userData and similar.


jan-est:
lgtm

smoshiur1237:
lgtm

PR #115, opened 2020-07-15
:running: update baremetalhost crd to include boot mode field

dhellmann:

Update the version of BMO we use so we get the newer BareMetalHost with
the additional field for the boot mode. This prevents the controller from
overwriting the host with partial data and losing the user-configured
value for booting the host.

Implements https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/implicit-boot-mode.md

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/golang/mock v1.4.3
 	github.com/google/gofuzz v1.1.0
 	github.com/mailru/easyjson v0.7.1 // indirect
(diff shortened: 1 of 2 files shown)


maelk:
This will be needed for the other branch too (release-0.3)
/lgtm
/test-v1a4-integration

dhellmann:
> This will be needed for the other branch too (release-0.3)

See #116

dhellmann:
> > This will be needed for the other branch too (release-0.3)
> 
> See #116

Err, #117

PR #119, opened 2020-07-18
🏃 Update namings in images

furkatgofurov7:
Updates old namings in images

Relevant to #118


furkatgofurov7:
> Just tiny comment, is it still possible to rewrite CAPM³ to CAPM3 ?

@fmuyassarov I was following [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/images/components.png) and [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/getting-started.md) to keep consistency, in that case we have to change these as well to CAPM3? I can rewrite CAPM3 only in my patch as well.

furkatgofurov7:
> We should maybe agree on the spelling of the acronym. Until now we have been using more CAPM3 I would say

Agree, [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/api.md) we have `CAPM3` in other places `CAPM³`

fmuyassarov:
> > Just tiny comment, is it still possible to rewrite CAPM³ to CAPM3 ?
> 
> @fmuyassarov I was following [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/images/components.png) and [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/getting-started.md) to keep consistency, in that case we have to change these as well to CAPM3? I can rewrite CAPM3 only in my patch as well.

In those docs, it is only Metal³, but not CAPM³. So, as you said for the sake of consistency maybe we could use CAPM3?

furkatgofurov7:
> > > Just tiny comment, is it still possible to rewrite CAPM³ to CAPM3 ?
> > 
> > 
> > @fmuyassarov I was following [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/images/components.png) and [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/docs/getting-started.md) to keep consistency, in that case we have to change these as well to CAPM3? I can rewrite CAPM3 only in my patch as well.
> 
> In those docs, it is only Metal³, but not CAPM³. So, as you said for the sake of consistency maybe we could use CAPM3?

Sure, fixed

PR #126, opened 2020-08-10
🐛 Fix shell parameters expansion for CAPI deployment

### LINKED ISSUE #125: CAPI isn't deployed correctly after "make deploy"
I use fresh K8s cluster deployed with GCP Kubernetes engine
Performing **make deploy** in fresh cloned metal3-io/cluster-api-provider-metal3 repo
results in
```
capi-kubeadm-bootstrap-controller-manager-5bdc7dbf5d-d5ml5     CrashLoopBackOff
capi-controller-manager-7894bbf65d-hq9xc                       CrashLoopBackOff
capi-controller-manager-6dd9f648cf-zf2hw                       CrashLoopBackOff
capi-kubeadm-bootstrap-controller-manager-867f46d767-k4862     CrashLoopBackOff
```

```
kubectl logs capi-controller-manager-7894bbf65d-hq9xc manager -n capi-system
invalid argument "MachinePool=${EXP_MACHINE_POOL:=false},ClusterResourceSet=${EXP_CLUSTER_RESOURCE_SET:=false}" for "--feature-gates" flag: invalid value of MachinePool=${EXP_MACHINE_POOL:=false}, err: strconv.ParseBool: parsing "${EXP_MACHINE_POOL:=false}": invalid syntax

```
CAPI controllers up and running

[Miscellaneous information that will assist in solving the issue.]
Probably it's related to https://github.com/kubernetes-sigs/cluster-api/issues/3339
but I don't understand what is exactly wrong in my case. I don't see any clusterctl uses during **make deploy**

**Environment:**

- Cluster-api version: 0.3.7
- CAPBM version:
- Minikube version:
- environment (metal3-dev-env or other): GCP Kubernetes engine
- Kubernetes version: (use `kubectl version`): v1.17.8-gke.17

/kind bug

DmitryGrayscale:
In CAPI v0.3.7+ yaml files expressions like ${PARAM:=DEFAULTVALUE} were added.
GNU envsubst can't handle such kind of expressions, so enhnaced
implementation of envsubst were used to overcome this.

Fixes #125

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -114,16 +114,20 @@ echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 # Get Cert-manager provider components file
 curl -L -o "${COMPONENTS_CERT_MANAGER_GENERATED_FILE}" https://github.com/jetstack/cert-manager/releases/download/v0.13.0/cert-manager.yaml
 
+# Get enhanced envsubst version to evaluate expressions like ${VAR:=default}
+curl -L -o "${SOURCE_DIR}"/envsubst-go https://github.com/a8m/envsubst/releases/download/v1.2.0/envsubst-"$(uname -s)"-"$(uname -m)"
+chmod +x "${SOURCE_DIR}"/envsubst-go
+
 # Generate Cluster API provider components file.
-"$KUSTOMIZE" build "github.com/kubernetes-sigs/cluster-api/config/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+"$KUSTOMIZE" build "github.com/kubernetes-sigs/cluster-api/config/?ref=master" | "${SOURCE_DIR}"/envsubst-go > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
 echo "Generated ${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
 
 # Generate Kubeadm Bootstrap Provider components file.
-"$KUSTOMIZE" build "github.com/kubernetes-sigs/cluster-api/bootstrap/kubeadm/config/?ref=master" > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
+"$KUSTOMIZE" build "github.com/kubernetes-sigs/cluster-api/bootstrap/kubeadm/config/?ref=master" | "${SOURCE_DIR}"/envsubst-go > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
 echo "Generated ${COMPONENTS_KUBEADM_GENERATED_FILE}"
 
 # Generate Kubeadm Controlplane components file.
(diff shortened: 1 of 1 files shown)


examples/generate.sh
```diff
@@ -114,16 +114,20 @@ echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 # Get Cert-manager provider components file
 curl -L -o "${COMPONENTS_CERT_MANAGER_GENERATED_FILE}" https://github.com/jetstack/cert-manager/releases/download/v0.13.0/cert-manager.yaml
 
+# Get enhanced envsubst version to evaluate expressions like ${VAR:=default}
+curl -L -o envsubst-go https://github.com/a8m/envsubst/releases/download/v1.2.0/envsubst-`uname -s`-`uname -m`
+chmod +x envsubst-go
```
maelk, line 119: Should this be in a specific folder ? like a tmp, or the examples folder, rather than where this script is executed from ?

examples/generate.sh
```diff
@@ -114,16 +114,20 @@ echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
...
+
 # Generate Cluster API provider components file.
-"$KUSTOMIZE" build "github.com/kubernetes-sigs/cluster-api/config/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+"$KUSTOMIZE" build "github.com/kubernetes-sigs/cluster-api/config/?ref=master" | ./envsubst-go > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
```
maelk, line 122: most variables in the manifests are not defined when we call envsubst. You could define a set of default variable used just to populate the example manifests.

DmitryGrayscale: Yes, it should be like that :)
By the way, do you think that replacing "vanilla" envsubst entirely with the "enhanced" version could be a thing there?
With the PR we will use two different envsubst when one of them just supports both use cases.

examples/generate.sh
DmitryGrayscale: Yes, it should be like that :)
By the way, do you think that replacing "vanilla" envsubst entirely with the "enhanced" version could be a thing there?
With the PR we will use two different envsubst when one of them just supports both use cases.

examples/generate.sh
DmitryGrayscale: I didn't quite understood you probably, added the following just in case 
https://github.com/metal3-io/cluster-api-provider-metal3/pull/126/commits/6bbfb1b8d5d90b43fff5df66e8ea13c99ac1d328#diff-2b085cd4a651deec0d4f1b1850814f02R43-R44

maelk: I didn't see that they are now with a default within the manifests, sorry, my bad.

DmitryGrayscale: yes, I will remove default values from the generate.sh then

PR #135, opened 2020-09-22
📖  Add note about node<--->bmh mapping after pivoting

### LINKED ISSUE #70: Node.Spec.ProviderID not updated after Move
After a clusterctl move from say a bootstrap cluster to a target cluster, the K.Node.Spec.ProviderID still contains the UID of the old BMH object. 

Simply updating the Node.Spec.ProviderID is not currently allowed as validation checks prevent Node.Spec.ProviderID to be changed after it has been set to a non-nil value (see https://github.com/kubernetes/kubernetes/pull/51761) 
	```Forbidden: node updates may not change providerID except from "" to valid, []: ```

Possible Solutions:
	1) Provide an option to have ProviderID be something other than UID of BMH -- for example: BMH 'Namespace/Name' may be unique for most environments. 
	2) Don’t use BMH.UID which is generated automatically during object creation but instead create a new field (say BMH.ProviderID) that we can populate with a random id. This field would be copied over as part of the move.
	3) Open an issue with K/K to allow updates to Node.Spec.ProviderID -- ie. Relax the validation checks.

/kind bug

xref: https://github.com/metal3-io/baremetal-operator/issues/431
@kashifest @maelk

furkatgofurov7:
This PR adds note regarding mapping K.Node <--> BMH after pivoting
 
Fixes #70

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -267,3 +267,10 @@ a target cluster upgrade:
 - Ironic should always be the first component upgraded, before a CAPM3 / BMO
   upgrade using clusterctl for example, or before nodes upgrades. This is to
   ensure that the cluster is in a stable condition while upgrading Ironic.
+
+**Important Note:**
+Currently, when target cluster is up and node appears, CAPM3 will fetch the node
+and set the providerID value to BMH UUID, meaning that it is not advisable to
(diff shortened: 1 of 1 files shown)


maelk (approved):

fmuyassarov (commented):
/lgtm

PR #136, opened 2020-10-06
🌱 Modify Metal3Machine association order

furkatgofurov7:
​This PR waits for a BMH to be available and associated before creating the Metal3DataClaim. This ensures that the previous node
is properly deleted before giving an index to the new one (in case of scale-in upgrade).


maelk (approved):

Jaakko-Os:
looks good to me

PR #138, opened 2020-10-09
:bug: Add missing namespace ironic configmap

kashifest:
Fixes # Add missing namespace ironic configmap


maelk:
this is in deadlock with metal3-dev-env PR. Merging manually

PR #149, opened 2020-12-02
✨ Add manifest linting script

fmuyassarov:
Add a script for validation of the k8s core objects within the repo. This is the first step to have a mandatory CI job on every future pull request. The second step will be to configure the prow to run the script.

There are some files such as kustomization patches - for which we need to skip the validation because their deployment definitions are not complete. As such, kubeval will through an error for them. However, we can give a pattern of the files in the same directory that need to be skipped from the check. That's why we add the word `patch` in all the kustomization patches so that we can filter them out with a single pattern, which in this case is patch.


kashifest (approved):

furkatgofurov7 (commented):
Looks good to me. Are you planning to add this to the Makefile target, in this repo as well?

fmuyassarov:
> Looks good to me. Are you planning to add this to the Makefile target, in this repo as well?

Yes, added.
/test-integration

PR #150, opened 2020-12-07
:seedling: apply our own rules for host availability

dhellmann:
Do not rely on the host's Available() method to decide when a host can
be selected for provisioning.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -788,7 +788,7 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
 		if host.GetDeletionTimestamp() != nil {
 			continue
 		}
-		if host.HasError() {
+		if host.Status.ErrorMessage != "" {
 			continue
 		}
 		switch host.Status.Provisioning.State {
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -782,7 +782,13 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
 			helper, err := patch.NewHelper(&hosts.Items[i], m.client)
 			return &hosts.Items[i], helper, err
 		}
-		if !host.Available() {
+		if host.Spec.ConsumerRef != nil {
+			continue
+		}
+		if host.GetDeletionTimestamp() != nil {
+			continue
+		}
+		if host.HasError() {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -782,7 +782,13 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
...
 			return &hosts.Items[i], helper, err
 		}
-		if !host.Available() {
+		if host.Spec.ConsumerRef != nil {
+			continue
+		}
+		if host.GetDeletionTimestamp() != nil {
+			continue
+		}
+		if host.HasError() {
```
zaneb, line 791: Maybe we shouldn't use this convenience method either.

zaneb (commented):
/approve

baremetal/metal3machine_manager.go
dhellmann: Done

s3rj1k:
@dhellmann what about ClearError() SetErrorMessage() ?

dhellmann:
> @dhellmann what about ClearError() SetErrorMessage() ?

I added another patch to the series to remove those.

I think we can deal with the other public methods separately, to avoid making this PR gigantic.

s3rj1k:
@dhellmann Do you intend to make a stable API promise?

I mean the only reason that I can think of removing those methods and make them private is to prepare a codebase for a new API version release.

dhellmann:
> @dhellmann Do you intend to make a stable API promise?
> 
> I mean the only reason that I can think of removing those methods and make them private is to prepare a codebase for a new API version release.

I see this as part of continual cleanup. Those methods shouldn't have been implemented the way they are to begin with. I don't really think it's necessary to commit to making them a stable API.

We do try to keep other parts of the code stable, like github.com/metal3-io/baremetal-operator/pkg/bmc, but even in that case it's possible to address changes in the API when a dependency is updated in a consuming project.

Are you anticipating a specific issue from these changes? Or just concerned that we don't consider "internals" as needing to have stable APIs?

s3rj1k:
> Are you anticipating a specific issue from these changes? Or just concerned that we don't consider "internals" as needing to have stable APIs?

Just concerned that some of the `internals` could actually be used by external projects, good example would be HasError() method.

On a side note not directly connected to this PR, can we have more frequent tag releases for code under `metal3-io` umbrella?
This should help with vendoring.

maelk (commented):
/lgtm

PR #153, opened 2021-01-11
:sparkles: Update go.mod

kashifest:


maelk:
/approve
This goes back in the K8S version to be able to support the latest CAPI version that requires controller runtime 0.5.14 (instead of the 0.6.X we were using before). When we will take the v1a4 CAPI we will be able to properly uplift controller-runtime dependencies and match the kubernetes version.

PR #155, opened 2021-01-27
Parameterize docker registry

Xenwar:
This PR parameterizes the docker registry for building images.
Which registry is used determined by the following variable, with the give default value.
Exporting the variables per Dockerfile allows this repo to be independent of metal3-dev-env
This has an effect when using Makefile and the value is overridden.

export OVERRIDE_DOCKER_HUB_REGISTRY=${OVERRIDE_DOCKER_HUB_REGISTRY:-"registry.hub.docker.com"}

If you wish to use a different registry, please assign value in a HOST:PORT/path format.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 # Build the manager binary on golang image
-FROM registry.hub.docker.com/library/golang:1.15.3 as builder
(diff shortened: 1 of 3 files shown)


smoshiur1237:
LGTM

Xenwar:
replaced by an alternative solution.

PR #157, opened 2021-02-05
✨ Add new Remediation Controller to CAPM3

jan-est:
This PR implements [capm3-remediation-controller proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/capm3-remediation-controller-proposal.md). 

This PR calls for a new `Metal3Remediation` object and a new controller in CAPM3 that uses BMO APIs to perform automated remediation of unhealthy nodes in Metal3 project.

This proposal also introduces Metal3RemediationTemplate CRD. The Metal3RemediationTemplate is used by CAPI MachineHealthCheck to create Metal3Remediation objects when it detects unhealthy Machines in the cluster.


api/v1alpha4/metal3remediation_types.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+type RemediationType string
+
+const (
+	// RemediationFinalizer allows Metal3RemediationReconciler to clean up resources associated with Metal3Remediation before
+	// removing it from the apiserver.
+	RemediationFinalizer = "metal3remediation.infrastructure.cluster.x-k8s.io"
+
+	// ScaleDownRolloutType constant is to set rollout strategy type to scale down.
```
n1r1, line 30: The doc text seems to be unrelated?

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,219 @@
...
+			remediationLog.Error(err, "failed to Patch metal3Remediation")
+		}
+	}()
+
+	// Fetch the Machine.
+	capiMachine, err := util.GetOwnerMachine(ctx, r.Client, metal3Remediation.ObjectMeta)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "metal3Remediation's owner Machine could not be retrieved")
+	}
+	remediationLog = remediationLog.WithValues("machine", capiMachine.Name)
```
fmuyassarov, line 80: Does `metal3Remediation` has the `ownerReference` pointing to a Machine? I thought `ownerReference` would be pointing to the CAPI `machineHealthCheck`.

jan-est: Yes, it does have `ownerRefence` of the unhealthy `Machine`. There is no point to have `machineHealthCheck` ownerRef.

fmuyassarov: In the [proposal](https://github.com/metal3-io/metal3-docs/blob/master/design/capm3-remediation-controller-proposal.md#metal3remediation-crd) I see the [ownerRefence](url) to the `MachineHealthCheck`

```
apiVersion: remediation.metal3.io/v1alphaX
kind: Metal3Remediation
metadata:
  name: NAME_OF_UNHEALTHY_MACHINE
  namespace: NAMESPACE_OF_UNHEALTHY_MACHINE
  ownerReferences:
   - apiVersion: cluster.x-k8s.io/v1alphaX
     kind: MachineHealthCheck
     name: REMEDIATION_GROUP
spec:
  type: reboot
  rebootLimit: 3
  timeout: 300s
status:
  phase: rebooting
  retryCount: 1

```

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// If user has set bmh.Spec.Online to false or annotated the host with any poweroff annotations
+	// do not try to remediate the host
+	if !remediationMgr.OnlineStatus(host) || remediationMgr.HasRebootAnnotation(host) {
+		r.Log.Info("Host is powered off, unable to remediatiate")
+	}
+
+	// If the Metal3Remediation doesn't have finalizer, add it.
+	// This is for the future work, if finalizers are needed
+	//remediationMgr.SetFinalizer()
```
fmuyassarov, line 128: Why is this commented out?

jan-est: Currently we don't use finalizers but the functions are implemented. This is for the future work but, maybe it is better to remove it here.

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,219 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *Metal3RemediationReconciler) reconcileDelete(ctx context.Context,
+	remediationMgr baremetal.RemediationManagerInterface,
+) (ctrl.Result, error) {
+
+	// metal3remediation is marked for deletion and ready to be deleted,
+	// so remove the finalizer.
+	// remediationMgr.UnsetFinalizer()
```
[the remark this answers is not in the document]
jan-est: Currently we don't use finalizers but the functions are implemented. This is for the future work but, maybe it is better to remove it here.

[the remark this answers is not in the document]
fmuyassarov: Yes, I think it is better if we remove it, if there is no use of it.

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,219 @@
...
+			okToStop, nextCheck := remediationMgr.TimeToRemediate(remediationMgr.GetTimeout().Duration)
+
+			if okToStop {
+				remediationMgr.SetRemediationPhase(capm3.PhaseDeleting)
+				// When machine is still unhealthy after remediation, setting of OwnerRemediatedCondition
+				// moves control to CAPI machine controller. The owning controller will do
+				// preflight checks and handles the Machine deletion
+				err = remediationMgr.SetOwnerRemediatedCondition(ctx)
+				if err != nil {
+					return ctrl.Result{}, errors.Wrapf(err, "error setting condition")
```
n1r1, line 183: missing log

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,219 @@
...
+				}
+
+				// Remediation failed set unhealthy annotation on BMH
+				// This prevents BMH to be selected as a host.
+				err = remediationMgr.SetAnnotation(ctx, "unhealthy")
+				if err != nil {
+					return ctrl.Result{}, errors.Wrapf(err, "error setting unhealthy annotation")
```
n1r1, line 190: missing log

controllers/metal3remediation_controller.go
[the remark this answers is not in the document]
n1r1: WDYT about removing the `unhealthy` annotation from the host once the remediation request is being deleted?

n1r1 (commented):
Overall looks good.
Left some comments

controllers/metal3remediation_controller.go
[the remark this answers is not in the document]
jan-est: Do we really want to remove the `unhealthy` annotation? If our host is not healthy and possible needs manual maintenance annotation will make sure the host is not getting provisioned before annotation is removed manually.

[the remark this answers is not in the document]
n1r1: if `metal3remediation` was deleted, it means that the host is healthy again, isn't it?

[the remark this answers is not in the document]
jan-est: That is true in cases where MHC find out Machine healthy again and deletes the metal3remediation. 

But when RC tryies to reboot host n times and timeouts after last retry, it will mark BMH as unhealthy and set condition on Machine for deletion. So in latter case we are not sure if host is healthy or not and is it capable of serving as a host for any new Machine objects.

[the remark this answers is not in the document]
n1r1: Good point.
So before removing the finalizer we can check the `metal3remediation` CR.
If remediation is failed and machine is going to be deleted we won't remove the unhealthy annotation.
Does that make sense?

controllers/metal3remediation_controller.go
jan-est: I would leave logging out from reconcile loop right now. But we may need to thing later, how we would like to do logging in the future. Since we are talking about remediation it would be good to have solid logging, but we might need to thing this over at some point.

baremetal/metal3remediation_manager.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// GetLastRemediatedTime returns last remediation time
+func (r *RemediationManager) GetLastRemediatedTime() *metav1.Time {
+	return r.Metal3Remediation.Status.LastRemediated
+}
+
+// SetLastRemediationTime sets the last remediation timestamp on Status
+func (r *RemediationManager) SetLastRemediationTime(remediationTime *metav1.Time) {
+	r.Log.Info("Last remediation time", "remediationTime", remediationTime)
```
n1r1, line 233: s/Last/setting last

baremetal/metal3remediation_manager.go
```diff
@@ -0,0 +1,260 @@
...
+
+// IncreaseRetryCount increases the retry count on Status
+func (r *RemediationManager) IncreaseRetryCount() {
+	r.Metal3Remediation.Status.RetryCount++
+}
+
+func (r *RemediationManager) SetOwnerRemediatedCondition(ctx context.Context) error {
+	machineHelper, err := patch.NewHelper(r.Machine, r.Client)
+	if err != nil {
+		r.Log.Info("Unable to create patch helper for Machine")
```
n1r1, line 250: we should probably change log level to error and log the `err`

baremetal/metal3remediation_manager.go
```diff
@@ -0,0 +1,260 @@
...
+		return err
+	}
+	conditions.MarkFalse(r.Machine, capi.MachineOwnerRemediatedCondition, capi.WaitingForRemediationReason, capi.ConditionSeverityWarning, "")
+	err = machineHelper.Patch(ctx, r.Machine)
+	if err != nil {
+		r.Log.Info("Unable to patch Machine %d", r.Machine)
```
n1r1, line 256: we should probably change log level to error and log the `err`

baremetal/metal3remediation_manager.go
n1r1: also could be good to add the machine name

config/rbac/role.yaml
```diff
@@ -52,6 +52,26 @@ rules:
...
+  - cluster.x-k8s.io
+  resources:
+  - machine
+  verbs:
+  - create
+  - delete
+  - list
+  - patch
+  - update
+  - verbs=get
```
n1r1, line 65: looks like an artificat. should be just "get"

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Fetch the Metal3Remediation instance.
+	metal3Remediation := &capm3.Metal3Remediation{}
+
+	if err := r.Client.Get(ctx, req.NamespacedName, metal3Remediation); err != nil {
+		if apierrors.IsNotFound(err) {
+			remediationLog.Error(err, "unable to get metal3Remediation")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
```
n1r1, line 62:
```suggestion
		if apierrors.IsNotFound(err) {
  		        return ctrl.Result{}, nil
		}
		remediationLog.Error(err, "unable to get metal3Remediation")
		return ctrl.Result{}, err
		
```

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	helper, err := patch.NewHelper(metal3Remediation, r.Client)
+	if err != nil {
+		remediationLog.Error(err, "failed to init patch helper")
+		return ctrl.Result{}, errors.Wrap(err, "failed to init patch helper")
```
n1r1, line 67: I think we can just return the `err` as is.
The err will be used to trigger a requeue, and we already logged the detailed message. 
```suggestion
		remediationLog.Error(err, "failed to init patch helper")
		return ctrl.Result{}, err
```

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	// Always patch metal3Remediation when exiting this function so we can persist any metal3Remediation changes.
+	defer func() {
+		err := helper.Patch(ctx, metal3Remediation)
```
n1r1, line 71: `helper` might be undefined if the function returns before it's being declared, e.g. line 60

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			remediationLog.Error(err, "failed to Patch metal3Remediation")
+		}
+	}()
+
+	// Fetch the Machine.
+	capiMachine, err := util.GetOwnerMachine(ctx, r.Client, metal3Remediation.ObjectMeta)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "metal3Remediation's owner Machine could not be retrieved")
```
n1r1, line 80: missing error log

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	// Fetch Metal3Machine
+	metal3Machine := capm3.Metal3Machine{}
+	key := client.ObjectKey{
+		Name:      capiMachine.Spec.InfrastructureRef.Name,
+		Namespace: capiMachine.Spec.InfrastructureRef.Namespace,
+	}
+	err = r.Get(ctx, key, &metal3Machine)
+	if err != nil {
+		remediationLog.Error(err, "metal3machine not found")
+		return ctrl.Result{}, errors.Wrapf(err, "metal3machine not found")
```
n1r1, line 93:
```suggestion
		err = remediationLog.Error(err, "metal3machine not found")
		return ctrl.Result{}, err
```

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	// If host is gone, exit early
+	host, _, err := remediationMgr.GetUnhealthyHost(ctx)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "unable to find a host for unhealthy machine")
+	}
+
+	// If user has set bmh.Spec.Online to false
+	// do not try to remediate the host
+	if !remediationMgr.OnlineStatus(host) {
+		r.Log.Info("Host is powered off, unable to remediate")
```
n1r1, line 127: I think it might be easier for someone to understand if we log something like "Host is powered off (`spec.Online` is `false`).
also, could be good to add the host name

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+		case capm3.PhaseRunning:
+			// host is not rebooted yet
+			if remediationMgr.GetLastRemediatedTime() == nil {
+				r.Log.Info("Rebooting the host")
+				err := remediationMgr.SetRebootAnnotation(ctx)
+				if err != nil {
+					return ctrl.Result{}, errors.Wrap(err, "error setting reboot annotation")
+				}
+				now := metav1.Now()
+				remediationMgr.SetLastRemediationTime(&now)
```
n1r1, line 155: if we will fail to patch the metal3remediation CR, we will reboot the host repeatedly until this succeeds.
I'm wondering if this is something we should bother to fix or leave it as is

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+				if err != nil {
+					return ctrl.Result{}, errors.Wrapf(err, "error setting condition")
+				}
+
+				// Remediation failed set unhealthy annotation on BMH
+				// This prevents BMH to be selected as a host.
+				err = remediationMgr.SetUnhealthyAnnotation(ctx)
+				if err != nil {
+					return ctrl.Result{}, errors.Wrapf(err, "error setting unhealthy annotation")
+				}
```
n1r1, line 198: for the sake of prevent future errors it might be good to return here

examples/metal3plane/hosts.yaml
```diff
@@ -19,6 +19,7 @@ spec:
     credentialsName: demo-externally-provisioned-secret
 status:
   errorMessage: ""
+  errorCount: 0
```
n1r1, line 22: can you please elaborate what's the purpose of this change?

go.mod
```diff
@@ -63,5 +64,7 @@ replace k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.17.9
 replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.5.14
 
 replace github.com/go-logr/zapr => github.com/go-logr/zapr v0.2.0
+
 replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-bcb3869e6f29
-replace github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.0
\ No newline at end of file
+
+replace github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.0
```
n1r1: seems like there's no actual change here so I'd revert formatting changes

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	remediationType := remediationMgr.GetRemediationType()
+
+	if remediationType != capm3.RebootRemediationStrategy {
+		r.Log.Info("unsupported remediation strategy")
+		return ctrl.Result{}, nil
+	}
+
+	if remediationType == capm3.RebootRemediationStrategy {
```
n1r1, line 139: I believe this line can be removed as we already checked this in line 134

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,222 @@
...
+				okToRemediate, nextRemediation := remediationMgr.TimeToRemediate(remediationMgr.GetTimeout().Duration)
+
+				if okToRemediate {
+					err := remediationMgr.SetRebootAnnotation(ctx)
+					if err != nil {
+						return ctrl.Result{}, errors.Wrapf(err, "error setting reboot annotation")
+					}
+					now := metav1.Now()
+					remediationMgr.SetLastRemediationTime(&now)
+					remediationMgr.IncreaseRetryCount()
```
n1r1, line 169: looks like this code block appears twice so it might be worth to extract it into a function

api/v1alpha4/metal3remediationtemplate_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+var (
+	// Default retry timeout is 100 seconds.
+	defaultTimeout = metav1.Duration{Duration: 100 * time.Second}
```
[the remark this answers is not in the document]
jan-est: It is quite hard to know what would be the good timeout for actual hardware. Do you think 10 minutes would be enough?

controllers/metal3remediation_controller.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to init patch helper")
+	}
+	// Always patch metal3Remediation when exiting this function so we can persist any metal3Remediation changes.
+	defer func() {
+		err := helper.Patch(ctx, metal3Remediation)
+		if err != nil {
+			remediationLog.Error(err, "failed to Patch metal3Remediation")
+		}
+	}()
```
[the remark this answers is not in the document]
jan-est: This is common in every capm3 controller. I don't see big performance issue here.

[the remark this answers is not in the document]
n1r1: unless the Patch functions can tell if there was an actual change, it will send a PATCH request to the api-server for every reconcile, even if the object hasn't really changed.
Since this is a common pattern here, let's leave it as is :+1:

controllers/metal3remediation_controller.go
jan-est: Let's leave this as it is.

controllers/metal3remediation_controller.go
jan-est: Should we keep this in place if we wanna support multiple remediation strategies in the future.

api/v1alpha4/metal3remediation_types.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type RemediationType string
+
+const (
+	// RemediationFinalizer allows Metal3RemediationReconciler to clean up resources associated with Metal3Remediation before
+	// removing it from the apiserver.
+	RemediationFinalizer = "metal3remediation.infrastructure.cluster.x-k8s.io"
```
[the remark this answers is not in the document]
jan-est: All other finalizer used in CAPM3 are defined like this. For example one used with metal3machines https://github.com/metal3-io/cluster-api-provider-metal3/blob/8008a8017021ed7a70489900147b909bdae7319e/api/v1alpha4/metal3machine_types.go#L30

baremetal/metal3remediation_manager.go
```diff
@@ -0,0 +1,260 @@
...
+	host, helper, err := r.GetUnhealthyHost(ctx)
+	if err != nil {
+		return err
+	}
+	if host == nil {
+		return nil
+	}
+
+	r.Log.Info("Adding Reboot annotation to BareMetalHost")
+	host.Annotations[rebootAnnotation] = rebootAnnotation
```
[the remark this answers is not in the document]
jan-est: Should the user be able to select rebootMode? If not I would trust what ever is the default and keep this as it is. Also I am bit unsure how to use the new rebootMode here?

If we enable user to make that decision we may need to change the Proposal and add new field into Metal3RemediationSpec for selection  of rebootMode. Any thoughts?

controllers/metal3remediation_controller.go
beekhof: the defer function is only used if the parent function progresses past the point the defer is declared.

so at least this version is safe

controllers/metal3remediation_controller.go
beekhof: seems like a reasonable idea, we already know there will be a new strategy that deletes Node objects

n1r1 (commented):
Looks good to me, and from my PoV, we can merge it.
The only missing thing IMO is the correct api group (see below)

config/crd/bases/infrastructure.cluster.x-k8s.io_metal3remediations.yaml
```diff
@@ -0,0 +1,99 @@
...
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.5.0
+  creationTimestamp: null
+  name: metal3remediations.infrastructure.cluster.x-k8s.io
+spec:
+  group: infrastructure.cluster.x-k8s.io
```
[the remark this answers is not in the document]
n1r1: Note that you will need to add a cluster role so that MHC in CAPI can create such CRs.
see example here:
https://github.com/medik8s/poison-pill/pull/3

but please note this is aimed for OCP and poison pill. There are some more details on that slack conversation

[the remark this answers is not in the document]
jan-est: Now when controller implementation is done within the CAPM3 the CRD are also defined in it. For changing CRDs to metal3 group we should move the whole controller outside the CAPM3. Right?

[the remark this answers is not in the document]
jan-est: Implementation of Remediation Controller is now part of the CAPM3:  https://github.com/metal3-io/cluster-api-provider-metal3 

So the controller is similar to metal3machine controller etc. When generated it within the CAPM3 all the apis and CRD defined in it belongs to this groupversion: https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/api/v1alpha4/groupversion_info.go

If we would like to use something like remediation.metal3.io IMO we should take the controller outside the CAPM3 and make it stand alone controller like ip-address-manager: https://github.com/metal3-io/ip-address-manager/blob/master/api/v1alpha1/groupversion_info.go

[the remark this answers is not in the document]
jan-est: Maybe taking controller outside is the good idea anyway. Any thoughts?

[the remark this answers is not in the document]
n1r1: yeah, that makes sense to me.
In [medik8s](https://github.com/medik8s) we also went this path of having a separate operator for remediation controllers.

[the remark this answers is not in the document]
kashifest: Since the implementation for this almost ready, I would vote for keeping it inside CAPM3 and then add a separate controller and take it out of CAPM3 in a future workitem, we can also modify the proposal to reflect this.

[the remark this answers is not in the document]
zaneb: Following up on discussion in the community meeting: I don't see a lot of value in creating an extra repo for this. If it's something that CAPI interacts with then it seems to me that conceptually it belongs in the CAPM3 repo unless there's some technical reason for splitting it (which I understand there's not).

That said, I'm not clear why CAPI folks are apparently recommending that this not be part of the same API group as the Metal3Machine &c. CRDs, so I don't have all the context here.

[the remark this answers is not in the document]
n1r1: Since there could be many external remediation providers, there was a concern that using the same api group would increase the chances for a collision of two different remediators using the same kind.
I believe the same concern is applicable for all the existing machines so it's not completely different.

I'm not sure if CAPI will object using the same api group, and I believe it should technically work.
more context here:
https://kubernetes.slack.com/archives/C8TSNPY4T/p1615886930004900

in medik8s, we used different api group per remediator, and used role aggregation to grant MHC permissions to interact with those external api groups.

I'm fine with both ways. whatever works for you.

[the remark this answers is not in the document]
jan-est: I agree with @kashifest and @zaneb I would rather go with the infrastructure specific remediation and keep the controller within CAPM3. It make more sense to have remediationRequest and remediationTemplate defined in the same API group as Machines we want to remediate.

kashifest (approved):

furkatgofurov7:
@jan-est this needs a rebase.

PR #162, opened 2021-02-21
🌱 Install cert-manager with extension

fmuyassarov:
Install cert-manager with extension instead of installing it ourselves. Related to https://github.com/kubernetes-sigs/cluster-api/issues/3765
Link to the extension: [cert-manager](https://github.com/tilt-dev/tilt-extensions/tree/master/cert_manager)

### DIFF
--- a/tilt_modules/cert_manager/Tiltfile
+++ b/tilt_modules/cert_manager/Tiltfile
@@ -1,0 +1,62 @@
+cert_manager_test_resources = """
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: cert-manager-test
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
+metadata:
+  name: test-selfsigned
+  namespace: cert-manager-test
+spec:
+  selfSigned: {}
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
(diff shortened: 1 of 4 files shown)


tilt_modules/cert_manager/Tiltfile
```diff
@@ -0,0 +1,62 @@
...
+  secretName: selfsigned-cert-tls
+  issuerRef:
+    name: test-selfsigned
+"""
+
+# Deploys cert manager to your environment
+def deploy_cert_manager(registry="quay.io/jetstack", version="v0.16.1", load_to_kind=False, kind_cluster_name="kind"):
+    silent=True
+
+    # check if cert-mamager is already installed, otherwise pre-load images & apply the manifest
```
furkatgofurov7, line 32:
```suggestion
    # check if cert-manager is already installed, otherwise pre-load images & apply the manifest
```

furkatgofurov7 (commented):
Not blocker, a little nit, otherwise looks great.

tilt_modules/cert_manager/Tiltfile
furkatgofurov7: Hmm, good point. I see other providers have same thing, so origin was from cert-manager file. It is fine, in that case.

kashifest (approved):

PR #179, opened 2021-03-05
Setup source cluster with capi e2e framework

Insullone:
This is a part of e2e testing. More work is on the way.


furkatgofurov7:
These changes will be taken care of in #194.

PR #188, opened 2021-04-06
✨  Install kustomize also in Mac OS.

levsha:
Tests automatically install and use own copy of 'kustomize' binary. Current installation script  does this for linux only, even that all tests run properly on Mac OS too.
This change adds 'kustomize' binary installation also on Mac OS.

Tested: make test


fmuyassarov (commented):
/lgtm

PR #192, opened 2021-04-12
:sparkles: Add golint and generate test in travis

kashifest:


hack/generate.sh
```diff
@@ -39,5 +39,5 @@ else
     --entrypoint sh \
     --workdir /go/src/github.com/metal3-io/cluster-api-provider-metal3 \
     registry.hub.docker.com/library/golang:1.16 \
-    /go/src/github.com/metal3-io/cluster-api-provider-metal3/hack/codegen.sh
+    /go/src/github.com/metal3-io/cluster-api-provider-metal3/hack/generate.sh
```
fmuyassarov, line 42: This change requires also update in prow config in project-infra.

kashifest: Ok , I ll do the change. Thanks for pointing out

kashifest: Actually I have decided otherwise since codegen is used across repos. It will require more changes in project-infra. So keeping it as it is for now.

PR #197, opened 2021-04-19
🐛 Fix bug that BMH NetworkData and MetaData cleaned on deprovisioning when they should not

### LINKED ISSUE #196: BMH NetworkData and userData cleaned on deprovisioning when they should not
This bug was raised by the airship project. CAPM3 deletes the content of the networkData and metaData fields even when it does not manage them.

CAPM3 should not always delete those, only when the networkData and metaData fields are set in the status of the metal3machine

The problem is on line https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L513 and https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L518 . There should be a check similar to https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L924 and https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/baremetal/metal3machine_manager.go#L930

**Environment:**

- Cluster-api version:  NA
- Cluster-api-provider-metal3 version: v0.4.X
- Environment (metal3-dev-env or other): NA
- Kubernetes version: (use `kubectl version`): NA

/kind bug

namnx228:

Fixes #196

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -506,15 +506,15 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			host.Spec.Image = nil
 			bmhUpdated = true
 		}
-		if host.Spec.UserData != nil {
+		if m.Metal3Machine.Status.UserData != nil && host.Spec.UserData != nil {
 			host.Spec.UserData = nil
 			bmhUpdated = true
 		}
-		if host.Spec.MetaData != nil {
+		if m.Metal3Machine.Status.MetaData != nil && host.Spec.MetaData != nil {
 			host.Spec.MetaData = nil
 			bmhUpdated = true
 		}
-		if host.Spec.NetworkData != nil {
+		if m.Metal3Machine.Status.NetworkData != nil && host.Spec.NetworkData != nil {
 			host.Spec.NetworkData = nil
 			bmhUpdated = true
 		}
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -98,6 +98,10 @@ func m3mSecretStatus() *capm3.Metal3MachineStatus {
 	}
 }
 
+func m3mSecretStatusNil() *capm3.Metal3MachineStatus {
+	return &capm3.Metal3MachineStatus{}
+}
+
 func consumerRef() *corev1.ObjectReference {
 	return &corev1.ObjectReference{
 		Name:       "mym3machine",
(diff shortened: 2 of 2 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -510,11 +510,11 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			host.Spec.UserData = nil
```
furkatgofurov7, line 510: @maelk what about the `UserData`, is it fine to delete the content of this field?

furkatgofurov7 (commented):
This LGTM.

baremetal/metal3machine_manager.go
maelk: that one is pretty much always present since it comes from CAPI. but we could put the same check there, it would not hurt

baremetal/metal3machine_manager_test.go
```diff
@@ -1397,6 +1397,12 @@ var _ = Describe("Metal3Machine manager", func() {
 					expectedName = tc.ExpectedConsumerRef.Name
 				}
 				Expect(name).To(Equal(expectedName))
+				if machineMgr.Metal3Machine.Status.MetaData == nil {
```
maelk, line 1400: None of the bmh you are working with in the tests have the networkData and metaData set I think, so you should modify bmhSpecSomeImg for example to add it there, so that it is actually tested

namnx228:
Thanks for review, @maelk @furkatgofurov7. I have fixed all change requests. PTAL again

baremetal/metal3machine_manager_test.go
```diff
@@ -1664,6 +1681,17 @@ var _ = Describe("Metal3Machine manager", func() {
 			ExpectSecretDeleted:       true,
 			ExpectClusterLabelDeleted: false,
 		}),
+		Entry("BMH NetworkData and MetaData should not be cleaned on deprovisioning", testCaseDelete{
```
furkatgofurov7, line 1684: maybe UserData too, as you have added it in the `bmhSpecSomeImg`?

kashifest (approved):

baremetal/metal3machine_manager_test.go
namnx228: Fixed! Thanks

furkatgofurov7 (commented):
/lgtm

PR #198, opened 2021-04-26
Fix kind version comparison

Xenwar:
Fix comparison of kind versions.


namnx228:
LGTM

fmuyassarov:
Force merging it, as prow is still not in place.

PR #203, opened 2021-05-02
🌱 Revert Add golint and generate test in travis

fmuyassarov:
This reverts https://github.com/metal3-io/cluster-api-provider-metal3/pull/192 as Prow cluster is back on track and disables unit & linting test.


kashifest:
This is not needed anymore since we have disabled travis from the settings

PR #204, opened 2021-05-06
🌱 Fix example manifests generation

wgslr:
- fixes the `examples/generate.sh` script to successfully generate example manifests. It also pins the versions of images used in example manifests to avoid breaking compatibility with the manifests.
- updates `ensure-kustomize.sh` to use an up-to-date URL for downloading kustomize

This allows generating examples and running `make deploy`, which is a step in a good direction. There is still an issue with `capm3-baremetal-operator-controller-manager` not receiving basic auth credentials, which will be addressed in a future PR.

I will squash the commits after review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ kubeconfig
 bin/*
 hack/tools/bin/*
 examples/_out/*
+examples/envsubst-go
 examples/provider-components/*-components.yaml
 out/*
 
--- a/Makefile
+++ b/Makefile
@@ -334,8 +334,8 @@ deploy: generate-examples
 	kubectl wait --for=condition=Available --timeout=300s -n cert-manager deployment cert-manager-webhook
 	kubectl apply -f examples/_out/provider-components.yaml
 
-deploy-examples:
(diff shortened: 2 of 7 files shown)


kashifest (commented):
Thanks a lot for taking caring of this.
/test-integration
/approve

namnx228 (commented):
Overall, it looks good to me. Just a question: Why do we need to have an image version patch instead of directly fix the versions of CAPI components.

wgslr:
> Overall, it looks good to me. Just a question: Why do we need to have an image version patch instead of directly fix the versions of CAPI components.

If I understand the situation correctly, moving to the newer versions in the examples would be moving to CAPM3 version v1alpha4. Which is a separate effort, and for now we also want to have working examples with v1alpha3.

I'd also argue that even if a never version should be used, it still should be pinned to specific tags rather than the master to make the examples are reliable and reproducible.

fmuyassarov:
@wgslr This looks good. Can you please squash the commits before we merge the PR?

wgslr:
@fmuyassarov done

PR #210, opened 2021-05-18
🌱 Add rbac rules for MachineSets 

furkatgofurov7:
There is a rbac issue with MachineSets while scaling up worker and Metal3MachineController logs outputs:
```
`E0518 07:49:42.976339       1 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.9/tools/cache/reflector.go:105: Failed to list *v1alpha3.MachineSet: machinesets.cluster.x-k8s.io is forbidden: User "system:serviceaccount:capm3-system:default" cannot list resource "machinesets" in API group "cluster.x-k8s.io" at the cluster scope
`
```
Fixes #

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -86,6 +86,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - cluster.x-k8s.io
+  resources:
+  - machinesets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - infrastructure.cluster.x-k8s.io
   resources:
   - metal3clusters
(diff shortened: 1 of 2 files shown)


config/rbac/role.yaml
```diff
@@ -85,6 +85,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - cluster.x-k8s.io
+  resources:
+  - machinesets
+  verbs:
+  - create
```
fmuyassarov, line 93: do we need all of these permissions? Maybe list, get, watch would be enough ?

furkatgofurov7: IAlthough errors logs complain with listing problem only,  I have enabled all of those permissions because that's what was done for KCP, MD in previous patch, so I am not sure.

kashifest (approved):
I agree with @fmuyassarov lets minimize the verbs to those that is needed only.

furkatgofurov7:
> I agree with @fmuyassarov lets minimize the verbs to those that is needed only.

Okay, fixed.

PR #216, opened 2021-05-24
🐛 Fix code duplication

fmuyassarov:
Fix unnecessary code duplication in test code.


furkatgofurov7 (commented):
We can enable some kind of dead code finder linter in the repo to avoid such duplications in the future. 

/lgtm

kashifest (approved):

PR #167, opened 2021-02-27
⚠️Prepare for CAPI v1alpha4 changes

furkatgofurov7:
**What this PR does / why we need it**:
Updated list which follows the sequence from the providers document, ref: [Cluster API v1alpha3 compared to v1alpha4 ](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md)

- Uplifts [Controller Runtime version](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#controller-runtime-version) to v0.9.1 and CAPI to latest v0.4.0
- Updates [Controller tools version](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#controller-tools-version-if-used) to v.0.6.0
- Updates [minimum kind version to v0.11.0](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#kind-version) 
- Makes [Go Module changes ](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#warning-go-module-changes-warning)
- Upgrades [klog version to v2](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#klog-version)
- Aligns [manager flag names with upstream Kubernetes components](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#align-manager-flag-names-with-upstream-kubernetes-components)
- Replaces functions` fake.NewFakeClientWithScheme` and `fake.NewFakeClient ` with [`fake.NewClientBuilder().WithObjects().Build()`](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#use-to-controller-runtimes-new-fake-client-builder)
- Adds [optional flag `--webhook-cert-dir={string-value}` ](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#optional-flag-for-specifying-webhook-certificates-dir) 
[...]

PR #218, opened 2021-06-02
🌱E2E template parametrization

wgslr:
This PR replaces hardcoded values in e2e testing template with env variable substitution.


furkatgofurov7 (commented):
Mostly looks good, have you tested it locally? Also, we can't for now run tests in this patch since we do not have jenkins jobs + according changes in place yet.

hack/e2e/environment.sh
```diff
@@ -0,0 +1,40 @@
...
+export IMAGE_OS="Ubuntu"
+export CONTAINER_RUNTIME="docker"
+export EPHEMERAL_CLUSTER="kind"
+export CAPI_VERSION="v1alpha3"
+export CAPM3_VERSION="v1alpha4"
+export NUM_NODES=4
+
+export KUBERNETES_VERSION=v1.21.1
+export API_ENDPOINT_HOST=192.168.111.249
+export API_ENDPOINT_PORT=6443
```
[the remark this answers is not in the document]
wgslr: Ok, I quoted all of them.

hack/e2e/environment.sh
```diff
@@ -0,0 +1,40 @@
...
+export PROVISIONING_URL_HOST="172.22.0.1"
+export IMAGE_NAME="UBUNTU_20.04_NODE_IMAGE_K8S_${KUBERNETES_VERSION}.qcow2"
+export IMAGE_LOCATION="https://artifactory.nordix.org/artifactory/airship/images/k8s_${KUBERNETES_VERSION}"
+export IMAGE_CHECKSUM="http://${PROVISIONING_URL_HOST}/images/UBUNTU_20.04_NODE_IMAGE_K8S_${KUBERNETES_VERSION}-raw.img.md5sum"
+export IMAGE_CHECKSUM_TYPE=md5
+export IMAGE_URL="http://${PROVISIONING_URL_HOST}/images/UBUNTU_20.04_NODE_IMAGE_K8S_${KUBERNETES_VERSION}-raw.img"
+
+# preserve the template, to be substituted based on value defined in e2e_test.go
+# shellcheck disable=SC2016
+export CONTROL_PLANE_MACHINE_COUNT='${CONTROL_PLANE_MACHINE_COUNT}'
```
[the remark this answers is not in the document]
wgslr: Hm, I can define the variable and have it substituted in envsubst. But I'm not sure why is decoupling from e2e_test.go better?

wgslr:
Yes, I did run the tests manually on my VM and they pass.

hack/e2e/environment.sh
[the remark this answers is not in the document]
furkatgofurov7: Because, that way it is easier IMO to set number of controlplane and worker machines for the user rather than looking for go code.

templates/test/cluster-template-prow-ha.yaml
```diff
@@ -177,45 +175,42 @@ spec:
...
     - systemctl start monitor.keepalived.service
     users:
-    - name: metal3
+    - name: ${IMAGE_USERNAME}
       sshAuthorizedKeys:
       - ${SSH_PUB_KEY_CONTENT}
       sudo: ALL=(ALL) NOPASSWD:ALL
-  nodeDrainTimeout: 0s
-  replicas: 3
-  #rolloutStrategy:
```
furkatgofurov7, line 191: why are these removed?

furkatgofurov7 (commented):
There couple of more items that needs parameterization, please go through it once again. These are the ones I spotted, you can use below names borrowed from m3-dev-env if you have not defined them yet in env.sh, otherwise you can reuse already defined ones: 

- [CLUSTER_APIENDPOINT_HOST](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/templates/test/cluster-template-prow-ha.yaml#L126)

- [IRONIC_ENDPOINT_BRIDGE](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/templates/test/cluster-template-prow-ha.yaml#L135) 

- [REGISTRY](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/templates/test/cluster-template-prow-ha.yaml#L147)

- [IMAGE_USERNAME](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/templates/test/cluster-template-prow-ha.yaml#L173-L176)

- same [IRONIC_ENDPOINT_BRIDGE](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/templates/test/cluster-template-prow-ha.yaml#L368) and [REGISTRY](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/templates/test/cluster-template-prow-ha.yaml#L380) too

templates/test/cluster-template-prow-ha.yaml
```diff
@@ -177,45 +175,42 @@ spec:
...
-  #  rollingUpdate:
-  #    maxSurge: 1
```
furkatgofurov7, line 193: why are these removed?

wgslr: To not keep commented out code around

hack/e2e/environment.sh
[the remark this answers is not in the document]
wgslr: Ok, I've reverted that change. Although it means that the number given in e2e_test.go will be a bit confusing; it can be even set to a random number like `5` and the test will still pass alright.

templates/test/cluster-template-prow-ha.yaml
```diff
@@ -170,52 +168,52 @@ spec:
...
+    - mkdir -p /home/${IMAGE_USERNAME}/.kube
+    - cp /etc/kubernetes/admin.conf /home/${IMAGE_USERNAME}/.kube/config
     - systemctl enable --now keepalived
-    - chown metal3:metal3 /home/metal3/.kube/config
+    - chown ${IMAGE_USERNAME}:${IMAGE_USERNAME} /home/${IMAGE_USERNAME}/.kube/config
     preKubeadmCommands:
     - netplan apply
     - systemctl enable --now crio kubelet
-    - if (curl -sk --max-time 10 https://192.168.111.249:6443/healthz); then echo "keepalived already running";else systemctl start keepalived; fi
+    - if (curl -sk --max-time 10 https://${API_ENDPOINT_HOST}:${API_ENDPOINT_PORT}/healthz); then echo "keepalived already running";else systemctl start keepalived; fi
```
macaptain, line 178: I think this should be `CLUSTER_APIENDPOINT_HOST` and `CLUSTER_APIENDPOINT_PORT` to keep the same variable names as are used in the metal3-dev-env.

scripts/ci-e2e.sh
```diff
@@ -26,12 +26,12 @@ WORKING_DIR=/opt/metal3-dev-env
 sudo mkdir -p ${WORKING_DIR}
 sudo chown "${USER}":"${USER}" ${WORKING_DIR}
 
-M3_DEV_ENV_REPO="https://github.com/metal3-io/metal3-dev-env.git"
-M3_DEV_ENV_BRANCH=master
+M3_DEV_ENV_REPO="https://github.com/nordix/metal3-dev-env.git"
+M3_DEV_ENV_BRANCH="run-arbitrary-command-wgslr"
```
wgslr, line 30:
```suggestion
M3_DEV_ENV_REPO="https://github.com/metal3-io/metal3-dev-env.git"
M3_DEV_ENV_BRANCH=master
```
TODO Revert the change before merging

templates/test/cluster-template-prow-ha.yaml
wgslr: Good catch. I changed it elsewhere in the file but not here

fmuyassarov (commented):
LGTM

scripts/ci-e2e.sh
fmuyassarov: is this needed for testing with integration test ?

wgslr: It was needed for checking that e2e tests work. The change can be removed as soon as [m3-dev-env#703](https://github.com/metal3-io/metal3-dev-env/pull/703) gets merged

hack/e2e/environment.sh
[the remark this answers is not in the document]
wgslr: Judging by today's call adjusting it in the template is not easier than in go. Maybe let's restore the `CONTROL_PLANE_MACHINE_COUNT`?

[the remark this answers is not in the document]
furkatgofurov7: Yes, let's restore it.

test/e2e/config/e2e_conf.yaml
```diff
@@ -59,13 +59,11 @@ providers:
     files:
     - sourcePath: "${PWD}/metadata.yaml"
       targetName: "metadata.yaml"
-    - sourcePath: "${PWD}/templates/test/cluster-template-prow-single-master.yaml"
-      targetName: "cluster-template-single-master.yaml"
-    - sourcePath: "${PWD}/templates/test/cluster-template-prow-ha.yaml"
+    - sourcePath: "${PWD}/templates/test/cluster-template-prow-ha_envsubst.yaml"
```
namnx228, line 62: You sure it is not `cluster-template-prow-ha.yaml`?

namnx228 (approved):
Overall, it looks good to me. Just leave a comment.

hack/e2e/environment.sh
```diff
@@ -0,0 +1,19 @@
+# File used as metal3-dev-env config and as a source for variables used by envsubst
+
+# metal3-dev-env customization
+export IMAGE_OS=${IMAGE_OS:-"Ubuntu"}
+export CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"docker"}
+export EPHEMERAL_CLUSTER=${EPHEMERAL_CLUSTER:-"kind"}
+export CAPI_VERSION=${CAPI_VERSION:-"v1alpha4"}
+export CAPM3_VERSION=${CAPM3_VERSION:-"v1alpha5"}
```
furkatgofurov7, line 8: why v1a4/v1a5? we are not there yet right? I think we can keep both `CAPI_VERSION` and `CAPM3_VERSION` set to `v1a3` && `v1a4` accordingly for now, and we will update it as part of [#167](https://github.com/metal3-io/cluster-api-provider-metal3/pull/167/files#diff-6cd24b3345ce830e47a9bb53e46046248df09d8cb27775f49cd514877ed78c17R4-R5)

test/e2e/config/e2e_conf.yaml
wgslr: Yes. The `cluster-template-prow-ha_envsubst.yaml` is generated by passing `cluster-template-prow-ha.yaml` through envsubst, which is done in the `e2e-substitutions` make target.

templates/test/cluster-template-prow-single-master.yaml
```diff
@@ -1,397 +0,0 @@
-apiVersion: cluster.x-k8s.io/v1alpha3
```
furkatgofurov7, line 1: To clarify, we can live without this file and control number of ctpl/workers from the go code itself using the vars from environment.sh?

namnx228: > To clarify, we can live without this file and control number of ctpl/workers from the go code itself using the vars from environment.sh?

Have the same concern. It would be nice if we can have that

hack/e2e/environment.sh
wgslr: When running in the CI it's overridden by env vars, which [are already set to v1alpha4/alpha5](https://gerrit.nordix.org/gitweb?p=infra/cicd.git;a=blob;f=jjb/airship/job_capi_capm3_v1a5_e2e_tests_ubuntu.yml;h=04024720ff8a76b843fff6c4a9c191ae99c98ecd;hb=HEAD#l61). But I can stick to the old values as the defaults for manual runs.

templates/test/cluster-template-prow-single-master.yaml
wgslr: Yes. We can control the number of ctpl and workers using ControlPlaneMachineCount and WorkerMachineCount fields in `ConfigClusterInput` in the go code. No need to modify environment.sh either, except for setting NUM_NODES high enough to accomodate all the machines.

lentzi90 (commented):
This is certainly an improvement over having everything hard-coded. But I'm wondering if it would make sense to use go templates instead of envsubst in the long run?

Environment variables are probably fine as long as we just want them set to one value for all tests. However, I guess we may want to set different values for different e2e tests and still run them all as part of one test suite. For example, we may want to create a HA cluster in one test and a single machine cluster in another. With go templates it would be very natural to define these parameters in the test itself instead of setting them through environment variables.
What do you think?

wgslr:
> For example, we may want to create a HA cluster in one test and a single machine cluster in another.

This particular example - differentiating node numbers - is already achievable by chaning values in the go code. But I see your point; if we reach a use case that demands more customization of the templates it may be necessary to move to a templating solution. Do you have some particular templating mechanism in mind? For reference, AWS provider [uses kustomize](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/infrastructure-aws/README.md)

lentzi90:
> Do you have some particular templating mechanism in mind? For reference, AWS provider [uses kustomize](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/infrastructure-aws/README.md)

Interesting! Then for AWS it would still require changes outside of the go code to modify the flavors. :thinking: 
I was thinking about [go template](https://golang.org/pkg/text/template/) which is already built in to the language and should be familiar to many from Helm templates. This would allow us to specify the values we want *inside* the test code.

jan-est:
LGTM

Makefile
```diff
@@ -105,16 +105,21 @@ GINKGO_NOCOLOR ?= false
...
 E2E_CONF_FILE_ENVSUBST ?= $(ROOT_DIR)/test/e2e/config/e2e_conf_envsubst.yaml
+E2E_TEMPLATE_FILE ?= $(ROOT_DIR)/templates/test/cluster-template-prow-ha.yaml
+E2E_TEMPLATE_FILE_ENVSUBST ?= $(ROOT_DIR)/templates/test/cluster-template-prow-ha_envsubst.yaml
 SKIP_CLEANUP ?= false
 SKIP_CREATE_MGMT_CLUSTER ?= true
 
+.PHONY: e2e-substitutions
+e2e-substitutions: $(ENVSUBST)
+	$(ENVSUBST) < $(E2E_CONF_FILE) > $(E2E_CONF_FILE_ENVSUBST)
+	$(ENVSUBST) < $(E2E_TEMPLATE_FILE) > $(E2E_TEMPLATE_FILE_ENVSUBST)
```
furkatgofurov7, line 116: do we really need a separate make target for env substitution? previously we had it within the e2e-tests target is not it?

test/e2e/config/e2e_conf.yaml
furkatgofurov7: I missed this part while reviewing, would taking e2e-substitutions into e2e-tests make target help with that, because otherwise we would have to have `_envsubst` added for each template yaml naming where we could have multiple of them in the future

furkatgofurov7 (commented):
Last thing to clarify before we move on with this one:

hack/e2e/environment.sh
furkatgofurov7: Yes, the point is as of now we do not have a working CAPM3 v1a5 with CAPI v1a4 yet, the vars can be set to anything, it just to avoid confusion for users thinking they are running e2e with latest CAPI.

lentzi90 (approved):

furkatgofurov7 (approved):
/approve

PR #222, opened 2021-06-08
 ⚠️ Add go.mod under the /apis

kashifest:
**Important**:
Once this PR goes in, those who need to uplift CAPM3, and they are only using CAPM3 APIs , you need to change your go modules to use

github.com/metal3-io/cluster-api-provider-metal3/api
instead of:

github.com/metal3-io/cluster-api-provider-metal3
If you are using/referring to other parts of code in CAPM3 not residing in /api, you need to use both the modules:

github.com/metal3-io/cluster-api-provider-metal3/api
github.com/metal3-io/cluster-api-provider-metal3/


jan-est:
LGTM

hack/gofmt.sh
```diff
@@ -12,7 +12,7 @@ if [ "${IS_CONTAINER}" != "false" ]; then
   cd /tmp/unit
   make fmt > /tmp/fmt-output.log
   FILE_LENGTH="$(wc -l /tmp/fmt-output.log | awk '{ print $1 }')"
-  if [ "${FILE_LENGTH}" != "1" ]; then
+  if [ "${FILE_LENGTH}" != "2" ]; then
```
maelk, line 15: why did this change ?

kashifest: The gofmt is failing otherwise since we are doing fmt twice once in root folder once in /api

furkatgofurov7: @maelk since we are doing go fmt inside the /api folder in Makefile, without above change gofmt fails with:
```
+ wc -l /tmp/fmt-output.log
+ awk { print $1 }
+ FILE_LENGTH=2
+ [ 2 != 1 ]
+ echo Formatting error! Please run 'make fmt' to correct the problem.
Formatting error! Please run 'make fmt' to correct the problem.
+ echo The problematic files are listed below, after the command that should be run
The problematic files are listed below, after the command that should be run
+ cat /tmp/fmt-output.log
go fmt ./controllers/... ./baremetal/... .
cd api; go fmt  ./...
+ exit 1
```
We thought that since we have run gofmt twice and FILE_LENGTH=2 as a result of that, the check for it was changed from 1 to 2. Does that make sense?

furkatgofurov7: running `make fmt` was passing fine and was not changing anything actually.

maelk: yes, makes sense, thanks. Can you please add a short comment saying what those two lines are ?

maelk: basically those two lines are the headers from two gofmt runs

furkatgofurov7: cc @kashifest

kashifest: @maelk resolved

maelk:
oops sorry, I confused the branches
/test-integration

furkatgofurov7 (commented):
/lgtm

furkatgofurov7 (commented):
/lgtm

go.mod
```diff
@@ -6,22 +6,19 @@ require (
 	cloud.google.com/go v0.78.0 // indirect
 	github.com/go-logr/logr v0.4.0
 	github.com/golang/mock v1.5.0
-	github.com/google/go-cmp v0.5.5 // indirect
-	github.com/google/gofuzz v1.2.0
 	github.com/google/uuid v1.2.0 // indirect
 	github.com/googleapis/gnostic v0.5.4 // indirect
 	github.com/imdario/mergo v0.3.12 // indirect
 	github.com/metal3-io/baremetal-operator/apis v0.0.0-20210416073321-c927d1d8da76
-	github.com/metal3-io/ip-address-manager/api v0.0.0-20210609163946-48b0ce9a1ac0
```
furkatgofurov7, line 15: This has been removed, instead of `github.com/metal3-io/ip-address-manager v0.0.6` bit down

maelk (approved):

PR #223, opened 2021-06-08
✨ Add BMO inject-ca-from custom patching

ardaguclu:
This PR patches `inject-ca-from` annotations in baremetalhost crd and validatingwebhookconfiguration of baremetal operator and also certificate dnsNames. Currently this PR has no effect. But this is prerequisite for https://github.com/metal3-io/baremetal-operator/pull/865 to integration tests pass and bmo validating webhook works properly in capm3.

Problem is baremetal operator itself is running under `baremetal-operator-system` namespace and it renders certificates with `baremetal-operator-system/baremetal-operator-serving-cert`. However, capm3 runs baremetal operator under the `capm3-system` namespace and that's why, webhook in bmo can not find correct certificate to run.

### DIFF
--- a/config/bmo/bmo_inject_ca_patch.yaml
+++ b/config/bmo/bmo_inject_ca_patch.yaml
@@ -1,0 +1,3 @@
+- op: replace
+  path: "/metadata/annotations/cert-manager.io~1inject-ca-from"
+  value: capm3-system/capm3-baremetal-operator-serving-cert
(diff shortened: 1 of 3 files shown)


furkatgofurov7:
/test-integration 

Triggering it again, since previous run passed v1a4 tests whereas there has been a change yesterday and we have to run v1a5 integration tests on PR towards master branch.

PR #227, opened 2021-06-11
✨ Add Furkat and Anwar as approvers

kashifest:
Adding Furkat and Anwar as approvers as agreed. Russel is being removed as per his request.


dhellmann:
This matches the outcome of the email conversation.

Thanks!

/approve

PR #232, opened 2021-06-28
✨ Add pivoting test based on the e2e test framework

namnx228:
This PR adds the pivoting test based on the e2e test framework.


test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+	capm3Version := os.Getenv("CAPM3_VERSION")
+	if capm3Version == "v1alpha5" {
+		By("Add labels to BMO CRDs")
+		labelBMOCRDs(nil)
+		By("Install BMO")
+		installIronicBMO(targetCluster, "false", "true")
+	}
+
+	By("Initialize Provider component in target cluster")
+	clusterctl.Init(context.TODO(), clusterctl.InitInput{
```
wgslr, line 63: Since we have the global `ctx` variable, maybe it should be used here?
Although I must admit I don't see any reason or benefit for passing all this ctx around...

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+	By("Check if bmh is in provisioned state")
+	Eventually(func() error {
+		bmoList := &bmo.BareMetalHostList{}
+		if err := targetCluster.GetClient().List(ctx, bmoList, client.InNamespace(namespace)); err != nil {
+			return err
+		}
+		for _, bmo := range bmoList.Items {
+			if !bmo.WasProvisioned() {
+				return errors.New("BMOs cannot be provisioned")
```
wgslr, line 128:
```suggestion
		if err := targetCluster.GetClient().List(ctx, bmhList, client.InNamespace(namespace)); err != nil {
			return err
		}
		for _, bmh := range bmhList.Items {
			if !bmh.WasProvisioned() {
				return errors.New("BMHs cannot be provisioned")
```

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+		"USER=ubuntu",
+	}
+	cmd := exec.Command("./deploy.sh", args...)
+	cmd.Dir = path
+	cmd.Env = append(env, os.Environ()...)
+
+	err := cmd.Start()
+	Expect(err).To(BeNil(), "Fail to deploy Ironic")
+	err = cmd.Wait()
+	Expect(err).To(BeNil(), "Fail to deploy Ironic")
```
wgslr, line 211:
```suggestion
	err := cmd.Run()
	Expect(err).To(BeNil(), "Fail to deploy Ironic")
```

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+		"ironic-inspector-log-watch",
+	}
+	dockerClient, err := docker.NewEnvClient()
+	Expect(err).To(BeNil(), "Unable to get docker client")
+	removeOptions := dockerTypes.ContainerRemoveOptions{
+		Force: true,
+	}
+	for _, container := range ironicContainerList {
+		err = dockerClient.ContainerRemove(ctx, container, removeOptions)
+		Expect(err).To(BeNil(), "Unable to delete the container %s, ", container, err)
```
wgslr, line 232:
```suggestion
		Expect(err).To(BeNil(), "Unable to delete the container %s: %v", container, err)
```

wgslr (commented):
Good job :tada: Just a few suggestions

test/e2e/pivoting_test.go
namnx228: Fixed! Thanks

test/e2e/pivoting_test.go
namnx228: I will keep this one a bit longer for the sake of debugging. I will change to use Run() in the last time commit

Rozzii (approved):

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+		Force: true,
+	}
+	for _, container := range ironicContainerList {
+		err = dockerClient.ContainerRemove(ctx, container, removeOptions)
+		Expect(err).To(BeNil(), "Unable to delete the container %s: %v", container, err)
+	}
+}
+func removeIronicDeployment() {
+	deploymentName := os.Getenv("NAMEPREFIX") + "-ironic"
+	ironicNamespace := os.Getenv("IRONIC_NAMESPACE")
```
furkatgofurov7, line 237: I do also suggest we take all vars like these here and in all other funcs so we could define them at the top once and re-use them (if applicable) later, since some of them redefined twice in different funcs, i.e `IRONIC_NAMESPACE`, `BMOPATH`

furkatgofurov7 (commented):
This is great effort and basement for all other upcoming e2e tests after this, good job @namnx228! I have few comments/suggestions/questions inline:

go.mod
```diff
@@ -3,16 +3,25 @@ module github.com/metal3-io/cluster-api-provider-metal3
 go 1.16
 
 require (
+	github.com/Microsoft/go-winio v0.5.0 // indirect
```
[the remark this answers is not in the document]
namnx228: Good catch. That was somehow imported by IDE, but we don't need it now. Will remove it.

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	bmo "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	capm3 "github.com/metal3-io/cluster-api-provider-metal3/api/v1alpha5"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
```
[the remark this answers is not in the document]
namnx228: Fixed! Thanks

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+	Expect(controlPlane).ToNot(BeNil())
+
+	By("Check if bmh is in provisioned state")
+	Eventually(func() error {
+		bmhList := &bmo.BareMetalHostList{}
+		if err := targetCluster.GetClient().List(ctx, bmhList, client.InNamespace(namespace)); err != nil {
+			return err
+		}
+		for _, bmh := range bmhList.Items {
+			if !bmh.WasProvisioned() {
```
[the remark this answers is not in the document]
namnx228: I also think that we should change the name of this function, but its definition is not a part of CAPM3, unfortunately. It is in BMO.

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}, e2eConfig.GetIntervals(specName, "wait-object-provisioned")...).Should(BeNil())
+
+	By("Check if metal3machines become provisioned.")
+	Eventually(func() error {
+		m3Machines := &capm3.Metal3MachineList{}
+		if err := targetCluster.GetClient().List(ctx, m3Machines, client.InNamespace(namespace)); err != nil {
+			return err
+		}
+		for _, m3Machine := range m3Machines.Items {
+			if !m3Machine.Status.Ready {
```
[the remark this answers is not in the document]
namnx228: This is what we have in m3-dev-env. https://github.com/metal3-io/metal3-dev-env/blob/54abb30bb076e237de68b804d3bffe52235ccdb9/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L167
I will change the `By` statement. Thanks

test/e2e/pivoting_test.go
```diff
@@ -0,0 +1,263 @@
...
+	args := []string{
+		isBMO,
+		isIronic,
+		os.Getenv(ironicTLSEnable),
+		os.Getenv(ironicBasicAuth),
+		"true",
+	}
+	env := []string{
+		fmt.Sprintf("IRONIC_HOST=%s", ironicHost),
+		fmt.Sprintf("IRONIC_HOST_IP=%s", ironicHost),
```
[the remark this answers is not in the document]
namnx228: I think they are the same, but it is required by `deploy.sh`: https://github.com/metal3-io/metal3-dev-env/blob/54abb30bb076e237de68b804d3bffe52235ccdb9/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L122

test/e2e/pivoting_test.go
namnx228: I agree that some variables are redefined some times in this PR. However, I think it is not too bad since these variables are mostly in type string, not big objects.I kind of reluctant to use global variables if we only need them in one test to avoid unnecessary access to these variables from other tests. Therefore, I would use local variables in this case.

test/e2e/pivoting_test.go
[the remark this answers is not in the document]
furkatgofurov7: Strange, in fact they are the same and are coming from m3-dev-env

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/metal3-io/cluster-api-provider-metal3
 go 1.16
 
 require (
-	github.com/Microsoft/go-winio v0.5.0 // indirect
+	// github.com/Microsoft/go-winio v0.5.0 // indirect
```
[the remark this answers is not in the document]
namnx228: Oh, ok. I commented to test if the code can work without that module, then I forgot to actually delete it ^^

namnx228:
> I meant to define them in the pivoting_test.go not in e2e_suite_test.go or somewhere else, do you still think it would be accessible by other tests?
@furkatgofurov7  Yes, even when we define the global variables in the `pivoting_test.go`, other tests in other files in package `e2e` can still access it.

test/e2e/config/e2e_conf.yaml
```diff
@@ -73,6 +73,7 @@ providers:
 variables:
   KUBERNETES_VERSION: "${KUBERNETES_VERSION}"
   CNI: "./data/cni/calico/calico.yaml"
+  IRONIC_HOST_IP: "172.22.0.2"
```
furkatgofurov7, line 76: IMO we have this var used in CI in m3-dev-env, can we use that by calling directly wherever instead  of hard-coding here?
`CLUSTER_PROVISIONING_IP` or `CLUSTER_URL_HOST`?

furkatgofurov7 (approved):
I am happy with these changes, we can think of combining calico manifests into a single one and use that for both Ubuntu/CentOS e2e test runs in the future patches.

/approve

PR #237, opened 2021-07-01
🌱Add generation of empty-ironic-cacert secret for handling TLS disabled scenario

smoshiur1237:
In non-TLS scenario BMO is waiting for an empty secret which was removed earlier with cert manager setup in ironic.  Because of that the non-tls run was broken. This PR will add the empty secret creation and  during TLS-disabled, we will substitute the secret name as `empty-ironic-cacert` before kustomization.  The change of secret name will be done in [metal3-dev-env](https://github.com/metal3-io/metal3-dev-env/pull/720)
The changes fix the issue with non-TLS.

### DIFF
--- a/config/bmo/secrets.yaml
+++ b/config/bmo/secrets.yaml
@@ -1,4 +1,12 @@
 apiVersion: v1
+data:
+  tls.crt: ""
+kind: Secret
+metadata:
+   name: empty-ironic-cacert
+type: Opaque
+---
+apiVersion: v1
 stringData:
   password: ${IRONIC_PASSWORD:-""}
   username: ${IRONIC_USERNAME:-""}


smoshiur1237:
Test is going on in Metal3-dev-env for TLS disabled.
/hold

lentzi90 (commented):
Would it be possible to make the volume optional instead of creating this empty secret? I'm thinking that we could add `optional: true` here: https://github.com/metal3-io/cluster-api-provider-metal3/blob/07cb2d8852b2389f726d751c1f0075d403ac4864/config/bmo/secret_mount_patch.yaml#L22-L24
But I'm not sure if BMO can handle a missing file or if it has to be an empty file?

smoshiur1237:
> Would it be possible to make the volume optional instead of creating this empty secret? I'm thinking that we could add `optional: true` here:
> 
> https://github.com/metal3-io/cluster-api-provider-metal3/blob/07cb2d8852b2389f726d751c1f0075d403ac4864/config/bmo/secret_mount_patch.yaml#L22-L24
> 
> 
> But I'm not sure if BMO can handle a missing file or if it has to be an empty file?

At this current scenario when BMO is running as part of CAPM3, BMO will wait for the secret in both TLS enabled or disabled and the CI was failing without adding the empty secret. As most of our CI master job are running with TLS enabled, secret is necessary with the settings in BMO(part of CAPM3).  So the empty secret will be used only to support non-TLS.  I think when we are taking BMO outside of CAPM3, don't need to create an empty secret as the setting has separate TLS and non-TLS configuration.

lentzi90 (approved):
Thanks for the explanation! I think this is fine then :slightly_smiling_face:

namnx228 (approved):
LGTM

furkatgofurov7:
/retitle 🌱Add generation of empty-ironic-cacert secret for handling TLS disabled scenario

furkatgofurov7 (commented):
/approve

PR #245, opened 2021-07-07
🌱 Drop Jaakko from the OWNERS list

fmuyassarov:
Drop @Jaakko-Os as he is no longer in Metal3 project. Thanks for your work @Jaakko-Os !


furkatgofurov7 (commented):
Thanks @Jaakko-Os for all your work!

/approve

PR #247, opened 2021-07-08
🌱Add generation of empty-ironic-cacert secret for handling TLS disabled scenario(back port from master)

smoshiur1237:
In non-TLS scenario BMO is waiting for an empty secret which was removed earlier with cert manager setup in ironic. Because of that the non-tls run was broken. This PR will add the empty secret creation and during TLS-disabled, we will substitute the secret name as empty-ironic-cacert before kustomization. The change of secret name will be done in [metal3-dev-env](https://github.com/metal3-io/metal3-dev-env/pull/720).
The changes fix the issue with non-TLS. This changes merged in [master](https://github.com/metal3-io/cluster-api-provider-metal3/pull/237) and needs to back port to the rlease-0.4.

### DIFF
--- a/config/bmo/secrets.yaml
+++ b/config/bmo/secrets.yaml
@@ -1,4 +1,12 @@
 apiVersion: v1
+data:
+  tls.crt: ""
+kind: Secret
+metadata:
+   name: empty-ironic-cacert
+type: Opaque
+---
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (commented):
/approve

Issue #238, opened 2021-07-01
controllers/Metal3LabelSync/metal3-label-sync-controller back off not correctly set


MaxRink:
**What steps did you take and what happened:**
Im currently seeing a lot of these Log messages in my CAPM3 logs
```
I0701 21:30:57.853102       1 metal3labelsync_controller.go:141] controllers/Metal3LabelSync/metal3-label-sync-controller "msg"="Could not find Node Ref on Machine object, will retry" "metal3-label-sync"={"Namespace":"metal3-refsa2-bn","Name":"ilocz20300gcv.lom.refsa2.bn.schiff.telekom.de"} 
I0701 21:30:58.855701       1 metal3labelsync_controller.go:141] controllers/Metal3LabelSync/metal3-label-sync-controller "msg"="Could not find Node Ref on Machine object, will retry" "metal3-label-sync"={"Namespace":"metal3-refsa2-bn","Name":"ilocz20300gcv.lom.refsa2.bn.schiff.telekom.de"} 
I0701 21:30:59.857972       1 metal3labelsync_controller.go:141] controllers/Metal3LabelSync/metal3-label-sync-controller "msg"="Could not find Node Ref on Machine object, will retry" "metal3-label-sync"={"Namespace":"metal3-refsa2-bn","Name":"ilocz20300gcv.lom.refsa2.bn.schiff.telekom.de"} 

```
As the basically reoccur every half a second or so and only pause after  about 10-15 tries for 10 secons i assume that the backoff for recnciliation isnt properly working as that should propably not run every half a second.

**What did you expect to happen:**

Not getting spammed by log messages. :D

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api version: 0.3.x
- Cluster-api-provider-metal3 version: 0.4.2
- Environment (metal3-dev-env or other): other
- Kubernetes version: (use `kubectl version`): n/a

/kind bug

fmuyassarov:
I agree, at some point I have seen this too and it is annoying.
@Arvinderpal would you have some time to fix this?

Arvinderpal:
Sorry about that. 

[It's currently set to 1 second](https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/controllers/metal3labelsync_controller.go#L48). 

[...]

PR #249, opened 2021-07-08
🐛 Increase retry interval in label-sync controller to 30 seconds

Arvinderpal:
https://github.com/metal3-io/cluster-api-provider-metal3/issues/238

Fixes # https://github.com/metal3-io/cluster-api-provider-metal3/issues/238


furkatgofurov7 (approved):
/approve

PR #259, opened 2021-07-28
✨ Add internal logging mechanism for e2e test framework

namnx228:
This PR adds an internal log for the e2e tests. The reason is that we cannot use the [log](https://github.com/kubernetes-sigs/cluster-api/blob/test/v0.4.0/test/framework/internal/log/log.go) from the cluster-api e2e test framework before the package is [internal](https://pkg.go.dev/cmd/go#hdr-Internal_Directories).

### DIFF
--- a/test/e2e/internal/log/log.go
+++ b/test/e2e/internal/log/log.go
@@ -1,0 +1,28 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


test/e2e/internal/log/log.go
```diff
@@ -0,0 +1,28 @@
...
+// Package log implements test framework logging.
+package log
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+)
+
+func Logf(format string, a ...interface{}) {
```
wgslr, line 26: I'd be fine with putting this next to the `Byf` function in `common.go` instead of creating a new module. But this is also fine

wgslr (approved):

test/e2e/internal/log/log.go
namnx228: Yeah, I think both places are fine, too, but prefer to put the log function to a dedicated package because we may have more advanced log functions in the future.

furkatgofurov7: Indeed we could fetch more logs if needed like [here](https://github.com/oomichi/kubernetes/blob/master/test/e2e/framework/log.go#L1-L106) per need basis.

furkatgofurov7 (approved):
/approve

fmuyassarov (commented):
/lgtm

PR #271, opened 2021-08-19
🐛 Fix the cloud-init in the centos template

namnx228:
This PRs fixes an issue in the Centos cluster template. This issue leads to a networking problem in the Centos target nodes. |The ptoblem is that the `ironicendpoint` bridge doesn't have IP address although it should have it.

### DIFF
--- a/templates/test/cluster-template-prow-ha-centos.yaml
+++ b/templates/test/cluster-template-prow-ha-centos.yaml
@@ -182,8 +182,8 @@ spec:
         ONBOOT=yes
         USERCTL=no
         BOOTPROTO="static"
-        IPADDR={{ "{{ ds.meta_data.provisioningIP }}" }}
-        PREFIX={{ "{{ ds.meta_data.provisioningCIDR }}" }}
+        IPADDR={{ ds.meta_data.provisioningIP }}
+        PREFIX={{ ds.meta_data.provisioningCIDR }}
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
Does it mean that we have the same  issue in metal3-dev-env, because we have the same configuration there: https://github.com/metal3-io/metal3-dev-env/blob/4e53921a232ed02a1cbbaf8c46da961dfaf01744/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers-kubeadm-config-centos.yaml#L44
I see that CentOS job passes in CI though.

namnx228:
> Does it mean that we have the same issue in metal3-dev-env, because we have the same configuration there: https://github.com/metal3-io/metal3-dev-env/blob/4e53921a232ed02a1cbbaf8c46da961dfaf01744/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers-kubeadm-config-centos.yaml#L44
> I see that CentOS job passes in CI though.

No, we don't have the issue in m3-dev-env because the template is processed by Ansible before it is sent  KCP/machinedeployment. Ansbile should removes the first "{{ }}". However, in the e2e test, the template is not used by Ansible, so having two layers of brackets leads to the issue that the variable is not substituted correctly.

kashifest (approved):

PR #273, opened 2021-08-21
🌱 Use same version of Calico images as in metal3-dev-env

fmuyassarov:
Since the latest calico images version is v3.20.0, we are pre-pulling them into the vanilla images to avoid Docker Hub pull rate limit issue. However, e2e is using v3.19.1 images, causing to pull images from docker.


fmuyassarov:
We need to find a way that keep calico images version in sync with what we ship inside the node images.

furkatgofurov7:
> We need to find a way that keep calico images version in sync with what we ship inside the node images.

configure renovate for CAPM3 repo?

kashifest (approved):

PR #278, opened 2021-08-30
🐛 Webhook reg should accept both v1 and v1beta1 admissions

kashifest:
Webhook registration should accept both v1 and v1beta1 admissions

Admission Review was previously set to only accept v1beta1 objects given
that Controller Runtime didn't support v1 (even though the two revisions
are effectively the same). This change adds support for both within our
manifests.

equivalent of [kubernetes-sigs/cluster-api#4942](https://github.com/kubernetes-sigs/cluster-api/pull/4942)


fmuyassarov (commented):
/lgtm

furkatgofurov7 (approved):
/approve

PR #282, opened 2021-09-02
🐛 Fix the problem that "scale in" operation does not work after e2e pivoting

namnx228:
This PR fixes the problem that "scale in" operation does not work after e2e pivoting on Ubuntu. The problem is that the keepalived container is deleted forcefully, so it does not have time to clean the ironicendpoint IP address on the host. This leftover makes target nodes confuse and cannot connect to ironic.

### DIFF
--- a/test/e2e/pivoting_test.go
+++ b/test/e2e/pivoting_test.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


Rozzii (approved):

kashifest (approved):

fmuyassarov (commented):
/lgtm

PR #289, opened 2021-09-09
✨ Add AGE column to metal3 CRDs

kashifest:
This adds the AGE column to all our CRDs to make the printed output consistent with the built in K8s resources.

Keeping things consistent with CAPI# https://github.com/kubernetes-sigs/cluster-api/pull/5180


furkatgofurov7 (approved):
/approve

We are not planning to have that information in v1a4 types?

kashifest:
> /approve
> 
> We are not planning to have that information in v1a4 types?

I didn't see it in the CAPI PR, so for now it seems ok

PR #293, opened 2021-09-15
🐛 Use the infrastructure components built by m3-dev-env

namnx228:
Currently, with e2e tests, the source cluster infrastructure components are initialized by the file made by m3-dev-env, while the target cluster uses the file that is built by the test framework itself. This causes a mismatch between the source and the target cluster.
This PR makes the e2e test framework use the infrastructure components issued by m3-dev-env, leading to the consistency between the source and target clusters.

Note: By default, m3-dev-env clones CAPM3 repo from upstream and build the infrastructure components from here. This behavior is not what we want in an e2e test because the components should be built from the local CAPM3 repo when the test is executed. Therefore, this PR switches the toggle that prevents m3-dev-env to clone repos. The test framework would copy the local CAPM3 repo and clone other repos to the directory where the dev-env works with. By doing that, m3-dev-env should build the infra components based on the local CAPM3 repo.

### DIFF
--- a/hack/e2e/environment.sh
+++ b/hack/e2e/environment.sh
@@ -31,6 +31,11 @@ export CONTROL_PLANE_MACHINE_COUNT='${CONTROL_PLANE_MACHINE_COUNT}'
 # shellcheck disable=SC2016
 export WORKER_MACHINE_COUNT='${WORKER_MACHINE_COUNT}'
 
+# The e2e test framework would itself handle the cloning. It clones all the repos that are cloned in M3-DEV-ENV expect CAPM3.
+# It would use the local CAPM3 repo where the e2e test is running. 
+# Set this variable to false to avoid the dev-env to override what the test framework cloned. 
+export FORCE_REPO_UPDATE="false"
+
 os_check
 
 if [[ "${OS}" == ubuntu ]]; then
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -24,6 +24,14 @@ function clone_repo() {
   fi
 }
 
+function clone_repos() {
+  mkdir -p "${M3PATH}"
+  pushd "${M3PATH}"
+  clone_repo "${BMOREPO}" "${BMOBRANCH}" "${BMOPATH}"
+  clone_repo "${IPAMREPO}" "${IPAMBRANCH}" "${IPAMPATH}"
+  popd
+}
+
 WORKING_DIR=/opt/metal3-dev-env
 sudo mkdir -p ${WORKING_DIR}
 sudo chown "${USER}":"${USER}" ${WORKING_DIR}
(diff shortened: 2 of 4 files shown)


Rozzii (approved):

hack/e2e/environment.sh
```diff
@@ -31,6 +31,11 @@ export CONTROL_PLANE_MACHINE_COUNT='${CONTROL_PLANE_MACHINE_COUNT}'
 # shellcheck disable=SC2016
 export WORKER_MACHINE_COUNT='${WORKER_MACHINE_COUNT}'
 
+# The e2e test framework would itself handle the clonning. It clones all the repos that are clonned in M3-DEV-ENV expect CAPM3.
```
jan-est, line 34:
```suggestion
# The e2e test framework would itself handle the cloning. It clones all the repos that are cloned in M3-DEV-ENV expect CAPM3.
```

namnx228: Fixed! Thanks

kashifest (commented):
/lgtm

smoshiur1237 (approved):
LGTM

hack/e2e/environment.sh
```diff
@@ -31,6 +31,11 @@ export CONTROL_PLANE_MACHINE_COUNT='${CONTROL_PLANE_MACHINE_COUNT}'
 # shellcheck disable=SC2016
 export WORKER_MACHINE_COUNT='${WORKER_MACHINE_COUNT}'
 
+# The e2e test framework would itself handle the cloning. It clones all the repos that are cloned in M3-DEV-ENV expect CAPM3.
```
furkatgofurov7, line 34: maybe:
```suggestion
# The e2e test framework would itself handle the cloning. It clones all the repos that are cloned in M3-DEV-ENV except CAPM3.
```
?

scripts/ci-e2e.sh
```diff
@@ -36,7 +44,17 @@ clone_repo "${M3_DEV_ENV_REPO}" "${M3_DEV_ENV_BRANCH}" "${M3_DEV_ENV_PATH}"
 cp "${REPO_ROOT}"/hack/e2e/environment.sh "${M3_DEV_ENV_PATH}/config_${USER}.sh"
 
 pushd ${M3_DEV_ENV_PATH}
-make
+# Golang needs to be installed before cloning the needed repos to the Go source directory.
+make install_requirements configure_host 
+# shellcheck disable=SC1091
+source lib/common.sh
```
furkatgofurov7, line 50: I wonder if this is needed, we do source it in prev. step (01_* and 02_* scripts), no?

furkatgofurov7 (commented):
Minor suggestions, overall lgtm

scripts/ci-e2e.sh
namnx228: Yes, it is needed, otherwise it will fail. I have tested it.

furkatgofurov7 (approved):
/approve

PR #297, opened 2021-09-20
🐛 Remove capm3- prefix from root kustomize

kashifest:
Adding `capm3-` prefix in root kustomize adds `capm3- `prefix on ipam objects as well. This created issue in IPAM kustomization since the cert-mnanager/ca-inject-from annotation doesn't get updated. This PR only keeps `capm3-` prefix for capm3 objects.


namnx228 (commented):
/lgtm

smoshiur1237:
LGTM

Rozzii (approved):

PR #313, opened 2021-10-05
🌱 Allow fetching repo tags when checking in GitHub Actions

fmuyassarov:
This patch ensures that we fetch all history for all tags
and branches when checking out with GitHub actions.


Rozzii:
LGTM

kashifest (approved):

PR #323, opened 2021-10-06
🐛 Fix bug with missing checkout to a tag during releasing

fmuyassarov:
We need to make sure that we checkout to the correct branch/tag
before we go to release, because otherwise we use wrong tag
version as the prior one.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,6 +10,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Export RELEASE_TAG var
+        run:  echo "RELEASE_TAG=${GITHUB_REF:10}" >> $GITHUB_ENV
       - name: checkout code
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -10,6 +10,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Export RELEASE_TAG var
+        run:  echo "RELEASE_TAG=${GITHUB_REF:10}" >> $GITHUB_ENV
```
namnx228, line 14: Why should we hard code it to 10 here?

namnx228 (commented):
LGTM, one tiny comment

.github/workflows/release.yaml
fmuyassarov: `:10 `strips first ten characters from `GITHUB_REF` var. In other words, instead of  `refs/tags/vx.y.z`  we will get `vx.y.z`

PR #325, opened 2021-10-06
✨ Uplift CAPI and Controller Runtime to latest

furkatgofurov7:
This PR uplifts:

- CAPI to `v0.4.4`
- CR to `v0.10.2`


furkatgofurov7:
> Humm it seems they are keeping the CR to 0.9.7 in [release-0.4 branch](https://github.com/kubernetes-sigs/cluster-api/blob/release-0.4/go.mod#L43) , I think we shall also downgrade it since we are referencing v0.4.x currently. Would you please also check what else is downgraded?

Done

api/go.mod
```diff
@@ -4,17 +4,17 @@ go 1.16
...
 	k8s.io/utils v0.0.0-20210819203725-bdf08cb9a70a
-	sigs.k8s.io/cluster-api v0.4.3
-	sigs.k8s.io/controller-runtime v0.10.1
+	sigs.k8s.io/cluster-api v0.4.4
+	sigs.k8s.io/controller-runtime v0.9.7
 
 )
 
-replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v0.4.3
+replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v0.4.4
```
fmuyassarov, line 20: do we even need replace here, it seems redundant to me since the version is the same as in L15

furkatgofurov7: Yes, otherwise it will fail to fetch it. Also was suggested here: https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/providers/v1alpha3-to-v1alpha4.md#warning-go-module-changes-warning
Example of other provider doing the same thing: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/go.mod#L46-L53

PR #329, opened 2021-10-12
📖 Remove BMO vars from docs

kashifest:
There are still some leftover documentation which was needed when BMO was part of CAPM3. These are not needed anymore and misleading also perhaps.


furkatgofurov7 (approved):
/approve

PR #334, opened 2021-10-14
🌱 Keep k8s var aligned across repo

furkatgofurov7:


kashifest (approved):
I am planning to uplift the m3-dev-env to 1.22.2 soon. So this might need changes again probably next week.

PR #336, opened 2021-10-15
:seedling: E2e upgrade ironic test

mboukhalfa:
**What this PR does / why we need it**: Adds ironic upgrade test


furkatgofurov7 (changes requested):

test/e2e/e2e_test.go
```diff
@@ -84,7 +84,9 @@ var _ = Describe("Workload cluster creation", func() {
 			test_remediation()
 			pivoting()
 			cert_rotation()
+			upgrade_ironic()
```
macaptain, line 87: nit: I know this is consistent with the other test names, but is there a reason we're using snake_case instead of camelCase here? I wonder if we should follow up and make these camelCase.

test/e2e/upgrade_ironic_test.go
```diff
@@ -0,0 +1,59 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func upgrade_ironic() {
+	clientSet := targetCluster.GetClientSet()
+	ironicDeployName := "baremetal-operator-ironic"
+	By("Upgrading ironic image based containers")
+	getIronicDeployment := func() (*appsv1.Deployment, error) {
+		return clientSet.AppsV1().Deployments("baremetal-operator-system").Get(ctx, ironicDeployName, metav1.GetOptions{})
+	}
```
macaptain, line 17: I think it would be a bit more natural if `getIronicDeployment` were defined with a top level `func` statement rather than assigning an anonymous function. What do you think?

test/e2e/upgrade_ironic_test.go
```diff
@@ -0,0 +1,59 @@
...
+	deploy, err := getIronicDeployment()
+	Expect(err).To(BeNil())
+	for i := range deploy.Spec.Template.Spec.Containers {
```
macaptain, line 20: You can loop over the container structs themselves instead of using indexing.

test/e2e/upgrade_ironic_test.go
```diff
@@ -0,0 +1,59 @@
...
+			"ironic-dnsmasq",
+			"ironic-conductor",
+			"ironic-log-watch",
+			"ironic-inspector":
+			deploy.Spec.Template.Spec.Containers[i].Image = "quay.io/metal3-io/ironic:latest"
+		}
+	}
+
+	_, err = clientSet.AppsV1().Deployments("baremetal-operator-system").Update(ctx, deploy, metav1.UpdateOptions{})
+	checkError(err)
```
macaptain, line 34: Could you use `Expect(err).To(BeNil())` instead of your `checkError` function here?

test/e2e/upgrade_ironic_test.go
```diff
@@ -0,0 +1,59 @@
...
+		return deploymentRolledOut(getIronicDeployment, deploy.Status.ObservedGeneration+1)
+	},
+		e2eConfig.GetIntervals(specName, "wait-deployment")...,
+	).Should(Equal(true))
+
+}
+
+func deploymentRolledOut(getDeploymentFunc func() (*appsv1.Deployment, error), desiredGeneration int64) bool {
+	deploy, err := getDeploymentFunc()
+	checkError(err)
```
macaptain, line 47: This `checkError` is run in the polling loop, so if at any point you get an error here, you `os.Exit(1)` and I think that will drop you out of all the tests. Is that what you wanted? Did you try `ginkgo.Fail` if `err != nil`? I think this will also end the test in the `Eventually` block in a more idiomatic way.

test/e2e/common.go
```diff
@@ -28,6 +28,13 @@ func Logf(format string, a ...interface{}) {
 	fmt.Fprintf(GinkgoWriter, "INFO: "+format+"\n", a...)
 }
 
+func checkError(err error) {
+	if err != nil {
+		fmt.Fprintf(GinkgoWriter, "ERROR: %v\n", err)
+		os.Exit(1)
```
macaptain, line 34: I'm not sure I would expect a function named `check...` to `os.Exit(1)`. I think this is more like `panic` than `check`. Does this exit the test process when the error is not nil?

I am not sure we need this function at all. Could we use `ginkgo.Fail` instead?

mboukhalfa: Thanks @macaptain I have changed this function with `Expect(err).To(BeNil())`

test/e2e/upgrade_ironic_test.go
mboukhalfa: I think it is only possible to update the struct on the object using indexes? 
if you meant the following it wont work 

```
for _, container := range deploy.Spec.Template.Spec.Containers {
		switch container.Name {
		case
			"mariadb",
			"ironic-api",
			"ironic-dnsmasq",
			"ironic-conductor",
			"ironic-log-watch",
			"ironic-inspector":
			container.Image = "/metal3-io/ironic:latest"  // should be updated by index instead
		}
	}
```

test/e2e/upgrade_ironic_test.go
```diff
@@ -0,0 +1,73 @@
...
+	for i, container := range deploy.Spec.Template.Spec.Containers {
+		switch container.Name {
+		case
+			"mariadb",
+			"ironic-api",
+			"ironic-dnsmasq",
+			"ironic-conductor",
+			"ironic-log-watch",
+			"ironic-inspector":
+			deploy.Spec.Template.Spec.Containers[i].Image = containerRegistry + "/metal3-io/ironic:" + ironicImageTag
```
macaptain, line 40: Tiny improvement: you could swap `i` for `_` and use `container.Image` here.

mboukhalfa: I have tried this and add a description to the previous comment above but it did not work since that way `deploy` object will not be updated.

test/e2e/upgrade_ironic_test.go
mboukhalfa: @macaptain check this

test/e2e/upgrade_ironic_test.go
macaptain: Oops, yes, sorry. This would have to be a slice to `*Container` for this to work. And it's not. Indexing is totally fine.

test/e2e/remediation_test.go
```diff
@@ -39,7 +39,7 @@ const (
 
 const defaultNamespace = "default"
 
-func remediation() {
+func testRemediation() {
```
furkatgofurov7, line 42: please leave this as is since all other tests do not have `test` prefix

test/e2e/e2e_test.go
```diff
@@ -81,11 +81,11 @@ var _ = Describe("Workload cluster creation", func() {
 			}, result)
 			cluster = result.Cluster
 			targetCluster = bootstrapClusterProxy.GetWorkloadCluster(ctx, namespace, clusterName)
-			remediation()
+			testRemediation()
 			pivoting()
+			upgradeIronic()
 			certRotation()
 			nodeReuse()
-			rePivoting()
```
furkatgofurov7, line 88: this is in place, so revert it

test/e2e/e2e_test.go
```diff
@@ -81,11 +81,11 @@ var _ = Describe("Workload cluster creation", func() {
 			}, result)
 			cluster = result.Cluster
 			targetCluster = bootstrapClusterProxy.GetWorkloadCluster(ctx, namespace, clusterName)
-			remediation()
+			testRemediation()
```
furkatgofurov7, line 84: revert please

furkatgofurov7 (changes requested):

test/e2e/e2e_test.go
fmuyassarov: this is how it was in the PR, but I can replace it

furkatgofurov7: Yes, please

furkatgofurov7 (approved):

PR #324, opened 2021-10-06
✨ Generate providerID as a combination of BMH and M3M resources

Xenwar:
Generate providerID as a combination of BareMetalHost UID and dynamic part of the Metal3Machine resource.

As we have it now, the providerID takes the value of the BMH uid, making it static for the life time of a metal3-dev-env deployment. The purpose of this PR is to make the provider ID unique per kubernetes node (CP or worker), deployment.

Verification:

1. scale out and upgrade succeeds with Legacy Label
2. scale out and upgrade succeeds with new Label
3. Given existing legacy (capm3 and kcp):
    upgrading capm3, scaling out results in legacy providerID. Also, upgrading results legacy providerID
4. Given existing legacy (capm3 and KCT):
    upgrading capm3 and KCT, scaling out results in new providerID. Also, upgrading results new providerID

PR #360, opened 2021-11-04
🌱 parameterize providerID parts in tests

Xenwar:
The providerID generation is changing by https://github.com/metal3-io/cluster-api-provider-metal3/pull/324. 
Tests also need to change. However, providerID and related objects ids are scattered throughout the tests.

This PR centralizes definition of providerID related objects. This is required before modifying existing providerID related tests.

### DIFF
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -35,6 +35,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	clientfake "k8s.io/client-go/kubernetes/fake"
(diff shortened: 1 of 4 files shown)


fmuyassarov:
/retitle 🌱 parameterize providerID parts in tests

baremetal/metal3machine_manager_test.go
```diff
@@ -55,7 +56,8 @@ const (
 	kcpName                   = "kcp-pool1"
 )
 
-var ProviderID = "metal3://12345ID6789"
+var Bmhuid = types.UID("222222-2222-2222-2222-222222222222")
```
kashifest, line 59: I know this is just an example but why not use something that looks proper?

Xenwar: Thanks, will change that.

Xenwar: Done

furkatgofurov7 (commented):
At first glance, this patch seems to be diverging from the idea that #375 is introducing or we have to take this first and rebase the 375 after?

PR #368, opened 2021-11-08
🌱 Add a function to retrieve nodes using labels

Xenwar:
Add a function to retrieve nodes using labels.
Needed for avoiding code duplication as nodes will be retrieved using new and legacy labels of the providerID.

Required by: https://github.com/metal3-io/cluster-api-provider-metal3/pull/324

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -1226,14 +1226,13 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
 		return errors.Wrap(err, "Error creating a remote client")
 	}
 
-	nodes, err := corev1Remote.Nodes().List(ctx, metav1.ListOptions{
-		LabelSelector: fmt.Sprintf("metal3.io/uuid=%v", bmhID),
-	})
+	nodeLabel := fmt.Sprintf("metal3.io/uuid=%v", bmhID)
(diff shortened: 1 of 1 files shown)


kashifest (approved):

fmuyassarov:
/retitle 🌱 Add a function to retrieve nodes using labels

baremetal/metal3machine_manager.go
```diff
@@ -1620,3 +1619,23 @@ func (m *MachineManager) getMachineSet(ctx context.Context) (*capi.MachineSet, e
 	}
 	return nil, errors.New("MachineSet is not found")
 }
+
+// getNodeWithLabel gets kubernetes nodes with a given label
```
furkatgofurov7, line 1623: /s
```suggestion
// getNodesWithLabel gets kubernetes nodes with a given label
```

furkatgofurov7 (commented):
/hold
/lgtm

This looks good, small nit but not a blocker, can be fixed in the same patch or in follow up or even in #324. Feel free to unhold if you think this should go in as is :)

kashifest (approved):
Please address @furkatgofurov7's comment

baremetal/metal3machine_manager.go
Xenwar: Thanks. fixed.

PR #372, opened 2021-11-08
🌱 Refactor gomega expect in eventually blocks

mboukhalfa:
Refactor Gomega expect in eventually blocks so when expect fails it does not break the test but continue looping in the eventually block


namnx228 (commented):
/lgtm

fmuyassarov:
/retitle 🌱 Refactor gomega expect in eventually blocks

PR #382, opened 2021-11-15
✨ Update IPAM image reference

furkatgofurov7:

Fixes #

### DIFF
--- a/config/ipam/image_patch.yaml
+++ b/config/ipam/image_patch.yaml
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/ip-address-manager:v0.1.1
+      - image: quay.io/metal3-io/ip-address-manager:latest
         name: manager


config/ipam/image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/ip-address-manager:v0.1.1
+      - image: quay.io/metal3-io/ip-address-manager:latest
```
fmuyassarov, line 11: what is that for?

fmuyassarov: isn't it better to pin the version?

furkatgofurov7: we do once we have image in quay with release version tag pinned, for now we do not have release with v1beta1 API changes so we use latest(once we release ipam v1.0.X from main branch, we should have quay image:  quay.io/metal3-io/ip-address-manager:v1.0.0 which we will reference from here)

fmuyassarov: I see. Makes sense
/lgtm

kashifest (approved):

PR #389, opened 2021-11-17
🌱 Refactor test scripts

fmuyassarov:
We have been seeing quite many flakiness in the unit test scripts. 
This is a first set of patches that will be trying to fix that flakiness.

1. When running make generate, some of the autogenerated code ends up
in github/ dir, which should be outside of staging area. This patch
adds github/ dir to .gitignore, so that we don't accidentally submit
a PR with those changes included.

2. Current `install_kubebuilder.sh` script downloads a tarball containing
other binaries apart from kubebuilder, such as etcd, kube-apiserver
and kubectl. We use kubebuilder binary only when running make generate
target, meaning there is no need to set up a test k8s env with those
binaries. Those are only needed to setup a fake k8s env while unit
testing, via make unit/test target.

3. Delete kubebuilder & kustomize installation when running go tests.
We currenrty install them on every single run of `test/unit` test
run, and in fact go test doesn't require them to perform tests.

4. Run `go mod verify` after tidy to ensure go.sum matches with
downloaded packages.

### DIFF
--- a/hack/fetch_ext_bins.sh
+++ b/hack/fetch_ext_bins.sh
@@ -26,17 +26,17 @@ if [[ -n "${TRACE}" ]]; then
   set -x
 fi
 
-k8s_version=1.19.2
-goarch=amd64
-goos="unknown"
+k8s_version=1.22.0
+arch=amd64
+os="unknown"
 
 if [[ "${OSTYPE}" == "linux"* ]]; then
-  goos="linux"
+  os="linux"
 elif [[ "${OSTYPE}" == "darwin"* ]]; then
-  goos="darwin"
+  os="darwin"
 fi
 
-if [[ "$goos" == "unknown" ]]; then
+if [[ "$os" == "unknown" ]]; then
   echo "OS '$OSTYPE' not supported. Aborting." >&2
   exit 1
 fi
@@ -58,9 +58,7 @@ function header_text {
   echo "$header$*$reset"
 }
 
-tmp_root=/tmp
-
-kb_root_dir=${tmp_root}/kubebuilder
+kb_root_dir="/tmp/kubebuilder"
 
 # Skip fetching and untaring the tools by setting the SKIP_FETCH_TOOLS variable
 # in your environment to any value:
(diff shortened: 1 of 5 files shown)


furkatgofurov7 (commented):
/lgtm

hack/tools/install_kubebuilder.sh
```diff
@@ -2,14 +2,7 @@
...
 mkdir -p ./bin
-curl -L -O "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${version}/kubebuilder_${version}_linux_${arch}.tar.gz"
-
-tar -zxvf kubebuilder_${version}_linux_${arch}.tar.gz
-mv kubebuilder_${version}_linux_${arch}/bin/* bin
-
-rm kubebuilder_${version}_linux_${arch}.tar.gz
-rm -r kubebuilder_${version}_linux_${arch}
+cd ./bin || exit
+curl -L -o kubebuilder https://go.kubebuilder.io/dl/latest/"$(go env GOOS)"/"$(go env GOARCH)"
```
kashifest, line 7: I am thinking isn't it wise to fix the version? If you download it latest always, theres a chance CI and user would have separate versions.

hack/fetch_ext_bins.sh
```diff
@@ -71,42 +69,42 @@ kb_root_dir=${tmp_root}/kubebuilder
...
-  fi
-}
+#   if [[ -z "${SKIP_FETCH_TOOLS}" ]]; then
+#     rm -rf "${kb_root_dir}"
+#   else
+#     rm -f "${kb_root_dir}/kubebuilder"
+#     rm -f "${kb_root_dir}/kubebuilder-gen"
+#     rm -f "${kb_root_dir}/vendor.tar.gz"
+#   fi
+# }
```
kashifest, line 82: why commenting out?

hack/tools/install_kubebuilder.sh
fmuyassarov: I agree that pinning a version would be much better and safer. However, I could find a URl to download only kubebuilder binary with a specific version. For example, if you replace latest with some semantic version, it won't work.

hack/fetch_ext_bins.sh
fmuyassarov: Ah, forgot to clean it. Sorry.

hack/tools/install_kubebuilder.sh
kashifest: what if we download it from github releases ? https://github.com/kubernetes-sigs/kubebuilder/releases/download/v3.2.0/kubebuilder_linux_arm64

fmuyassarov: ow, I missed that. This should work indeed! Thank you. Which version shall we pin @kashifest ?

kashifest: We havent uplifted kubebuilder to v3 for capm3. I am a bit unsure how to proceed in this. May be the latest version which support v2?

fmuyassarov: sure. [v2.3.2](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v2.3.2) seems to be the last v2 release.

kashifest (approved):

hack/tools/install_kubebuilder.sh
fmuyassarov: @kashifest Actually it is only from release v3.0.0 when only kubebuilder binary is shipped. So, if we use v2 based releases, then we will be downloading other binaries too as part of the same tarball

hack/fetch_ext_bins.sh
```diff
@@ -26,17 +26,17 @@ if [[ -n "${TRACE}" ]]; then
   set -x
 fi
 
-k8s_version=1.19.2
-goarch=amd64
-goos="unknown"
+k8s_version=1.22.0
```
furkatgofurov7, line 29: we are hardcoding this here, and is it based on the value we use somewhere or?

furkatgofurov7: okay, we pinned k8s version to this everywhere (tiltfile, e2e), I am fine with it.

furkatgofurov7 (commented):
/lgtm

fmuyassarov:
@furkatgofurov7 @kashifest 
Do we want this to be backported to release-0.5 branch ?

PR #390, opened 2021-11-17
🐛  Fix node reuse test workflow in E2E

furkatgofurov7:
This patch fixes the issue in node reuse testing workflow in e2e where:
- after testing KCP node reuse use case, we do scale down KCP to 1 and and it triggers de provisioning of BMHs to start testing MD use case, thus all BMH that were used before would have node reuse labels on them. So while testing MD case, we do not remove them and it will always chooses the same host since other BMHs have already KCP name set in node reuse label. To fix it, we remove old labels from all BMHs and then test MD use case to check if node reuse feature is working.  
- also removing marking/unmarking of BMHs with unhealthy annotation, since now when MD maxSurge field is updated with `0` it deletes the existing machine first and then recreates a new one, resulting in unneeded unhealthy annotation setting/unsetting. 
- since new m3mt has been created during remediation tests, MD was pointing to the old m3mt and tests were checking nodeReuse field in original m3mt which has no effect. This patch ensures, we do point to proper m3mt from MD first, and then scale up a worker so nodeReuse field in corresponding m3mt is respected accordingly. 
- also makes logging of starting and passing phase of the each tests more clearer with specific test name

### DIFF
--- a/test/e2e/cert_rotation_test.go
+++ b/test/e2e/cert_rotation_test.go
@@ -20,7 +20,7 @@ import (
 )
 
 func cert_rotation() {
-	Logf("Start the certificate rotation test")
+	Logf("Starting certificate rotation tests")
 	By("Check if Ironic pod is running")
 	targetClusterClientSet := targetCluster.GetClientSet()
 	ironicNamespace := os.Getenv("NAMEPREFIX") + "-system"
@@ -84,7 +84,7 @@ func cert_rotation() {
 		}
 		return errors.New("Ironic pod is not in running state")
 	}, e2eConfig.GetIntervals(specName, "wait-pod-restart")...).Should(BeNil())
-	By("PASSED!")
+	By("CERTIFICATE ROTATION TESTS PASSED!")
 
 }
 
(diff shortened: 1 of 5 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -812,36 +817,46 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmh.BareMetalHost, *p
...
 		if host.GetDeletionTimestamp() != nil {
+			m.Log.Info("Continue was hit, host DeletionTimestamp is", "host", host.Name, "GetDeletionTimestamp", host.GetDeletionTimestamp)
 			continue
 		}
 		if host.Status.ErrorMessage != "" {
+			m.Log.Info("Continue was hit, host status ErrorMessage is not nil", "host", host.Name, "ErrorMessage", host.Status.ErrorMessage)
 			continue
 		}
 		switch host.Status.Provisioning.State {
 		case bmh.StateReady, bmh.StateAvailable:
```
mboukhalfa, line 832: Can we add a log here to check the host state in case it is not ready or available

furkatgofurov7: Yeap, added.

namnx228 (commented):
/lgtm

kashifest (approved):

PR #393, opened 2021-11-18
Fix the kcp issue in e2e test

namnx228:
This PR aims to fix this issue: https://jenkins.nordix.org/view/Metal3/job/airship_master_e2e_v1b1_test_centos/10/console
Currently, when the CAPI e2e  framework checks the present of the cluster, it assumes that KCP is ready at the same time when the cluster object is ready. However, it may be not true because of the uncertainty of the underlying infra.
This PR makes the interval between two checks for the cluster present wider. The goal is when the cluster is checked and determined as "ready", the kcp object also has time to be ready.

### DIFF
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -83,7 +83,7 @@ variables:
 
 intervals:
   default/wait-controllers: ["3m", "10s"]
-  default/wait-cluster: ["20m", "10s"]
+  default/wait-cluster: ["20m", "30s"] # The second time to check the availibility of the cluster should happen late, so kcp object has time to be created
   default/wait-control-plane: ["30m", "10s"]
   default/wait-worker-nodes: ["30m", "10s"]
   default/wait-delete-cluster: ["20m", "10s"]


Rozzii (approved):

test/e2e/config/e2e_conf.yaml
```diff
@@ -83,7 +83,7 @@ variables:
 
 intervals:
   default/wait-controllers: ["3m", "10s"]
-  default/wait-cluster: ["20m", "10s"]
+  default/wait-cluster: ["20m", "15m"] # The second time to check the availibility of the cluster should happen late, so kcp object has time to be created
```
kashifest, line 86: 15 minutes??? that seems a long wait no?

namnx228: Yes, I want to make sure that KCP has enough time to be created. The 15m interval between two checks is still inside the range of 20 minutes maximum interval for the cluster wait check, so I think it is acceptable. 
Note that if there is no delay (it is in most of the case), the cluster object can be created and pass the first-time check, then it doesn't need to wait for 15 minutes. The 15-minute-wait only happens when there is a delay.

kashifest: 10s -> 15 min seems like a bug jump to me who has little idea about the use of this code, can you explain a bit more? Would it be that if the initial check of KCP readiness fails, the test will halt until KCP is ready again? And even if the KCP was ready right after the first failed check, the test will halt 15 minutes since thats when the next check is?

kashifest (approved):

test/e2e/config/e2e_conf.yaml
```diff
@@ -83,7 +83,7 @@ variables:
 
 intervals:
   default/wait-controllers: ["3m", "10s"]
-  default/wait-cluster: ["20m", "10s"]
+  default/wait-cluster: ["20m", "30s"] # The second time to check the availibility of the cluster should happen late, so kcp object has time to be created
```
furkatgofurov7, line 86:
```suggestion
  default/wait-cluster: ["20m", "30s"] # The second time to check the availability of the cluster should happen late, so kcp object has time to be created
```

furkatgofurov7 (commented):
Small nit, not a blocker. 

/lgtm

PR #397, opened 2021-11-22
🌱 Increase unit test converage for metal3labelsync controller

mboukhalfa:
Add `reconcile` and `reconcileBMHLabels` unit tests


mboukhalfa:
@kashifest 
cc @Xenwar @Arvinderpal

controllers/metal3labelsync_controller_test.go
```diff
@@ -329,6 +334,537 @@ var _ = Describe("Metal3LabelSync controller", func() {
...
+						Labels: map[string]string{
+							capi.ClusterLabelName: clusterName,
+						},
+					},
+				},
+				cluster:       newCluster(clusterName, nil, nil),
+				expectError:   true,
+				expectRequeue: true,
+			}),
+			Entry("Cluster is paused", testCaseReconcile{
```
Arvinderpal, line 611: Please correct me if I'm wrong, but this test and all the tests above it are not really specific to the label-sync controller. Is there any reason why they are necessary?

Arvinderpal (commented):
Thanks for working on this! Just had one initial comment.

controllers/metal3labelsync_controller_test.go
```diff
@@ -329,6 +334,537 @@ var _ = Describe("Metal3LabelSync controller", func() {
...
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+				}
+				if tc.expectRequeue {
+					Expect(result.Requeue || result.RequeueAfter > 0).To(BeTrue())
+				} else {
+					Expect(result.Requeue || result.RequeueAfter > 0).To(BeFalse())
+				}
+			},
+			Entry("Baremetal host not found", testCaseReconcile{}),
```
kashifest, line 413: no expectations?

controllers/metal3labelsync_controller_test.go
mboukhalfa: I am trying to cover all the cases that might hit label-sync reconcile, so any changes in that function should not break the test. As an example removing 
```
	if annotations.IsPaused(cluster, metal3Cluster) {
		controllerLog.Info("Cluster and/or Metal3Cluster are currently paused. Remove pause to continue reconciliation.")
		return ctrl.Result{RequeueAfter: bmhSyncInterval}, nil
	}
```
from lable-sync reconcile should be detected by this `Entry("Cluster is paused"` is not that needed for our unit test ?

controllers/metal3labelsync_controller_test.go
mboukhalfa: Here `expectError` is false and `expectRequeue` also false however since above we have `if true else ` they should be handled by the else in any case

controllers/metal3labelsync_controller_test.go
```diff
@@ -329,6 +334,543 @@ var _ = Describe("Metal3LabelSync controller", func() {
...
+				}
+				if tc.expectRequeue {
+					Expect(result.Requeue || result.RequeueAfter > 0).To(BeTrue())
+				} else {
+					Expect(result.Requeue || result.RequeueAfter > 0).To(BeFalse())
+				}
+				if tc.expectLabelsync {
+					node, _ := corev1Client.Nodes().Get(context.TODO(), "testNode", metav1.GetOptions{})
+					Expect(node.Labels).To(Equal(tc.host.ObjectMeta.Labels))
+				}
```
macaptain, line 415: Not a blocker, just an idea: what if instead of `tc.expectLabelsync` as a `bool`, we had `tc.expectedLabels` as a `[]string`? And we would always compare `tc.expectedLabels` against the node's actual labels for all tests? I think the main advantage of this would be that it's easy to see what labels the node really ends up with in every scenario.

macaptain (approved):
This looks great! lgtm

controllers/metal3labelsync_controller_test.go
```diff
@@ -329,6 +334,544 @@ var _ = Describe("Metal3LabelSync controller", func() {
...
+					},
+					Spec: bmh.BareMetalHostSpec{
+						ConsumerRef: &corev1.ObjectReference{
+							Name:       metal3machineName,
+							Namespace:  namespaceName,
+							Kind:       "Metal3Machine",
+							APIVersion: capm3.GroupVersion.String(),
+						},
+					},
+				},
```
furkatgofurov7, line 477: how about defining BMH the same way as `newMetalMachine` and reuse it in multiple places (at least 6-7) so we could get rid of the repetitive host structure?

furkatgofurov7 (commented):
Really nice work, overall looks good to me, some question/suggestion inline:

controllers/metal3labelsync_controller_test.go
mboukhalfa: done

controllers/metal3labelsync_controller_test.go
mboukhalfa: Thanks @furkatgofurov7 for the suggestion that has improved the code quality please check

furkatgofurov7 (approved):
/approve

/cc @namnx228 @fmuyassarov

macaptain (approved):

fmuyassarov (commented):
/lgtm

PR #400, opened 2021-11-24
Fix the flaky issue that one machine is failed instead of running

namnx228:
Currently, if the underlying infrastructure is unstable, the last check of remediation test will fail because one machine fails to be provisioned (it is in `failed` status, instead). 
```
kubectl get machine -A
NAMESPACE   NAME                    CLUSTER   NODENAME                PROVIDERID                                      PHASE     AGE   VERSION
metal3      test1-59d669478-q2mx9   test1     test1-59d669478-q2mx9   metal3://1cc3e4c3-a47c-458a-ac93-370e9554c760   Running   24h   v1.22.2
metal3      test1-8lfg4             test1     test1-8lfg4             metal3://31fb8362-f232-46ed-8f4c-d70de9b31b1d   Running   24h   v1.22.2
metal3      test1-9hrk7             test1     test1-9hrk7             metal3://96dfa070-c71f-497f-8c2c-e9218c73bbba   Running   24h   v1.22.2
metal3      test1-zxlpq             test1     test1-zxlpq             metal3://a8f24e01-44ef-48b0-9258-77f3b17aa044   Failed    24h   v1.22.2
```

This PR deletes the failed machine, so a new machine can be created and can reach the provisioned status.

### DIFF
--- a/test/e2e/remediation_test.go
+++ b/test/e2e/remediation_test.go
@@ -277,6 +277,7 @@ func test_remediation() {
 	Eventually(func(g Gomega) {
 		machines := capi.MachineList{}
 		g.Expect(bootstrapClient.List(ctx, &machines, client.InNamespace(namespace))).To(Succeed())
+		reprovisionFailedMachine(g, bootstrapClient, &machines)
 		g.Expect(filterMachinesByPhase(machines.Items, "Running")).To(HaveLen(allMachinesCount))
 	}, e2eConfig.GetIntervals(specName, "wait-machine-remediation")...).Should(Succeed())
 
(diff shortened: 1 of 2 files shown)


Rozzii (approved):
Let's hope it is neough to fix the problem.

mboukhalfa:
> Let's hope it is neough to fix the problem.

This work around can help retry provisioning after machine fail instead of failing the whole main test however more investigation to the issue can be done in separate PR

test/e2e/remediation_test.go
```diff
@@ -296,6 +297,14 @@ func (btm bmhToMachine) String() string {
...
 }
 
+func reprovisionFailedMachine(g Gomega, bootstrapClient client.Client, machines *capi.MachineList) {
+	for _, machine := range machines.Items {
+		if machine.Status.Phase == "Failed" {
+			g.Expect(bootstrapClient.Delete(ctx, &machine, &client.DeleteOptions{})).Should(Succeed())
+		}
+	}
+}
+
```
kashifest, line 307: Is it the recommended workflow or shall we find the machine failure reason and target that?

namnx228: We didn't some investigations on this issue, but no useful result was found. The reason of this failure seems to be related to the infra because @smoshiur1237 also saw the seem issue in Ansible upgrade test.

kashifest: Do we also delete the machine in ansible tests?

namnx228: I think we don't have anything in Ansible scripts to fix this issue yet since. The deletion was only made manually when the failure happened in the Ansible test.

PR #404, opened 2021-11-25
🌱 Add unit tests for remediation controller

macaptain:
Some unit tests for the `reconcileNormal` function on the remediation controller to try and describe what the expected behavior is.

**What this PR does / why we need it**: Ensures that the logic in the remediation controller is adequately covered by tests.

### DIFF
--- a/controllers/metal3remediation_controller_test.go
+++ b/controllers/metal3remediation_controller_test.go
@@ -1,0 +1,221 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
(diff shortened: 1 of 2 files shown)


controllers/metal3remediation_controller_test.go
```diff
@@ -0,0 +1,229 @@
...
+	if tc.RemediationPhase == capm3.PhaseWaiting {
+		m.EXPECT().GetTimeout().Return(&metav1.Duration{Duration: time.Minute})
+		if tc.IsTimeToRemediate {
+			m.EXPECT().TimeToRemediate(time.Minute).Return(true, time.Duration(0))
+			// If we're here, we went into PhaseWaiting because the retry limit was hit on the loop
+			// which repeatedly sets the reboot annotation.
+		} else {
+			m.EXPECT().TimeToRemediate(time.Minute).Return(false, time.Minute)
+		}
+	}
```
furkatgofurov7, line 133: how about something like, i.e:
```
if tc.RemediationPhase == capm3.PhaseWaiting || tc.RemediationPhase == capm3.PhaseRunning {
    
    if tc.IsTimeToRemediate {
	    m.EXPECT().TimeToRemediate(time.Minute).Return(true, time.Duration(0))
    } else {
	    m.EXPECT().TimeToRemediate(time.Minute).Return(false, time.Minute)
    }
}
```
so we could get rid of the lines 118-122?

furkatgofurov7 (commented):
Overall looks good, small suggestion inline:

controllers/metal3remediation_controller_test.go
macaptain: Nice suggestion, I've implemented it.

furkatgofurov7 (approved):
/test-v1b1-integration
/approve

namnx228 (commented):
/lgtm

PR #407, opened 2021-11-25
Fix the flaky issue with cert-rotation

namnx228:
This PR fixes the flaky issue: 
https://jenkins.nordix.org/job/airship_metal3io_capi_m3_v1b1_e2e_test_ubuntu/143/console 
In details, in the e2e cert rotation test, right after the pod is running, the secrets are deleted.
However, the containers in the pod may not yet execute the inotify/kill part of its entrypoint script. Maybe because of infra delay, it needs more time to reach this step. So the inotify and the-thing-need-to-be-kill are not there when the secret are deleted. That is the reason why the container doesn't restart as expected.

### DIFF
--- a/test/e2e/cert_rotation_test.go
+++ b/test/e2e/cert_rotation_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (approved):
/approve

PR #408, opened 2021-11-28
🐛Fix node reuse workflow in E2E in release-0.5 branch

furkatgofurov7:
This patch fixes the issue in node reuse testing workflow in e2e tests of release-0.5 branch where:
- after testing KCP node reuse use case, we do scale down KCP to 1 and and it triggers de provisioning of BMHs to start testing MD use case, thus all BMH that were used before would have node reuse labels on them. So while testing MD case, we do not remove them and it will always chooses the same host since other BMHs have already KCP name set in node reuse label. To fix it, we remove old labels from all BMHs and then test MD use case to check if node reuse feature is working.  
- also removing marking/unmarking of BMHs with unhealthy annotation, since now when MD maxSurge field is updated with `0` it deletes the existing machine first and then recreates a new one, resulting in unneeded unhealthy annotation setting/unsetting. 
- since new m3mt has been created during remediation tests, MD was pointing to the old m3mt and tests were checking nodeReuse field in original m3mt which has no effect. This patch ensures, we do point to proper m3mt from MD first, and then scale up a worker so nodeReuse field in corresponding m3mt is respected accordingly. 
- also makes logging of starting and passing phase of the each tests more clearer with specific test name

### DIFF
--- a/test/e2e/cert_rotation_test.go
+++ b/test/e2e/cert_rotation_test.go
@@ -19,7 +19,7 @@ import (
 )
 
 func cert_rotation() {
-	Logf("Start the certificate rotation test")
+	Logf("Starting certificate rotation tests")
 	By("Check if Ironic pod is running")
 	targetClusterClientSet := targetCluster.GetClientSet()
 	ironicNamespace := os.Getenv("NAMEPREFIX") + "-system"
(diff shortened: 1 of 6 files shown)


furkatgofurov7:
/retitle 🐛Fix node reuse workflow in E2E in release-0.5 branch
/test-v1a5-e2e

furkatgofurov7:
All retriggers are due to random server timeouts in CI)

/test-v1a5-centos-e2e

furkatgofurov7:
This is a backport of #390 with slightly more changes. 
/assign @kashifest @Xenwar 
/cc @namnx228 @fmuyassarov

namnx228 (commented):
/lgtm

kashifest (approved):

PR #412, opened 2021-11-29
🌱 Tidy up unit test running and test output

macaptain:
Includes a number of fixes to standardize the unit test running and output:

* Converts the baremetal/remote unit tests to use Ginkgo and Gomega for consistency with the rest of the project
* Use discard logger in test runs to keep the output clean
* Simplify and deduplicate test target in Makefile
* Use verbose Ginkgo argument in order to print out all spec names that are run as part of the tests.
* Run and display coverage report on build log.
* Fix the name of the Controllers test suite (previously named 'TestAPIs', now named 'TestControllers')

This is largely cosmetic. There shouldn't be any actual difference in what's being tested, but the unit test build output should be more helpful.

**What this PR does / why we need it**: While working on #404, I noticed my new tests weren't displayed in the build log. I think we should print the names of specs for the build log so that I can explicitly check that all the specs I was expecting were run and passed from the build log. In the course of fixing this, I decided to remove the log messages from mocks and standardize the 3 non-Ginkgo tests on Ginkgo.


Makefile
```diff
@@ -91,11 +91,16 @@ help:  ## Display this help
...
+		-ginkgo.v -ginkgo.noColor=$(GINKGO_NOCOLOR) \
+		-coverprofile ./cover.out; \
+	cd api; \
+	go test -v ./... \
+		-coverprofile ./cover.out
 
 .PHONY: test
-test: fmt lint ## Run tests
-	source ./hack/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./controllers/... ./baremetal/... -coverprofile ./cover.out; cd api; go test -v ./... -coverprofile ./cover.out
+test: fmt lint unit ## Run tests
```
fmuyassarov, line 103: :+1:

macaptain:
Example old build log: https://prow.apps.test.metal3.io/view/s3/prow-logs/pr-logs/pull/metal3-io_cluster-api-provider-metal3/404/unit/1465293207997779968
New build log: https://prow.apps.test.metal3.io/view/s3/prow-logs/pr-logs/pull/metal3-io_cluster-api-provider-metal3/412/unit/1465327625672069120

fmuyassarov (commented):
I think this looks much cleaner than before and more explicit on which tests it is actually running. 
/lgtm

furkatgofurov7 (commented):
This looks good, but have you checked it locally on your side, while running only unit tests locally from this branch, I was unable to scroll through up until where it started, maybe we should squeeze a bit lines the output of unit tests?
Also, this needs a rebase unfortunately.

macaptain:
> This looks good, but have you checked it locally on your side, while running only unit tests locally from this branch, I was unable to scroll through up until where it started, maybe we should squeeze a bit lines the output of unit tests? Also, this needs a rebase unfortunately.

Great point. In baremetal-operator, `make test` is not verbose. There are 'unit', 'unit-verbose' and 'unit-cover' targets, and in CAPI there's 'test' and 'test-verbose'. I'll align CAPM3's Makefile to be similar to the other projects.

I like having verbose output in the build log in Prow, so that I can Ctrl+F and sanity check that a particular test was run, so will choose to use one of the verbose targets in hack/unit.sh, if that's OK?

furkatgofurov7:
> I like having verbose output in the build log in Prow, so that I can Ctrl+F and sanity check that a particular test was run, so will choose to use one of the verbose targets in hack/unit.sh, if that's OK?

Yes, I am fine also with verbose output and agree that is easy to go through by simple search.

fmuyassarov (commented):
/lgtm

kashifest (approved):

PR #415, opened 2021-11-30
✨ Rework node reuse logic to handle corner cases

furkatgofurov7:

This PR addresses handling of some corner cases in node reuse feature, such as:
- avoids cluster deletion problem while testing stack of e2e tests that are run consequitevily by adding a DeletionTimestamp.IsZero() check just before going deep into node reuse logic to avoid fetching non existing resources if any requested while cluster deletion;
- handles m3data re-queuing in cases when machine is deleted during in place upgrade and new machine is immediately failing to get the Metal3Metadata thus breaking the in place upgrade case and also adds needed unit tests.
- improves logging in node reuse

**Note:** These changes were tested using e2e framework tests in #390 and feature tests in [metal3-dev-env/pull/855](https://github.com/metal3-io/metal3-dev-env/pull/855)

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -299,9 +299,11 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 	if host == nil {
 		host, helper, err = m.chooseHost(ctx)
 		if err != nil {
-			m.SetError("Failed to pick a BaremetalHost for the Metal3Machine",
-				capierrors.CreateMachineError,
-			)
+			if _, ok := err.(HasRequeueAfterError); !ok {
+				m.SetError("Failed to pick a BaremetalHost for the Metal3Machine",
+					capierrors.CreateMachineError,
+				)
+			}
 			return err
 		}
 		if host == nil {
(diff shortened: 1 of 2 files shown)


mboukhalfa:
LGTM thanks @furkatgofurov7  for the great work

baremetal/metal3machine_manager.go
```diff
@@ -522,60 +564,66 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+				}
+				if m3mt != nil {
+					if m3mt.Spec.NodeReuse {
+						if host.Labels == nil {
+							host.Labels = make(map[string]string)
+						}
+						// Check if machine is ControlPlane
+						if m.isControlPlane() {
+							// Fetch KubeadmControlPlane name for controlplane machine
+							m.Log.Info(fmt.Sprintf("Fetch KubeadmControlPlane name while deprovisioning host %v", host.Name))
```
fmuyassarov, line 605: Just nit: This seems a bit redundant and we have quite many logging in this if already.

baremetal/metal3machine_manager.go
```diff
@@ -522,60 +564,66 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+							kcpName, err := m.getKubeadmControlPlaneName(ctx)
+							if err != nil {
+								return err
+							}
+							m.Log.Info(fmt.Sprintf("Fetched KubeadmControlPlane name %v", kcpName))
+							// Set nodeReuseLabelName on the host to KubeadmControlPlane name
+							m.Log.Info(fmt.Sprintf("Setting nodeReuseLabelName in host %v to fetched KubeadmControlPlane name %v", host.Name, kcpName))
```
fmuyassarov, line 612: nit: perhaps these can be merged into one

baremetal/metal3machine_manager.go
```diff
@@ -522,60 +564,66 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+						} else {
+							// Fetch MachineDeployment name for worker machine
+							m.Log.Info(fmt.Sprintf("Fetch MachineDeployment name while deprovisioning host %v", host.Name))
+							mdName, err := m.getMachineDeploymentName(ctx)
+							if err != nil {
+								return err
+							}
+							m.Log.Info(fmt.Sprintf("Fetched MachineDeployment name %v", mdName))
+							// Set nodeReuseLabelName on the host to MachineDeployment name
+							m.Log.Info(fmt.Sprintf("Setting nodeReuseLabelName in host %v to fetched MachineDeployment name %v", host.Name, mdName))
```
fmuyassarov, line 623: same here with regard to merging

fmuyassarov (commented):
looks good, just some nit comments left inline

baremetal/metal3machine_manager.go
furkatgofurov7: I think this is helpful to know which BMH being deprovisioned and in which BMH we expect nodeReuse label while debugging

baremetal/metal3machine_manager.go
```diff
@@ -390,6 +392,46 @@ func (m *MachineManager) Associate(ctx context.Context) error {
...
+			m.SetError("Failed to get the BaremetalHost for the Metal3Machine",
+				capierrors.CreateMachineError,
+			)
+			return err
+		}
+
+		if err = m.setHostSpec(ctx, host); err != nil {
+			if _, ok := err.(HasRequeueAfterError); !ok {
+				m.SetError("Failed to set the BaremetalHost Specs",
+					capierrors.CreateMachineError,
```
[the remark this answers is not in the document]
furkatgofurov7: Yeah, agree, but at the same time if you check the manager code, you would find good number of `capierrors.CreateMachineError` calls in different use cases where we fail to get/set/associate/annotate certain objects, so that was the reason I followed the same pattern.

namnx228 (commented):
LGTM

furkatgofurov7:
cc @Xenwar

PR #416, opened 2021-11-30
📖 Add doc on how to tune Ironic disk cleaning feature from CAPM3

fmuyassarov:
We have exposed Ironic's automated cleaning feature as an 
API in Metal3, but have't provided clean user instructions on
how to turn on and off the feature. This patch fixes that by
adding some info about the feature and the way to switch it on/off.


docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+#### What happens if I unset automatedCleaningMode in template
+
+`automatedCleaningMode` accepts `metadata` and `disabled` or it can be unset. Please note that,
+there is no default value in CAPM3 template and machine objects for `automatedCleaningMode`. So,
+if you don't set either metadata and disabled in the template, it will be unset both in the
+template and machine objects.
+
+**Note:** There is a default value, `metadata`, in the host CR. For that reason, even if it is
+unset in the template and machine objects, Ironic node will still face automated cleaning because
+the host has enabled it by its default value.
```
furkatgofurov7, line 115: This part is a bit unclear. So, if m3mt takes precedence over m3m and if the field is unset in m3mt, that means m3m would have `automatedCleaningMode` field also unset. And as you outlined m3m has also precedence over BMH, CAPM3 machine controller will copy value from a machine to its corresponding BMH, but if BMH has default value, should not it also overriden and made unset?

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+1. Since unset in the machines, hosts will get its default value `metadata`.
+
+1. Now we have two hosts, both cleaning enabled. Since `automatedCleaningMode` is unset in the template,
+    you can set different cleaning modes for machines in the same pool. In other words, replicating the
+    value of `automatedCleaningMode` from template to machines is blocked because it is unset in the template.
+    This gives freedom to users to configure cleaning differently and selectively for each node regardless
+    of the fact that they belong to the same template.
+
+1. You can edit one of the machine to set `automatedCleaningMode` to `disabled`. Once edited,
+    CAPM3 machine controller replicates the value from the machine to its owning host. As such, you end with
```
furkatgofurov7, line 139:
```suggestion
    CAPM3 machine controller replicates the value from the machine to its owning host. As such, you end up with
```

furkatgofurov7 (commented):
looks good, few suggestions below:

docs/disk_cleaning.md
fmuyassarov: > And as you outlined m3m has also precedence over BMH, CAPM3 machine controller will copy value from a machine 
> to its corresponding BMH, but if BMH has default value, should not it also overriden and made unset?
>

No it should not unset on the host, because then you will not be able to selectively control the cleaning mode per node.

Machine takes precedence over the host, only if `automatedCleaningMode` is not unset, in all other cases the value will be copied from machine to the host.

fmuyassarov: No, it should not, because in that you case there is no way to selectively/individually switch on/off the cleaning. If unset, that means fall back to the default in BMH.

Machine takes precedence over a host only when `automatedCleaningMode` is unset, in all other cases machine controller ensures that machine and the host have the same config.

fmuyassarov:
@furkatgofurov7 PTAL, addresses comments. Thanks.

furkatgofurov7 (approved):
/approve

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+- BareMetalHost - host
+
+## What is automated cleaning
+
+Ironic offers two modes for node cleaning, automated and manual where automated
+cleaning is recommended for ironic deployments. Automated cleaning gets triggered
+automatically during the first provisioning and on every deprovisioning of a node.
+If automated cleaning is enabled, Ironic runs various cleaning steps by priority
+and meanwhile the node stays in `cleaning` state (machine/host in
+`deprovisioning`). To get more info, check Ironic's [documentation](https://docs.openstack.org/ironic/latest/admin/cleaning.html#automated-cleaning).
```
kashifest, line 31: If "Automated cleaning gets triggered automatically during the first provisioning" then is this line correct "(machine/host in `deprovisioning`)" , shouln't it also be provisioning ?

fmuyassarov: Cleaning takes place on every deprovisioning (active -> available) or first registration (manageable -> available).  As such, I don't think BMH can be in `provisioning` state when Ironic node is in `cleaning` state.
Reference to the Ironic host state machine:  https://docs.openstack.org/ironic/latest/admin/cleaning.html#automated-cleaning

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+even though they are in the same pool (i.e. referencing the same template). To achieve that:
+
+1. Don't set `automatedCleaningMode` in the template. If unset in the template, it will be unset in
+    machines too.
+
+1. Since unset in the machines, hosts will get its default value `metadata`.
+
+1. Now we have two hosts, both cleaning enabled. Since `automatedCleaningMode` is unset in the template,
+    you can set different cleaning modes for machines in the same pool. In other words, replicating the
+    value of `automatedCleaningMode` from template to machines is blocked because it is unset in the template.
```
kashifest, line 134: machine?

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+    This gives freedom to users to configure cleaning differently and selectively for each node regardless
+    of the fact that they belong to the same template.
+
+1. You can edit one of the machine to set `automatedCleaningMode` to `disabled`. Once edited,
+    CAPM3 machine controller replicates the value from the machine to its owning host. As such, you end up
```
kashifest, line 139: I would rather use metal3machine controller instead of CAPM3

kashifest (changes requested):
Some questions/concerns inline

docs/disk_cleaning.md
kashifest: but my point is can it only be in deprovisioning state?

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+
+```sh
+[conductor]
+automated_clean=false
+```
+
+Disadvantages:
+
+- once Ironic is deployed, changing the value of `automated_clean` requires rebuilding of
+    new Ironic container image. Thus Ironic container will need to be down until the new
```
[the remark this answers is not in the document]
fmuyassarov: General Ironic, basically whatever you get out from ironic-image repository.

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+If you are using CAPI and CAPM3 and on top of the Baremetal Operator, then you can and should
+control the cleaning from CAPM3 CR, specifically template.
+
+Although, you can edit `automatedCleaningMode` value from machine or host, we
+recommend doing it from template, because template controller always
+tries to replicate the value of `automatedCleaningMode` from the template to the machine. In
+other words, template takes precedence over machine, and machine's value for
+`automatedCleaningMode` will be overridden by the value in template.
+
+If you are wondering why `automatedCleaningMode` exists in machine too, that's because
```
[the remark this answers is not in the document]
fmuyassarov: machine == metal3machine. At the beginning of the document,  I've mentioned how we refer to different objects throughout  the document.

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+As such, by default you can not have different configurations for two nodes if they are under the same
+pool. They both will have the same value, either `metadata` or `disabled`.
+
+However, it is possible to have cleaning disabled for one of the machine, while enabled for the other,
+even though they are in the same pool (i.e. referencing the same template). To achieve that:
+
+1. Don't set `automatedCleaningMode` in the template. If unset in the template, it will be unset in
+    machines too.
+
+1. Since unset in the machines, hosts will get its default value `metadata`.
```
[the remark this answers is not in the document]
fmuyassarov: That's the feature of markdown. If you just put 1, you don't have manually put numbers, markdown will automatically convert them to 1, 2, 3 in its order for you.

docs/disk_cleaning.md
fmuyassarov: please, see my comment below for this

docs/disk_cleaning.md
fmuyassarov: sure, I can fix it

docs/disk_cleaning.md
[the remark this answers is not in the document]
kashifest: So when you say ironic container will need to be down, there is no ironic container, it should be ironic pod or ironic container`s` then.

docs/disk_cleaning.md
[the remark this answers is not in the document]
kashifest: Missed it. Its tough to follow this way btw, an m3m instead of machine is still better and less writing IMO.

docs/disk_cleaning.md
fmuyassarov: not only, but I don't know to be honest if it is preparing state or inspecting state when it is first time

docs/disk_cleaning.md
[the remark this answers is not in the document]
fmuyassarov: check the actual file not the raw format

docs/disk_cleaning.md
fmuyassarov: Let's remove that part not to confuse ourself + readers

kashifest:
I am satisfied with the changes. Thanks @fmuyassarov 
/hold cancel

docs/disk_cleaning.md
```diff
@@ -0,0 +1,140 @@
...
+    - [Via ironic.conf](#via-ironicconf)
+    - [Via CAPM3 APIs](#via-capm3-apis)
+      - [Which resources do I need to edit](#which-resources-do-i-need-to-edit)
+      - [When should I edit my CRs to disable cleaning](#when-should-i-edit-my-crs-to-disable-cleaning)
+      - [Selective configuration](#selective-configuration)
+
+How we refer to these objects later on:
+
+- Metal3Machine - m3m
+- Metal3Machinem3mt - m3mt
```
furkatgofurov7, line 21: typo

furkatgofurov7 (changes requested):
/hold
I can lgtm once again once it is fixed

furkatgofurov7 (approved):
Since this patch has been reviewed and had both labels initially, but due to a minor change(typo fix) labels were removed. Adding them back and merging it.
 
/approve
/lgtm

PR #417, opened 2021-11-30
🌱 e2e upgrade cp test v1a5

mboukhalfa:
**What this PR does / why we need it:** Adds e2e upgrade control-plane components test in v1alpha5.


mboukhalfa:
closed in favor of having a generic test considering both major and minor upgrades

PR #424, opened 2021-12-03
🐛 Require image-spec v1.0.2 

kashifest:
Address security  vulnerability GHSA-77vh-xpmg-72qh


fmuyassarov (commented):
/lgtm

go.mod
```diff
@@ -33,3 +33,5 @@ require (
 replace github.com/metal3-io/cluster-api-provider-metal3/api => ./api
 
 replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v0.4.4
+
+require github.com/opencontainers/image-spec v1.0.2 // indirect
```
furkatgofurov7, line 37: qq: Why not to add it in the **require** list?

kashifest: Wanted to make it separate so that its easy to remove in future

PR #428, opened 2021-12-07
🐛 Fix CAPI release fetching for Tilt environment

furkatgofurov7:
While testing Tilt development environment got to know that we are still fetching old CAPI release in `gen_tilt_settings.sh`. So, this patch fixes that and also updates: 
- cert-manager version to v1.5.3
- k8s version to v1.22.3


Tiltfile
```diff
@@ -12,9 +12,9 @@ settings = {
     "deploy_cert_manager": True,
     "preload_images_for_kind": True,
     "kind_cluster_name": "capm3",
-    "capi_version": "v1.0.0",
-    "cert_manager_version": "v1.5.0",
-    "kubernetes_version": "v1.22.0",
+    "capi_version": "v1.0.2",
+    "cert_manager_version": "v1.5.3",
+    "kubernetes_version": "v1.22.3",
```
kashifest, line 17: is there anyway to avoid this hard coding?

Xenwar: I think there are many more such values. They can be handled on a separate PR.

furkatgofurov7: Done for CAPI/K8S version, but since cert manager version is hardcoded in https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/tilt_modules/cert_manager/Tiltfile#L29 and I believe we have to provide the same version here to generate tilt-settings.json. Btw, to my surprise other providers also pinning the cert-manager version as we do https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/Tiltfile#L20

kashifest (approved):

PR #357, opened 2021-10-29
✨ Add base Conditions to Metal3Cluster and Metal3Machine Status

Arvinderpal:
**What this PR does / why we need it**:
Adds Conditions to the Status object for both Metal3Cluster and Metal3Machine. 

Most, if not all, CAPI providers utilize Conditions to provide granular status information. They also enable devs/users to problematically probe the state of an object. For example, has the Metal3Machine been successfully as associated with a BMH?

This PR adds some basic conditions and also by introducing the Conditions field it allows additional Conditions to be added over time. 

Fixes # https://github.com/metal3-io/metal3-docs/issues/127

Related:
This approach closely follows the CAPI approach for Conditions. See for example: https://github.com/kubernetes-sigs/cluster-api/issues/3486

PR #430, opened 2021-12-08
🌱 Bump code-generator to v0.23.0

fmuyassarov:
Fixes the issues mentioned in https://github.com/metal3-io/cluster-api-provider-metal3/pull/357#issuecomment-988278632.


hack/tools/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/golangci/golangci-lint v1.43.0
 	github.com/jteeuwen/go-bindata v3.0.7+incompatible
-	k8s.io/code-generator v0.22.2
+	k8s.io/code-generator v0.23.0
```
furkatgofurov7, line 10: should not this be `v0.23.0-alpha.4` as per linked comment in PR description?

furkatgofurov7 (commented):
Also there is an empty `test.txt` file in the commit which most probably added accidentally

hack/tools/go.mod
fmuyassarov: The release of 1.23.0 is out now, so I think that's okay?

kashifest (approved):

hack/tools/go.mod
furkatgofurov7: Okay

furkatgofurov7 (commented):
/lgtm

PR #431, opened 2021-12-08
🐛 Automated cleaning bug fix in machine template controller

fmuyassarov:
Due to this bug,  `automatedCleaning` will not be synchronized between
metal3MachineTemplate and metal3Machine when set. Because we are
verifying it from the metal3Machine which should be the metal3MachineTemplate.


kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #436, opened 2021-12-13
🌱 Use golang image as base for unit test containers

fmuyassarov:
Use `docker.io/golang:1.16`  container image for unit.sh script.
We don't necessarily need to create & use `quay.io/metal3-io/capm3-unit`
image, because the only thing this image has is kustomize and
kubebuilder installed via below RUN, which is redundant as those
binaries will be installed during `unit` Makefile target ( as part of
`setup_envs`). At the end, we currently install kubebuilder and
kustomize twice. 
```
 RUN ./install_kustomize.sh && \
    ./install_kubebuilder.sh

```
There will be a follow PR to this, which will remove the
Dockerfile, and update the project-infra respectively to migrate
fully to golang images.

### DIFF
--- a/hack/unit.sh
+++ b/hack/unit.sh
@@ -9,7 +9,6 @@ if [ "${IS_CONTAINER}" != "false" ]; then
   export XDG_CACHE_HOME=/tmp/.cache
   mkdir /tmp/unit
   cp -r . /tmp/unit
-  cp -r /usr/local/kubebuilder/bin /tmp/unit/hack/tools
   cd /tmp/unit
   make unit-cover-verbose
 else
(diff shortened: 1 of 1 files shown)


kashifest (commented):
Isn't changing image enough? Why changing other things?

fmuyassarov:
> Isn't changing image enough? Why changing other things?

Also, the mounting doesn't have to be read only, because
make unit-cover-verbose is not expected to modify files,
but only test them.

fmuyassarov:
/test-v1b1-integration 
@kashifest I've reverted that part and kept the rest. PTAL
/cc @furkatgofurov7

kashifest (approved):

furkatgofurov7 (commented):
/lgtm

furkatgofurov7 (commented):
/lgtm

PR #437, opened 2021-12-13
🌱 Drop capm3:unit container image Dockerfile

fmuyassarov:
Drop building capm3:unit image and instead use golang container
image for go tests in Prow. The reason for dropping this image
support is,  it has no benefit anymore but instead it brings extra
things like kubebuilder and kustomize which are not needed to be
build by shell scripts anymore because we donwload those bins via
setup_env Makefile target. This should be merged after the update
in the project-infra repo.
/hold
/cc @furkatgofurov7 @kashifest

### DIFF
--- a/hack/Dockerfile.unit
+++ b/hack/Dockerfile.unit
@@ -1,6 +1,0 @@
-FROM docker.io/golang:1.17
-
-WORKDIR /usr/local/kubebuilder
-COPY /hack/tools /usr/local/kubebuilder/
-RUN ./install_kustomize.sh && \
-    ./install_kubebuilder.sh


kashifest:
> **What this PR does / why we need it**: Drop building capm3:unit image and instead use golang container image for go tests in Prow. The reason for dropping this image support is, it has no benefit anymore but instead it brings extra things like kubebuilder and kustomize which are not needed to be build by shell scripts anymore because we donwload those bins via setup_env Makefile target. This should be merged after the update in the project-infra repo. /hold /cc @furkatgofurov7 @kashifest

You don't need to wait for any changes in Project-infra since the image is already there in quay and removing this wont break anything.

kashifest:
/approve
/lgtm
This Dockerfile is not used not anymore.

PR #451, opened 2021-12-23
✨ Pin calico version to v3.21.0

namnx228:
This PR pins the calico version used ine2e test to v3.21.0 which is the same as the default calico version in Metal3-dev-env. Pinning calico version helps to avoid the docker rate limit issue because in CI this calico version is baked into the node image.

### DIFF
--- a/hack/e2e/environment.sh
+++ b/hack/e2e/environment.sh
@@ -54,3 +54,7 @@ source "${M3_DEV_ENV_PATH}/scripts/feature_tests/feature_test_vars.sh"
 # shellcheck disable=SC1091
 # shellcheck disable=SC1090
 source "${M3_DEV_ENV_PATH}/scripts/feature_tests/node_reuse/node_reuse_vars.sh"
+
+# Pin Calico version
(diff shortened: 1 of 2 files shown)


Rozzii (approved):

fmuyassarov (commented):
I don't see the patching here. When download of YAML happens, it is either the latest from the minor release or the first. To use 3.21.0, we patch/replace it in metal3-dev-env https://github.com/metal3-io/metal3-dev-env/blob/25b0041fadb24471a09691ce3d29e666f66ab1af/vm-setup/roles/v1aX_integration_test/tasks/verify.yml#L68

furkatgofurov7:
> this PR faces quite often flaky error with e2e node-reuse test.

From the CI output, looks like remediation tests are also failing and at first view seems like we are still facing the original problem (calico pods are not running properly and crashing) which this Pr tries to fix?

mboukhalfa:
LGTM

furkatgofurov7 (approved):

PR #454, opened 2021-12-27
🌱 Don't set error on m3machine when it is not a terminal problem

fmuyassarov:
Do not set temporary errors in status.FailureMessage and Status.FailureReason
of Metal3Machine. Otherwise, due to such a temporary error referenced CAPI
Machine is set to Failed which is considered a terminal state that can only
be fixed by only deleting the Machine.

[`FailureReason`](https://github.com/metal3-io/cluster-api-provider-metal3/blob/3542693bea3af543f43aede404dfbb83d6915872/api/v1beta1/metal3machine_types.go#L118) & [`FailureMessage`](https://github.com/metal3-io/cluster-api-provider-metal3/blob/3542693bea3af543f43aede404dfbb83d6915872/api/v1beta1/metal3machine_types.go#L137) should be set only when there is a terminal
problem reconciling the metal3machine and will contain a succinct value suitable
for machine interpretation.

This field should not be set for transitive errors that a controller faces that are
expected to be fixed automatically over time (like service outages), but instead
indicate that something is fundamentally wrong with the metal3machine's spec
or the configuration of the controller, and that manual intervention is required.
Examples of terminal errors would be invalid combinations of settings in the
spec, values that are unsupported by the controller, or the responsible controller
itself being critically misconfigured.

Any transient errors that occur during the reconciliation of metal3machines can be
added as events to the metal3machine object and/or logged in the controller's output.

Also, I've noticed that some other infra providers like AWS, GCP, Azure are not also
setting an error when there is an error during cluster fetching.

I didn't remove [setErrorM3Machine](https://github.com/metal3-io/cluster-api-provider-metal3/blob/3542693bea3af543f43aede404dfbb83d6915872/controllers/metal3machine_controller.go#L523) because I would like to know first if I would
introduce any potential path for bugs with this patch.


controllers/metal3machine_controller_integration_test.go
```diff
@@ -321,20 +321,6 @@ var _ = Describe("Reconcile metal3machine", func() {
...
-				Objects: []client.Object{
-					metal3machineWithOwnerRefs(),
-					machineWithInfra(),
-				},
-				ErrorExpected:       true,
-				ErrorReasonExpected: true,
-				ErrorReason:         capierrors.InvalidConfigurationMachineError,
-				RequeueExpected:     false,
-			},
-		),
```
lentzi90, line 337: Would it be possible to keep the test case but change it to expect no error?

controllers/metal3machine_controller.go
```diff
@@ -519,14 +505,6 @@ func (r *Metal3MachineReconciler) Metal3DataToMetal3Machines(obj client.Object)
 	return requests
 }
 
-// setError sets the ErrorMessage and ErrorReason fields on the metal3machine
```
Xenwar, line 522: This is a generic function for setting error messages and should not be removed.
If error should not be set as is indicated by the title, then one can set to the fields to nil/"".

controllers/metal3machine_controller.go
```diff
@@ -115,20 +114,7 @@ func (r *Metal3MachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-		return ctrl.Result{}, errors.Wrapf(err, "Metal3Machine's owner Machine is missing label or the cluster does not exist")
-	}
-	if cluster == nil {
-		setErrorM3Machine(capm3Machine, fmt.Sprintf(
-			"The machine is NOT associated with a cluster using the label %s: <name of cluster>",
-			capi.ClusterLabelName,
-		), capierrors.InvalidConfigurationMachineError)
-		machineLog.Info(fmt.Sprintf("Please associate this machine with a cluster using the label %s: <name of cluster>", capi.ClusterLabelName))
+		machineLog.Info("Machine is missing cluster label or cluster does not exist")
 		return ctrl.Result{}, nil
```
kashifest, line 118: shall we use this now instead return `ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil`

controllers/metal3machine_controller.go
kashifest: Good point @Xenwar . Also if we are removing this function, there isn't any point to leave the 	`clearErrorM3Machine(capm3Machine)` function also?

Xenwar: If it helps my case, there is a similar setup on AWS
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a519d67fbcbd5e50b517576e06d578690f6b70a7/pkg/cloud/services/eks/cluster.go#L134-L142

controllers/metal3machine_controller.go
fmuyassarov: Sorry if I didn't understand. Are you suggesting to add re-queuing here?

controllers/metal3machine_controller.go
fmuyassarov: Thanks @Xenwar for pointing it out. I agree with you. I removed it because there is no use of it, and golinter doesn't allow to keep unused func(). I would personally would not set any error instead of setting nil, but I don't have strong opinion which way is better. As you suggested, we can set the message & error capi error type to nil. As such, no func() removal and no linter complains.

fmuyassarov: @Xenwar @kashifest PTAL updated accordingly

controllers/metal3machine_controller.go
kashifest: yes.

fmuyassarov: Perhaps no, after https://github.com/metal3-io/cluster-api-provider-metal3/pull/454#discussion_r777348701 because we are not setting an actual error.

kashifest (approved):

furkatgofurov7:
Not a blocker for this to go in, but @fmuyassarov as we discussed this offline, are we confident enough this is also tackling the failure that we were facing in feature tests and e2e tests, where during the scaling up of machines, the last one of them ends up in "Failed" state and has an error i.e:

`Failure detected from referenced resource infrastructure.cluster.x-k8s.io/v1beta1, Kind=Metal3Machine with name "test1-controlplane-xxx": failed to associate the Metal3Machine to a BaremetalHost` 

set on it by m3machine controller [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/bdfc763d574b16df00fb59ce09c40ec1c0f848d2/controllers/metal3machine_controller.go#L246)?

fmuyassarov:
> Not a blocker for this to go in, but @fmuyassarov as we discussed this offline, are we confident enough this is also tackling the failure that we were facing in feature tests and e2e tests as flake, where during the scaling up of machines, the last one of them ends up in "Failed" state and has an error i.e:
> 
> `Failure detected from referenced resource infrastructure.cluster.x-k8s.io/v1beta1, Kind=Metal3Machine with name "test1-controlplane-xxx": failed to associate the Metal3Machine to a BaremetalHost`
> 
> set on it by m3machine controller [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/bdfc763d574b16df00fb59ce09c40ec1c0f848d2/controllers/metal3machine_controller.go#L246)?

I'm not even sure if that issue is similar to what this patch is trying to fix. I've been trying to reproduce it in non e2e case, but I could not tbh.

controllers/metal3machine_controller.go
Xenwar: would it be helpful to re-add the logging part ?
We keep the log but ignore the result as is done now so that we visibility towards the logic.
https://github.com/metal3-io/cluster-api-provider-metal3/pull/454/files#diff-250f2503d81624c7304619762d56ee502ff1462e5ad0c965af1eb9bfed5f340cL118

fmuyassarov: We still have it, it is moved to https://github.com/metal3-io/cluster-api-provider-metal3/pull/454/files#diff-250f2503d81624c7304619762d56ee502ff1462e5ad0c965af1eb9bfed5f340cR119

Xenwar:
Could not remember why we decided to put a hold during the weekly meeting. But, based on the comments, removing it.
/hold cancel

PR #455, opened 2022-01-03
🐛 Rewrite MachineDeployment/MachineSet tests in m3machinemanager

furkatgofurov7:
Rewrites nodeReuse related MachineDeployment/MachineSet tests in m3machinemanager

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -95,10 +95,12 @@ type MachineManager struct {
 
 	Cluster               *capi.Cluster
 	Metal3Cluster         *capm3.Metal3Cluster
+	MachineList           *capi.MachineList
 	Machine               *capi.Machine
 	Metal3Machine         *capm3.Metal3Machine
 	Metal3MachineTemplate *capm3.Metal3MachineTemplate
-	MachineSetList        []*capi.MachineSet
+	MachineSet            *capi.MachineSet
+	MachineSetList        *capi.MachineSetList
 	Log                   logr.Logger
 }
 
@@ -121,13 +123,13 @@ func NewMachineManager(client client.Client,
 
 // NewMachineSetManager returns a new helper for managing a machineset
 func NewMachineSetManager(client client.Client,
-	machine *capi.Machine, machinesetlist []*capi.MachineSet,
-	machineset *capi.MachineSet, machineLog logr.Logger) (*MachineManager, error) {
+	machine *capi.Machine, machineSetList *capi.MachineSetList,
+	machineLog logr.Logger) (*MachineManager, error) {
 
 	return &MachineManager{
 		client:         client,
 		Machine:        machine,
-		MachineSetList: machinesetlist,
+		MachineSetList: machineSetList,
 		Log:            machineLog,
 	}, nil
 }
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -278,59 +278,23 @@ func m3mObjectMetaNoAnnotations() *metav1.ObjectMeta {
 	}
 }
 
-// func machineOwnerRefToMachineSet() *capi.Machine {
-// 	return &capi.Machine{
-// 		ObjectMeta: metav1.ObjectMeta{
-// 			OwnerReferences: []metav1.OwnerReference{
-// 				{
-// 					APIVersion: capi.GroupVersion.String(),
-// 					Kind:       "MachineSet",
-// 					Name:       "ms-test1",
-// 				},
-// 			},
-// 		},
-// 	}
-// }
-
-// func machineSetsList() []*capi.MachineSetList {
-// 	return []*capi.MachineSetList{
-// 		{
-// 			TypeMeta: metav1.TypeMeta{
-// 				APIVersion: capi.GroupVersion.String(),
-// 				Kind:       "MachineSet",
-// 			},
-// 			ListMeta: metav1.ListMeta{},
-// 			Items: []capi.MachineSet{
-// 				{
-// 					TypeMeta: metav1.TypeMeta{},
-// 					ObjectMeta: metav1.ObjectMeta{
-// 						Name: "test1",
-// 						OwnerReferences: []metav1.OwnerReference{
-// 							{
-// 								APIVersion: capi.GroupVersion.String(),
-// 								Kind:       "MachineDeployment",
-// 								Name:       "test1",
-// 							},
-// 						},
-// 					},
-// 				},
-// 				{
-// 					TypeMeta: metav1.TypeMeta{},
-// 					ObjectMeta: metav1.ObjectMeta{
-// 						Name: "test2",
-// 					},
-// 				},
-// 				{
-// 					TypeMeta: metav1.TypeMeta{},
-// 					ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


baremetal/metal3machine_manager_test.go
```diff
@@ -280,54 +280,19 @@ func m3mObjectMetaNoAnnotations() *metav1.ObjectMeta {
 
 func machineOwnerRefToMachineSet() *capi.Machine {
```
macaptain, line 281: I was a bit surprised that a function named `...ToMachineSet()` returns a `Machine` instead of a `MachineSet`, it sounds like a casting function from the name. Then I looked a bit closer and understood. For more clarity, what do you think about naming it `machineOwnedByMachineSet` ?

baremetal/metal3machine_manager_test.go
```diff
@@ -3472,38 +3437,84 @@ var _ = Describe("Metal3Machine manager", func() {
...
-			for ms := range machineSetObjects.Items {
-				tc.expectedMachineSet = &machineSetObjects.Items[ms]
-				Expect(result).To(Equal(tc.expectedMachineSet))
+			for ms := range tc.MachineSetList.Items {
+				foundMS := &tc.MachineSetList.Items[ms]
+				key := client.ObjectKey{
+					Name:      foundMS.Name,
+					Namespace: foundMS.Namespace,
+				}
+				err = c.Get(context.TODO(), key, foundMS)
```
macaptain, line 3467: I'm not sure we're actually checking anything here? We already know that the fake client is loaded with these objects because they were added in line 3443-3448 above.

baremetal/metal3machine_manager_test.go
```diff
@@ -3517,129 +3528,312 @@ var _ = Describe("Metal3Machine manager", func() {
...
-			for ms := range machineSetObjects.Items {
-				tc.expectedMachineSet = &machineSetObjects.Items[ms]
-				Expect(result).To(Equal(tc.expectedMachineSet))
+				for i := range tc.MachineSetList.Items {
+					foundMS := &tc.MachineSetList.Items[i]
+					key := client.ObjectKey{
+						Name:      foundMS.Name,
+						Namespace: foundMS.Namespace,
+					}
+					err = c.Get(context.TODO(), key, foundMS)
```
macaptain, line 3645: Again, here, I'm not sure we're testing anything by running `c.Get` on the fakeclient?

baremetal/metal3machine_manager_test.go
```diff
@@ -3472,38 +3437,84 @@ var _ = Describe("Metal3Machine manager", func() {
...
 
 			result, err := machineMgr.getMachineDeploymentName(context.TODO())
 			if tc.expectError {
 				Expect(err).To(HaveOccurred())
+				Expect(tc.expectedMachineSet).To(BeNil())
 			} else {
 				Expect(err).NotTo(HaveOccurred())
 			}
-			machineSetObjects := capi.MachineSetList{}
-			for ms := range machineSetObjects.Items {
```
macaptain, line 3494: Thanks for fixing this! These tests passed trivially!

macaptain (commented):
Great, thanks for fixing the tests so that they are actually testing something. 

Two minor nits: one function name that made me double take. Also testing that the fakeclient can find the objects we loaded into it doesn't seem necessary to me.

baremetal/metal3machine_manager_test.go
furkatgofurov7: @macaptain that is to check if can get a MachineSet I want in the same namespace and with the name. Lines 3443-3448 above are adding items to the MachineSetList.

baremetal/metal3machine_manager_test.go
furkatgofurov7: Please check above comment

baremetal/metal3machine_manager_test.go
furkatgofurov7: Agree, fixed as suggested!

furkatgofurov7:
@macaptain thanks a lot for the review, addressed your comments and answers to questions.

baremetal/metal3machine_manager_test.go
macaptain: Lines 3443-3448 add the MachineSetList Items specified in the test case to the fakeClient:

https://github.com/metal3-io/cluster-api-provider-metal3/blob/e1cb70ff290868eb8913d37cfa947da14fa57a5d/baremetal/metal3machine_manager_test.go#L3444-L3445

Then we're looping over the same list `tc.MachineSetList.Items` and checking we can fetch them out of the fakeClient with `c.Get`. But we know we can: we put them there, so I don't see how the `err` can ever be hit. We're looping over the same list twice.

I am not convinced these lines can ever fail, no matter what is in the test case, but not a blocker for me, just an observation, the tests themselves are still good. Let me know what you think!

furkatgofurov7: I have checked it a bit closer, and indeed, I have overlooked it. Initially, when started writing the tests, the structure of them was a bit different and I was expecting  `foundMS` to be passed to `Expect(result).To(Equal(tc.expectedMDName))`, but that is not the case now, I have changed them later, so that was left out and I removed it here and on the next tests. Thanks for pointing it out.

macaptain (approved):

smoshiur1237 (approved):
Thanks for rewriting the test.
/lgtm

baremetal/metal3machine_manager.go
```diff
@@ -95,10 +95,12 @@ type MachineManager struct {
 
 	Cluster               *capi.Cluster
 	Metal3Cluster         *capm3.Metal3Cluster
+	MachineList           *capi.MachineList
 	Machine               *capi.Machine
 	Metal3Machine         *capm3.Metal3Machine
 	Metal3MachineTemplate *capm3.Metal3MachineTemplate
-	MachineSetList        []*capi.MachineSet
+	MachineSet            *capi.MachineSet
+	MachineSetList        *capi.MachineSetList
```
kashifest, line 103: You are removing the array/List `[]`, so is it not a list anymore ?

baremetal/metal3machine_manager_test.go
```diff
@@ -3517,129 +3519,304 @@ var _ = Describe("Metal3Machine manager", func() {
 			expectedMD:     true,
 			expectedMDName: "md-test1",
 		}),
+		Entry("Should not find the expected MachineDeployment name, MachineSet OwnerRef Kind is not correct", testCaseGetMachineDeploymentName{
+			Machine: machineOwnedByMachineSet(),
+			MachineSetList: &capi.MachineSetList{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: capi.GroupVersion.String(),
+					Kind:       "MachineSet",
```
kashifest, line 3527: Is this Kind correct? Shouldn't it be MachineSetList ?

baremetal/metal3machine_manager.go
furkatgofurov7: Yes, it is struct and it has MachineSet list field inside

baremetal/metal3machine_manager_test.go
furkatgofurov7: thanks, fixed

kashifest (approved):

PR #459, opened 2022-01-04
🌱Remove static calico files from E2E tests

furkatgofurov7:
In e2e tests, currently, we fetch calico manifests on the fly and make corresponding changes to the manifest based on the OS, thus making static calico manifests redundant. This PR removes those static files and tweaks e2e suite tests accordingly.


furkatgofurov7:
/retitle 🌱Remove static calico files from E2E tests
/test-v1b1-integration

kashifest (approved):

PR #465, opened 2022-01-11
🌱 Use generic folder name for all volume mounted folders under hack scripts

furkatgofurov7:
We folder names that would be used eventually while mounting a volume inside the container while using hack/scripts are scattered all over the repo (in some places used as `workdir` and in some `go/src/github.com/metal3-io/cluster-api-provider-metal3`). This PR cleans that a bit and uses a single folder name under all over the hack scripts. 

Similar changes have been done in [IPAM](https://github.com/metal3-io/ip-address-manager/pull/80) repo some time ago.


lentzi90 (approved):
LGTM!

kashifest (approved):

Issue #319, opened 2021-10-05
Add E2E upgrade tests for CAPM3 v1alpha5 -> v1beta


furkatgofurov7:
**Detailed Description**

 With the new upgrade path being enabled for CAPI, we want to add the tests in e2e to upgrade from v1alpha5 to v1beta1

**Anything else you would like to add:**

[Miscellaneous information that will assist in solving the issue.]

/kind feature

PR #466, opened 2022-01-12
🌱 Add v1alpha5 to v1beta e2e upgrade tests

mboukhalfa:
This PR adds upgrade test from v1alpha5 to v1beta1 by:

1. Run dev-env which creates a bootstrap cluster with the version v1beta1
2. Create a target cluster using the e2e framework
3. Init the created cluster with the old clusterctl to run v1a4 capi and v1a5 capm3
4. Pivot Ironic/BMO and resource from bootstrap to the created cluster 
5. Test this old management cluster by using it to create a workload cluster 
6. Upgrade the management cluster to v1beta1

Fixes: #319 

TODO:

- [ ] Add the possibility to parameterize the upgrade [from/to] version including minor version


scripts/ci-e2e.sh
```diff
@@ -89,8 +89,15 @@ source "${REPO_ROOT}/hack/ensure-kustomize.sh"
 
 # This will run the tests with env variabls defined in environment.sh
 # or exported by metal3-dev-env scripts
-${M3_DEV_ENV_PATH}/scripts/run_command.sh make e2e-tests
+${M3_DEV_ENV_PATH}/scripts/run_command.sh make e2e-tests || sleep infinity
```
Rozzii, line 92: don't forget to remove this after it goes from draft to regular PR

scripts/ci-e2e.sh
```diff
@@ -89,8 +89,15 @@ source "${REPO_ROOT}/hack/ensure-kustomize.sh"
...
 pushd ${M3_DEV_ENV_PATH} || exit 1
 make clean
 popd || exit 1
+
+if upgradeTest then
+  docker pull quay.io/metal3-io/cluster-api-provider-metal3:release-0.5
+  docker image tag quay.io/metal3-io/cluster-api-provider-metal3:release-0.5 192.168.111.1:5000/localimages/cluster-api-provider-metal3:release-0.5
+  docker pull quay.io/metal3-io/ip-address-manager:release-0.1
+  docker image tag quay.io/metal3-io/ip-address-manager:release-0.1 192.168.111.1:5000/localimages/ip-address-manager:release-0.1
+fi
```
Rozzii, line 103: shouldn't this happen before the `${M3_DEV_ENV_PATH}/scripts/run_command.sh make e2e-tests` command?

test/e2e/config/e2e_conf.yaml
```diff
@@ -87,6 +128,7 @@ variables:
   CAPM3RELEASE: "${CAPM3RELEASE}"
   CONFIG_FILE_PATH: "${HOME}/.cluster-api/overrides/clusterctl.yaml"
 
+
```
Rozzii, line 131: This newline is not needed.

test/e2e/config/e2e_conf.yaml
```diff
@@ -95,7 +137,7 @@ intervals:
   default/wait-delete-cluster: ["20m", "10s"]
   default/wait-machine-upgrade: ["50m", "10s"]
   default/wait-machine-remediation: ["30m", "10s"]
-  default/wait-vm-state: ["20m", "100ms"]
+  default/wait-vm-state: ["10m", "100ms"]
```
Rozzii, line 140: Is this necessary if yes then why ? In case it  is this just a leftover from debugging please revert the change?

test/e2e/data/shared/v1alpha4_provider/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
+releaseSeries:
+  - major: 0
+    minor: 4
+    contract: v1alpha3
+  - major: 0
+    minor: 5
+    contract: v1alpha4
```
Rozzii, line 8: new line at the EOF is missing

test/e2e/data/shared/v1beta1_provider/metadata.yaml
```diff
@@ -0,0 +1,11 @@
...
+releaseSeries:
+  - major: 0
+    minor: 4
+    contract: v1alpha3
+  - major: 0
+    minor: 5
+    contract: v1alpha4
+  - major: 1
+    minor: 0
+    contract: v1beta1
```
Rozzii, line 11: new line at the EOF is missing

test/e2e/e2e_test.go
```diff
@@ -43,7 +44,9 @@ var _ = Describe("Workload cluster creation", func() {
 	})
 
 	AfterEach(func() {
-		dumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, artifactFolder, namespace, cluster, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
+		if false {
+			dumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, artifactFolder, namespace, cluster, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
```
Rozzii, line 48: Actually this could be controlled by some environment variable, at the moment we don't even really use this, maybe in the future but currently we have a single "It" function so this runs only after everything has already finished but then in CI we would delete the VMs anyways so it would be nice if this would be configurable.

Anyways I think you should just remove the "if" after this goes from draft to regular PR and later we will add the configuration support here.

test/e2e/upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+					break
+				} 
+			}
+			if !exist {
+				images = append(images, c.Image)
+				Logf("%v", c.Image)
+			}
+		}
+	}
+}
```
Rozzii, line 166: missing newline at EOF

test/e2e/e2e_test.go
```diff
@@ -73,12 +76,15 @@ var _ = Describe("Workload cluster creation", func() {
 			}, result)
 			cluster = result.Cluster
 			targetCluster = bootstrapClusterProxy.GetWorkloadCluster(ctx, namespace, clusterName)
-			remediation()
```
Rozzii, line 76: Don't forget to put remediation() back to the code when the PR goes from draft to normal state.

test/e2e/upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+
+func upgrade() {
+	var (
+		specName = "clusterctl-upgrade"
+
+		//	testNamespace *corev1.Namespace
+
+		managementClusterProxy framework.ClusterProxy
+
+	//	workLoadClusterName string
```
Rozzii, line 29: If the code is not needed please remove it don't leave commented code snippets in the code.

Rozzii (commented):
Just a few questions and a minor things.

test/e2e/upgrade.go
```diff
@@ -0,0 +1,166 @@
+package e2e
```
Rozzii, line 1: I think a better name for this file is upgrade_management_cluster.go

test/e2e/upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/config"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func upgrade() {
```
Rozzii, line 21: I think the function name should also reflect the contained logic so this should be upgradeManagementCluster

test/e2e/upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+
+	By("Restore original BMO configmap")
+	restoreBMOConfigmap()
+	By("Reinstate Ironic Configmap")
+	configureIronicConfigmap(false)
+
+	By("Add labels to BMO CRDs in the target cluster")
+	labelBMOCRDs(targetCluster)
+
+	By("Ensure API servers are stable before doing move")
```
Rozzii, line 98: "...before doing the upgrade."

test/e2e/upgrade.go
```diff
@@ -0,0 +1,166 @@
...
+	}, "1s", "100ms").Should(BeNil(), "Failed to assert target API server stability")
+
+	By("Get the management cluster images before upgrading")
+	pods, err := targetClusterClientSet.CoreV1().Pods("").List(ctx, metav1.ListOptions{})
+	printImages(pods)
+
+	By("Upgrading providers to the latest version available")
+	Logf("clusterv1.GroupVersion.Version: %v", clusterv1.GroupVersion.Version)
+	Logf("os.Getenv: %v", os.Getenv("CONFIG_FILE_PATH"))
+	CONFIG_FILE_PATH := os.Getenv("HOME")+"/.cluster-api/clusterctl.yaml"
```
Rozzii, line 115: If you stay with this method of configuring the config path then I think this should look something like:
`configFilePath := os.Getenv("HOME")+"/.cluster-api/clusterctl.yaml"
Logf("os.Getenv: %v", configFilePath)`

test/e2e/upgrade_management_cluster.go
```diff
@@ -0,0 +1,108 @@
...
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+)
+
+func upgradeManagementCluster() {
+	var (
+		specName = "clusterctl-upgrade"
+		managementClusterProxy framework.ClusterProxy
+	)
+	managementClusterProxy = bootstrapClusterProxy.GetWorkloadCluster(ctx, cluster.Namespace, cluster.Name)
```
lentzi90, line 25: I'm a bit confused about this. This is how I think it is going:
1. The bootstrap cluster is created and initialized with the latest CAPM3 + BMO and Ironic
2. A workload cluster is created (completely empty)
3. In this upgrade test we say that the workload cluster is now a management cluster and do `clusterctl init` with an old version (but we never install BMO or Ironic here)
4. We upgrade CAPI and CAPM3 in this management/workload cluster and if it succeeds the test is passed.

Did I understand this correctly?

If so, I think we should rename this second management cluster so it is less confusing. The bootstrap cluster is also a management cluster so it is quite easy to confuse them. Maybe "upgradeCluster" or "testCluster" or something like this?

Second, I'm a bit worried that we are not installing Ironic and BMO and checking that things are working after the upgrade. Maybe we can add these later?

lentzi90 (commented):
This is looking quite good now I think.

Added some notes below.

test/e2e/config/e2e_conf.yaml
```diff
@@ -15,66 +15,107 @@ images:
...
+        files:
+          - sourcePath: "../data/shared/v1beta1/metadata.yaml"
+        replacements:
+          - old: "--leader-elect"
+            new: "--leader-elect=false"
+  - name: metal3
+    type: InfrastructureProvider
+    versions:
+    - name: v0.5.3
+      value: "https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/v0.5.3/infrastructure-components.yaml"
```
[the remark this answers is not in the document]
mboukhalfa: For now using the release manifests to skip pushing the images to the local registry

test/e2e/upgrade_management_cluster.go
mboukhalfa: Yes, it is correct and I agree and consider changing the cluster name. Also checking the cluster is working correctly would be in a follow up.

mboukhalfa:
test-v1b1-e2e

templates/test/cluster-template-prow-ha-upgrade.yaml
```diff
@@ -0,0 +1,395 @@
+apiVersion: cluster.x-k8s.io/${UPGRADE_FROM_CAPI_VERSION}
```
lentzi90, line 1: This template is used by the bootstrap ("management") cluster which is using the latest version, right? So I would suggest changing it so it uses the `CAPI_VERSION` instead and the same for `CAPM3_VERSION`.

test/e2e/upgrade_management_cluster.go
```diff
@@ -0,0 +1,137 @@
...
+	Logf("Downloading clusterctl binary from %s", clusterctlBinaryURL)
+	clusterctlBinaryPath := downloadToTmpFile(clusterctlBinaryURL)
+	defer os.Remove(clusterctlBinaryPath) // clean up
+
+	err := os.Chmod(clusterctlBinaryPath, 0744) //nolint:gosec
+	Expect(err).ToNot(HaveOccurred(), "failed to chmod temporary file")
+
+	By("Initializing the workload cluster with older versions of providers")
+
+	contract := "v1alpha4"
```
lentzi90, line 44: This should probably come from the config file or environment. Maybe use `UPGRADE_FROM_CAPI_VERSION`?

lentzi90 (commented):
Added some more comments.

I would suggest renaming/moving the `v1*_provider/metadata.yaml` files. Now they are under `shared`, but they are specific for CAPM3, not shared with the other. The "shared" are for all the CAPI providers (bootstrap, control-plane and core). Something like this would be better I think:

```
├── capm3
│   └── v1beta1
│       └── metadata.yaml
└── shared
    └── v1beta1
        └── metadata.yaml
```

Also note that there is no need to have a separate v1alpha4 metadata.yaml if all versions are anyway included in v1beta1 metadata.yaml. Then we can just use v1beta1 metadata for everything.

templates/test/cluster-template-prow-ha-upgrade.yaml
mboukhalfa: Here `UPGRADE_FROM_CAPI_VERSION` is v1alpha4 while `CAPI_VERSION` is v1beta1

lentzi90: Yes I understand. I'm saying that this template is applied in the initial management cluster, which is running v1beta1. In the cluster that is created we then do `clusterctl init` with v1alpha4, but we never apply this template in there.

mboukhalfa: In fact this template is not used since the v1a5 management cluster does not create any cluster but it should use that template in that case (this will be introduced later as a test workload cluster) v1b1 uses the devenv template

lentzi90: Oh I see. In that case I think it would be better to remove it from this PR. We can add it when we actually make use of it.

lentzi90 (commented):
I think the cluster template could be removed since we do not use it. Otherwise I think this looks good.

lentzi90 (approved):

furkatgofurov7:
This looks behind the current main branch HEAD, please rebase the PR (noticed due to old CAPI v1-beta1 mentions)

Rozzii (approved):
lgtm

test/e2e/config/e2e_conf.yaml
```diff
@@ -18,6 +18,15 @@ providers:
...
   versions:
+  - name: v0.4.7
+    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/core-components.yaml"
+    type: "url"
+    contract: v1alpha4
+    replacements:
+      - old: --metrics-addr=127.0.0.1:8080
+        new: --metrics-addr=:8080
+    files:
+      - sourcePath: "../data/shared/v1beta1/metadata.yaml"
```
furkatgofurov7, line 29: I am not sure I like the idea of having single metadata regardless of an API type, v1beta1 metadata.yaml for the release that is from v1a4 API makes it confusing

test/e2e/config/e2e_conf.yaml
```diff
@@ -31,6 +40,15 @@ providers:
...
   versions:
+  - name: v0.4.7 # latest published release in the v1alpha4 series; this is used for v1alpha4 --> v1beta1 clusterctl upgrades test only.
+    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/bootstrap-components.yaml"
+    type: "url"
+    contract: v1alpha4
+    replacements:
+      - old: --metrics-addr=127.0.0.1:8080
+        new: --metrics-addr=:8080
+    files:
+      - sourcePath: "../data/shared/v1beta1/metadata.yaml"
```
furkatgofurov7, line 51: Same question/comment as above

test/e2e/config/e2e_conf.yaml
```diff
@@ -44,6 +62,15 @@ providers:
...
   versions:
+  - name: v0.4.7 # latest published release in the v1alpha4 series; this is used for v1alpha4 --> v1beta1 clusterctl upgrades test only.
+    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/control-plane-components.yaml"
+    type: "url"
+    contract: v1alpha4
+    replacements:
+      - old: --metrics-addr=127.0.0.1:8080
+        new: --metrics-addr=:8080
+    files:
+      - sourcePath: "../data/shared/v1beta1/metadata.yaml"
```
furkatgofurov7, line 73: Same question/comment as above

test/e2e/config/e2e_conf.yaml
```diff
@@ -57,6 +84,13 @@ providers:
 - name: metal3
   type: InfrastructureProvider
   versions:
+  - name: v0.5.4
+    value: "https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/v0.5.4/infrastructure-components.yaml"
+    type: "url"
+    contract: v1alpha4
+    files:
+    - sourcePath: "../data/capm3/v1beta1_provider/metadata.yaml"
```
furkatgofurov7, line 92: question/comment on the same direction as above

test/e2e/config/e2e_conf.yaml
```diff
@@ -72,9 +106,9 @@ providers:
     #   targetName: "cluster-template-ha-centos.yaml"
     - sourcePath: "${PWD}/test/e2e/_out/cluster-template-prow-ha-m3-dev-env.yaml"
       targetName: "cluster-template-ha-m3-dev-env.yaml"
-
 variables:
   KUBERNETES_VERSION: "${KUBERNETES_VERSION}"
+  INIT_WITH_KUBERNETES_VERSION: "${KUBERNETES_VERSION}"
```
furkatgofurov7, line 111: I don't find this being used anywhere?

test/e2e/e2e_suite_test.go
```diff
@@ -71,6 +74,7 @@ func init() {
 	flag.StringVar(&configPath, "e2e.config", "", "path to the e2e config file")
 	flag.StringVar(&artifactFolder, "e2e.artifacts-folder", "", "folder where e2e test artifact should be stored")
 	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+	flag.BoolVar(&upgradeTest, "e2e.trigger-upgrade-test", false, "if true, the e2e upgrade test will be triggered if true")
```
furkatgofurov7, line 77:
```suggestion
	flag.BoolVar(&upgradeTest, "e2e.trigger-upgrade-test", false, "if true, the e2e upgrade test will be triggered")
```

test/e2e/upgrade_management_cluster.go
```diff
@@ -0,0 +1,137 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/config"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+)
+
+func upgradeManagementCluster() {
```
furkatgofurov7, line 24: Adding starting log for tests as we do for other e2e tests would be nice

test/e2e/upgrade_management_cluster.go
```diff
@@ -0,0 +1,137 @@
...
+	}, e2eConfig.GetIntervals(specName, "wait-controllers")...)
+
+	By("THE MANAGEMENT CLUSTER WAS SUCCESSFULLY UPGRADED!")
+
+	By("Get the management cluster images after upgrading")
+	pods, err = targetClusterClientSet.CoreV1().Pods("").List(ctx, metav1.ListOptions{})
+	Expect(err).To(BeNil())
+	printImages(pods)
+
+	By("UPGRAD MANAGEMENT CLUSTER PASSED!")
```
furkatgofurov7, line 101: UPGRAD**E**

furkatgofurov7 (commented):
Nice work! I would also like to know your opinion on, how do you feel about merging/testing all upgrade-related e2e tests (this + BMO + ironic containers) we currently have together in e2e upgrade tests?

Few comments inline:

test/e2e/config/e2e_conf.yaml
lentzi90: Would it be ok if we just skip the `v1beta1` in the path? So we would have `data/shared/metadata.yaml`.
I mean it just lists all the releases and what contract to use, right? So why would we need to have them separate for v1alpha4 and v1beta1?

furkatgofurov7: @lentzi90 actually, why should that file be the same and reused for CAPI and CAPM3? 
CAPIs' release series was always mapped 1:1, i.e:

```
releaseSeries:
  - major: 1
    minor: 1
    contract: v1beta1
  - major: 1
    minor: 0
    contract: v1beta1
  - major: 0
    minor: 4
    contract: v1alpha4
  - major: 0
    minor: 3
    contract: v1alpha3
```

whereas CAPM3 had 2 minor releases for the same contract, i.e:


```
releaseSeries:
- major: 1
  minor: 0
  contract: v1beta1
- major: 0
  minor: 5
  contract: v1alpha4
- major: 0
  minor: 4
  contract: v1alpha3
- major: 0
  minor: 3
  contract: v1alpha3
```

So, it means in this PR, we have this metadata:

```
apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
kind: Metadata
releaseSeries:
  - major: 0
    minor: 4
    contract: v1alpha3
  - major: 0
    minor: 5
    contract: v1alpha4
  - major: 1
    minor: 0
    contract: v1beta1
```

being used for CAPI release v0.4.7. 

Moreover, as releases keep coming up, in my opinion, it is not convenient to use the same metadata.yaml for each e2e test and that is the practice I have seen in other CAPI provider repos, for ref. you can check here:

- https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/main/test/e2e/data/shared
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/test/e2e/data/shared

test/e2e/upgrade_management_cluster.go
```diff
@@ -0,0 +1,138 @@
+package e2e
```
furkatgofurov7, line 1: Also, this file has to be renamed to `upgrade_management_cluster_test.go` to avoid **undeclared_name** warnings that we recently fixed in https://github.com/metal3-io/cluster-api-provider-metal3/commit/9cb57bd2b19f928032a51ff990fa5ada679de882

furkatgofurov7 (changes requested):

test/e2e/config/e2e_conf.yaml
mboukhalfa: I have followed the same pattern in this PR I still think moving to a single metadata (if agreed) can be done independently of the upgrade changes

mboukhalfa:
test-v1b1-integration

test/e2e/config/e2e_conf.yaml
lentzi90: I didn't mean we would use the same for CAPI and CAPM3, I mean I don't understand why we need to have one for v1beta1 that has all the older versions in it and then also have one for v1alpha5 and v1alpha4? But I see the other providers do this so I guess there must be a good reason.

Rozzii (commented):
This is not my full review, I will continue .

scripts/ci-e2e.sh
```diff
@@ -53,7 +53,8 @@ clone_repos
 REPO_ROOT=$(realpath "$REPO_ROOT") 
 # Copy the current CAPM3 repo to the Go source directory
 rm -rf "${M3PATH}/cluster-api-provider-metal3" # To avoid 'permission denied' error when overriding .git/
-cp -R "${REPO_ROOT}" "${M3PATH}/cluster-api-provider-metal3/" 
+# TODO (Mohammed) Remove the next line since it overrides the cloned v1a5
```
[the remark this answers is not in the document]
mboukhalfa: I kept this here since I am still not sure why we needed this at the first place I will remove this once squash the commit but I think the comments just bring attention to this changes and make it easy to review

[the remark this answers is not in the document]
mboukhalfa: cc @namnx228

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		ctx                 = context.TODO()
+		specName            = "metal3"
+		namespace           = "metal3"
+		clusterName         = "test1"
+		workloadClusterName = "test2"
+		clusterctlLogFolder string
+		kubernetesVersion   = e2eConfig.GetVariable(KubernetesVersion)
+	)
+	/*---------------------------------------------*
+	| Create a target cluster with v1a4 clusterctl |
```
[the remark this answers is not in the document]
mboukhalfa: Yes, the var [INIT_WITH_BINARY](https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/clusterctl-{OS}-{ARCH}") is pointing to clusterctl 0.4.7 that would be used to create the cluster

scripts/ci-e2e.sh
```diff
@@ -53,7 +53,8 @@ clone_repos
 REPO_ROOT=$(realpath "$REPO_ROOT") 
 # Copy the current CAPM3 repo to the Go source directory
 rm -rf "${M3PATH}/cluster-api-provider-metal3" # To avoid 'permission denied' error when overriding .git/
-cp -R "${REPO_ROOT}" "${M3PATH}/cluster-api-provider-metal3/" 
+# TODO (Mohammed) Remove the next line since it overrides the cloned v1a5 
+# cp -R "${REPO_ROOT}" "${M3PATH}/cluster-api-provider-metal3/"
```
lentzi90, line 57: I'm afraid this will make us test the wrong code for non-upgrade tests but I'm not sure. For the other e2e tests (not upgrade) I think we checkout the code for the PR first, and we copy it here to the correct path because otherwise it would just get the main branch instead. So I think we must change this so the copy happens by default, but if it is an upgrade test we skip it.

test/e2e/e2e_suite_test.go
```diff
@@ -71,6 +74,7 @@ func init() {
 	flag.StringVar(&configPath, "e2e.config", "", "path to the e2e config file")
 	flag.StringVar(&artifactFolder, "e2e.artifacts-folder", "", "folder where e2e test artifact should be stored")
 	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+	flag.BoolVar(&upgradeTest, "e2e.trigger-upgrade-test", false, "if true, the e2e upgrade test will be triggered")
```
lentzi90, line 77: I think it would be good to add here that it will run *only* the upgrade tests and skip the others.
```suggestion
	flag.BoolVar(&upgradeTest, "e2e.trigger-upgrade-test", false, "if true, the e2e upgrade test will be triggered and other tests will be skipped")
```

test/e2e/config/e2e_conf.yaml
```diff
@@ -57,9 +84,19 @@ providers:
...
+    contract: v1alpha4
+    files:
+    - sourcePath: "../data/capm3/v1alpha5/metadata.yaml"
+      targetName: "metadata.yaml"
+    - sourcePath: "${PWD}/test/e2e/_out/cluster-template-prow-ha.yaml"
+      targetName: "cluster-template.yaml"
+    - sourcePath: "${PWD}/test/e2e/_out/cluster-template-prow-ha-upgrade-workload.yaml"
+      targetName: "cluster-template-ha-upgrade-workload.yaml"
+  - name: v1.1.0
+    value: "${PWD}/config/default"
```
lentzi90, line 99: Where is this coming from? Is it generated from metal3-dev-env?

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+	"runtime"
+	"strings"
+
+	bmo "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	clusterv1old "sigs.k8s.io/cluster-api/api/v1alpha3"
```
lentzi90, line 23: I think it would be better to spell out the version here.
```suggestion
	clusterv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
```
Or maybe
```suggestion
	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
```

Now that I'm thinking about it, why is it v1alpha3? Are we not going from v1alpha4 -> v1beta1?

lentzi90 (commented):
I'm not sure what to think about the "extra" workload cluster for checking that the old cluster works. It is quite nice to have a thorough test, but I think it may be too complicated and slow. I'll sleep on it and take another look tomorrow.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+			for _, machine := range machineList.Items {
+				if strings.EqualFold(machine.Status.Phase, "running") {
+					n++
+				}
+			}
+		}
+		return n, nil
+	}, e2eConfig.GetIntervals(specName, "wait-worker-nodes")...).Should(Equal(*controlPlaneMachineCount))
+
+	By("THE MANAGEMENT CLUSTER WITH OLDER VERSION OF PROVIDERS WORKS!")
```
lentzi90, line 255: After sleeping on this I think we should remove this part. We don't do this in the ansible upgrade test and I would prefer to keep it the same here to start with. We also already test that the old version can create a cluster in main integration tests.

I think there was a bit of confusion about this maybe because I didn't explain it clearly before. Earlier we didn't do pivoting in this PR and we also didn't have the "second workload cluster" so there was not really any test to show that the upgraded components actually worked. Now that we have pivoting in here that will be enough to make sure that things are working. So great job adding that pivot! :tada:

What I'm trying to say is basically that we don't need both pivot and a second cluster, it is enough to "just" have pivot.

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: done

mboukhalfa: PS: Changing that to capiv1alpha4 caused the error `no kind is registered for the type v1alpha4.MachineList in scheme "pkg/runtime/scheme.go:100"` so I added it in suite file to fix that !  please take a look :)

lentzi90: Good! I'm not sure if there are other implications from this `AddToScheme`. It could be that we should only us it for upgrade tests, but I'm not sure about it. Maybe someone else knows?

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		kubeSystem := &corev1.Namespace{}
+		return bootstrapClusterProxy.GetClient().Get(ctx, client.ObjectKey{Name: "kube-system"}, kubeSystem)
+	}, "5s", "100ms").Should(BeNil(), "Failed to assert bootstrap API server stability")
+	Consistently(func() error {
+		kubeSystem := &corev1.Namespace{}
+		return upgradeClusterClient.Get(ctx, client.ObjectKey{Name: "kube-system"}, kubeSystem)
+	}, "5s", "100ms").Should(BeNil(), "Failed to assert target API server stability")
+	By("Moving the cluster to self hosted")
+
+	cmd := exec.Command(clusterctlBinaryPath, "move", "--to-kubeconfig", upgradeClusterProxy.GetKubeconfigPath(), "-n", namespace, "-v", "10")
```
namnx228, line 170: Could you please remind me why do we have to use the clusterctl binary instead of using the test framework API?

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		for _, bmh := range bmhList.Items {
+			if !bmh.WasProvisioned() {
+				return errors.New("BMHs cannot be provisioned")
+			}
+		}
+		return nil
+	}, e2eConfig.GetIntervals(specName, "wait-object-provisioned")...).Should(BeNil())
+
+	Logf("Scale down KCP to 1 in order to 2 VMs for the workload cluster")
+	cmd = exec.Command("kubectl", "scale", "kcp", clusterName, "-n", namespace, "--replicas", "1", "--kubeconfig", upgradeClusterProxy.GetKubeconfigPath())
```
namnx228, line 190: Why should we not use the framework API here? It should be preference to use the native functions instead of using exec.Command

namnx228 (commented):
Look good to me. Left some comments and questions

templates/test/upgrade/clusterctl-upgrade-test.yaml
```diff
@@ -0,0 +1,5 @@
+APIVersion: infrastructure.cluster.x-k8s.io/v1beta1
+providers:
+- name: metal3
+  type: InfrastructureProvider
+  url: {{ .ARTIFACTS }}/repository/infrastructure-metal3/{{ .CAPM3_REL_TO_VERSION }}/components.yaml
```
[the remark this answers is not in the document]
mboukhalfa: path to the _artifacts folder

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: Since we are doing the move between v1alpha4 clusters cannot use the framework since it will use v1beta1 clusterctl

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: for now we are using scaling since we cannot specify replicas from the e2e once we start using template parameterized from the e2e we can directly specify 1 kcp instead of starting with 3 then scale in to 1.
if #537 merged first we can rebase this PR else the template PR will consider using 1kcp and remove the scale part so I would prefer doing the change after merging one of these PRs.

test/e2e/config/e2e_conf.yaml
```diff
@@ -18,6 +18,15 @@ providers:
 - name: cluster-api
   type: CoreProvider
   versions:
+  - name: v0.4.7
```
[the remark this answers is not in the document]
mboukhalfa: I do not think it is recommended to use a var for that for two reasons:
- It may lead to a conflict between the var value and what we expect in this file
- The name does not necessarily need to be the exact version value

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		ClusterctlConfigPath:    clusterctlConfigPath,
+		CoreProvider:            e2eConfig.GetProviderLatestVersionsByContract(contract, config.ClusterAPIProviderName)[0],
+		BootstrapProviders:      e2eConfig.GetProviderLatestVersionsByContract(contract, config.KubeadmBootstrapProviderName),
+		ControlPlaneProviders:   e2eConfig.GetProviderLatestVersionsByContract(contract, config.KubeadmControlPlaneProviderName),
+		InfrastructureProviders: e2eConfig.GetProviderLatestVersionsByContract(contract, e2eConfig.InfrastructureProviders()...),
+		LogFolder:               filepath.Join(artifactFolder, "clusters", clusterName),
+	}, e2eConfig.GetIntervals(specName, "wait-controllers")...)
+
+	/*-------------------------------------------------*
+	| Pivote to run ironic/BMO and resources on target |
```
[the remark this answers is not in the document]
mboukhalfa: done thanks

namnx228 (commented):
LGTM

test/e2e/config/e2e_conf.yaml
```diff
@@ -57,9 +84,19 @@ providers:
 - name: metal3
   type: InfrastructureProvider
   versions:
-  - name: v1.1.0
-    value: "file://${HOME}/.cluster-api/overrides/infrastructure-metal3/v1.1.0/infrastructure-components.yaml"
+  - name: v0.5.4
+    value: "https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/v0.5.4/infrastructure-components.yaml"
```
furkatgofurov7, line 88: These should be `v0.5.5` now

hack/e2e/environment.sh
```diff
@@ -21,6 +21,22 @@ function os_check() {
...
+export UPGRADE_TEST=${UPGRADE_TEST:-false}
+if $UPGRADE_TEST; then
+    export CAPI_VERSION="v1alpha4"
+    export CAPM3_VERSION="v1alpha5"
+fi
+
+# Override project infra vars that point to 
+# the current branch to build capm3 image and crds
+if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then 
+    export CAPM3RELEASE="v0.5.4"
```
furkatgofurov7, line 33: `v0.5,5` now

hack/e2e/environment.sh
```diff
@@ -21,6 +21,22 @@ function os_check() {
...
+    export CAPIRELEASE="v0.4.7"
+    export CAPM3BRANCH="release-0.5"
+    unset CAPM3_LOCAL_IMAGE
```
furkatgofurov7, line 36: What is the reason to unset this?

hack/e2e/environment.sh
```diff
@@ -21,6 +21,22 @@ function os_check() {
...
+    export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
+    export CAPM3PATH="${M3PATH}/cluster-api-provider-metal3"
+fi
```
furkatgofurov7, line 39: If all these are only upgrade related overriding vars, let's move them to the condition on line 25?

test/e2e/config/e2e_conf.yaml
```diff
@@ -72,7 +109,7 @@ providers:
     #   targetName: "cluster-template-ha-centos.yaml"
     - sourcePath: "${PWD}/test/e2e/_out/cluster-template-prow-ha-m3-dev-env.yaml"
       targetName: "cluster-template-ha-m3-dev-env.yaml"
-
+
```
furkatgofurov7, line 112: nit: extra tab space

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+	resp, err := http.Get(url) //nolint:gosec
+	Expect(err).ToNot(HaveOccurred(), "failed to get clusterctl")
+	defer resp.Body.Close()
+
+	// Write the body to file
+	_, err = io.Copy(tmpFile, resp.Body)
+	Expect(err).ToNot(HaveOccurred(), "failed to write temporary file")
+
+	return tmpFile.Name()
+}
```
furkatgofurov7, line 332: We could reuse the already existing `downloadFile` [func](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/common.go#L70-L89) from the common.go

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+func printImages(clusterProxy framework.ClusterProxy) {
```
furkatgofurov7, line 333: Missing new line before func defining

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		WorkerMachineCount:       pointer.Int64Ptr(1),
+		InfrastructureProvider:   clusterctl.DefaultInfrastructureProvider,
+		// setup clusterctl logs folder
+		LogFolder: clusterctlLogFolder,
+	})
+
+	Expect(upgradeClusterTemplate).ToNot(BeNil(), "Failed to get the cluster template")
+
+	Logf("Applying the cluster template yaml to the cluster:")
+	Logf(string(upgradeClusterTemplate))
```
furkatgofurov7, line 77:
```suggestion
	Logf("Applying the cluster template yaml to the cluster: %v\n", string(upgradeClusterTemplate))
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+	output, err = cmd.CombinedOutput()
+	Logf("scale: %v", string(output))
+	Expect(err).To(BeNil())
+
+	Logf("Waiting for 2 BMHs to be in Available state")
+	Eventually(func(g Gomega) {
+		bmhs, err := getAllBmhs(ctx, upgradeClusterClient, namespace, specName)
+		g.Expect(err).To(BeNil())
+		g.Expect(filterBmhsByProvisioningState(bmhs, bmo.StateAvailable)).To(HaveLen(2))
+	}, e2eConfig.GetIntervals(specName, "wait-machine-remediation")...).Should(Succeed())
```
furkatgofurov7, line 200:
```suggestion
	}, e2eConfig.GetIntervals(specName, "wait-bmh-available")...).Should(Succeed())
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+	| Create a test workload cluster |
+	*--------------------------------*/
+	var (
+		controlPlaneMachineCount = pointer.Int64Ptr(1)
+		workerMachineCount       = pointer.Int64Ptr(1)
+	)
+	By("Creating a test workload cluster")
+
+	Logf("Creating the workload cluster with name %q using the %q template (Kubernetes %s, %d control-plane machines, %d worker machines)",
+		workloadClusterName, "(default)", kubernetesVersion, *controlPlaneMachineCount, *workerMachineCount)
```
furkatgofurov7, line 212: Why not to directly use the 'default' in the log?
```suggestion
	Logf("Creating the workload cluster with name %q using the default template (Kubernetes %s, %d control-plane machines, %d worker machines)",
		workloadClusterName, kubernetesVersion, *controlPlaneMachineCount, *workerMachineCount)
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+	Logf("Getting the cluster template yaml")
+	Logf("clusterctlBinaryPath = %v", clusterctlBinaryPath)
+	Logf("ClusterctlConfigPath = %v", clusterctlConfigPath)
+	workloadClusterTemplate := clusterctl.ConfigClusterWithBinary(ctx, clusterctlBinaryPath, clusterctl.ConfigClusterInput{
+		// pass reference to the management cluster hosting this test
+		KubeconfigPath: upgradeClusterProxy.GetKubeconfigPath(),
+		// pass the clusterctl config file that points to the local provider repository created for this test,
+
```
furkatgofurov7, line 221: nit: extra new line

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+	Logf("Creating the workload cluster with name %q using the %q template (Kubernetes %s, %d control-plane machines, %d worker machines)",
+		workloadClusterName, "(default)", kubernetesVersion, *controlPlaneMachineCount, *workerMachineCount)
+
+	Logf("Getting the cluster template yaml")
+	Logf("clusterctlBinaryPath = %v", clusterctlBinaryPath)
+	Logf("ClusterctlConfigPath = %v", clusterctlConfigPath)
+	workloadClusterTemplate := clusterctl.ConfigClusterWithBinary(ctx, clusterctlBinaryPath, clusterctl.ConfigClusterInput{
+		// pass reference to the management cluster hosting this test
+		KubeconfigPath: upgradeClusterProxy.GetKubeconfigPath(),
+		// pass the clusterctl config file that points to the local provider repository created for this test,
```
furkatgofurov7, line 220:
```suggestion
		// pass the clusterctl config file that points to the local provider repository created for this test
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		machineList := &clusterv1alpha4.MachineList{}
+		if err := upgradeClusterClient.List(ctx, machineList, client.InNamespace(namespace), client.MatchingLabels{clusterv1.ClusterLabelName: workloadClusterName}); err == nil {
+			for _, machine := range machineList.Items {
+				if strings.EqualFold(machine.Status.Phase, "running") {
+					n++
+				}
+			}
+		}
+		return n, nil
+	}, e2eConfig.GetIntervals(specName, "wait-worker-nodes")...).Should(Equal(*controlPlaneMachineCount))
```
furkatgofurov7, line 253:
```suggestion
	}, e2eConfig.GetIntervals(specName, "wait-machine-running")...).Should(Equal(*controlPlaneMachineCount))
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+		if err := upgradeClusterClient.List(ctx, bmhList, client.InNamespace(namespace)); err != nil {
+			return err
+		}
+		for _, bmh := range bmhList.Items {
+			if !bmh.WasProvisioned() {
+				return errors.New("BMHs cannot be provisioned")
+			}
+		}
+		return nil
+	}, e2eConfig.GetIntervals(specName, "wait-object-provisioned")...).Should(BeNil())
```
furkatgofurov7, line 187:
```suggestion
	}, e2eConfig.GetIntervals(specName, "wait-bmh-provisioned")...).Should(BeNil())
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+	Logf("Upgrade management cluster to : %v", clusterv1.GroupVersion.Version)
+	clusterctl.UpgradeManagementClusterAndWait(ctx, clusterctl.UpgradeManagementClusterAndWaitInput{
+		ClusterctlConfigPath: e2eConfig.GetVariable("CONFIG_FILE_PATH"),
+		ClusterProxy:         upgradeClusterProxy,
+		Contract:             clusterv1.GroupVersion.Version,
+		LogFolder:            filepath.Join(artifactFolder, "clusters", workloadClusterName),
+	}, e2eConfig.GetIntervals(specName, "wait-controllers")...)
+
+	By("THE MANAGEMENT CLUSTER WAS SUCCESSFULLY UPGRADED!")
```
furkatgofurov7, line 295: This log is confusing since you have the same log on line 314

furkatgofurov7 (changes requested):

furkatgofurov7:
:warning:  **Also please link this PR to issue 319 and update it with "Fixes: #" in the PR description.**  Thanks!

hack/e2e/environment.sh
mboukhalfa: This var is set in[ project infra](https://github.com/metal3-io/project-infra/blob/9470543342be88a38a94e120b81ee04c90a9576b/jenkins/scripts/integration_test_env.sh#L63)  to direct the capm3 location to tested_pr while we want in the upgrade to start with an old version cloned under go/src/..

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -0,0 +1,336 @@
...
+	Logf("Waiting for 2 BMHs to be in Available state")
+	Eventually(func(g Gomega) {
+		bmhs, err := getAllBmhs(ctx, upgradeClusterClient, namespace, specName)
+		g.Expect(err).To(BeNil())
+		g.Expect(filterBmhsByProvisioningState(bmhs, bmo.StateAvailable)).To(HaveLen(2))
+	}, e2eConfig.GetIntervals(specName, "wait-bmh-available")...).Should(Succeed())
+
+	/*-------------------------------*
+	| Create a test workload cluster |
+	*--------------------------------*/
```
lentzi90, line 202: I still think this part should be removed. Before pivoting was added to this upgrade test I said we needed it, but now that we have pivoting I don't think there is a point to keep it. It makes the e2e upgrade take longer time and makes it different from the ansible-based upgrade test.

The point is that we already test the old version by creating the first workload cluster and pivot to it. Then we do the upgrade with all controllers running in the (first) target cluster so that is also quite well tested.

mboukhalfa: So this part might be an extra but since we are keeping the test steps close to CAPI I prefer to keep it since we are still working on reusing that test from CAPI

test/e2e/upgrade_management_cluster_test.go
furkatgofurov7: @lentzi90 I do not see it being set to `sigs.k8s.io/cluster-api/api/v1alpha3` anymore

mboukhalfa:
/test-v1b1-ubuntu-integration
gzip issue when extracting kustomize tar

mboukhalfa:
test-v1b1-ubuntu-integration

test/e2e/config/e2e_conf.yaml
```diff
@@ -57,12 +84,22 @@ providers:
...
   type: InfrastructureProvider
   versions:
-  - name: v1.1.0
-    value: "file://${HOME}/.cluster-api/overrides/infrastructure-metal3/v1.1.0/infrastructure-components.yaml"
+  - name: v0.5.5
+    value: "https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/v0.5.5/infrastructure-components.yaml"
     type: "url"
+    contract: v1alpha4
+    files:
+    - sourcePath: "../data/infrastructure-metal3/metadata.yaml"
```
mboukhalfa, line 92: Please note that we are using same metadata for both CAPM3 v1alpha5 and v1beta1 
@lentzi90 @furkatgofurov7

test/e2e/upgrade_management_cluster_test.go
lentzi90: After discussion we decided to merge as is. :heavy_check_mark:

Makefile
```diff
@@ -148,6 +149,7 @@ E2E_TEMPLATES_DIR ?= $(ROOT_DIR)/test/e2e/data/infrastructure-metal3
 cluster-templates: $(KUSTOMIZE) ## Generate cluster templates
 	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu > $(E2E_OUT_DIR)/cluster-template-ubuntu.yaml
 	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-centos > $(E2E_OUT_DIR)/cluster-template-centos.yaml
+	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-upgrade-workload --load-restrictor LoadRestrictionsNone > $(E2E_OUT_DIR)/cluster-template-upgrade-workload.yaml
```
lentzi90, line 152: I really don't like that we need `--load-restrictor` and I have a suggestion for how to get rid of it. It is needed only when trying to reference files outside of the working directory for patches, so we can move things around a bit to fix it.

```
.
└── infrastructure-metal3
    ├── bases
    │   ├── centos-kubeadm-config    # Put the kubeadm patches here but do NOT include the IP pool.
    │   ├── cluster
    │   ├── ippool
    │   └── ubuntu-kubeadm-config   # Put the kubeadm patches here but do NOT include the IP pool.
    ├── cluster-template-centos         # Here we put a kustomization that uses ippool and kubeadm-config from bases
    ├── cluster-template-ubuntu         # Here we put a kustomization that uses ippool and kubeadm-config from bases
    └── cluster-template-ubuntu-workload  # Here kustomization that uses only kubeadm-config
```

test/e2e/data/infrastructure-metal3/cluster-template-upgrade-workload/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+resources:
+- ../bases/cluster
+patchesStrategicMerge:
+  - ../cluster-template-ubuntu/controlplane-kubeadm-config.yaml
+  - ../cluster-template-ubuntu/workers-kubeadm-config.yaml
```
lentzi90, line 5: This would change to 
```suggestion
resources:
- ../bases/ubuntu-kubeadm-config
```

lentzi90 (commented):
Nice job!
I just want to get rid of the `--load-restrictor` flag again. Then let's get this merged!

test/e2e/data/infrastructure-metal3/cluster-template-upgrade-workload/kustomization.yaml
mboukhalfa: I think both ways have pros and cons instead of having everything in bases I think it is relatively explicit to to have patches in other folders in changes in different files rather then having duplication in bases based on the OS name

lentzi90: It depends on how you look at it I guess. But if you don't want more bases we could also solve it by pointing the `cluster-template-ubuntu` to use the `cluster-template-ubuntu-workload` under `resources` instead. IT would have the same effect so we can drop `--load-restrictors`.

Something like this:

```yaml
# cluster-template-ubuntu
resources:
- ../bases/ippool
- ../cluster-template-ubuntu-workload
```

```yaml
# cluster-template-ubuntu
resources:
- ../bases/cluster
patchesStrategicMerge:
  - ./cluster-template-ubuntu/controlplane-kubeadm-config.yaml
  - ./cluster-template-ubuntu/workers-kubeadm-config.yaml
```

test/e2e/config/e2e_conf.yaml
furkatgofurov7: @mboukhalfa why is that, it was using its own metadata for each API version?

mboukhalfa: Well I am convinced that first it not shared since it is specific for the provider and also one file can include all releases and it is clearly mapping it to the contract. 
Basically we were including all the previous release for the current contract I mean v1alpha4 was including v1alpha3 release so it would harm if it got the newer also like v1beta1. Although capi uses multiple file I will check the reason behind that

test/e2e/data/infrastructure-metal3/cluster-template-upgrade-workload/kustomization.yaml
mboukhalfa: I tried 
```yaml
# ubuntu
resources:
- ../bases/cluster
- ../bases/ippool
- ../cluster-template-upgrade-workload
patchesStrategicMerge:
  - ./controlplane-kubeadm-config.yaml
  - ./workers-kubeadm-config.yaml
```
```yaml
# cluster-template-upgrade-workload
resources:
- ../bases/cluster
patchesStrategicMerge:
  - ../cluster-template-ubuntu/controlplane-kubeadm-config.yaml
  - ../cluster-template-ubuntu/workers-kubeadm-config.yaml

```
But still failing 
`security; file 'cluster-api-provider-metal3/test/e2e/data/infrastructure-metal3/cluster-template-ubuntu/controlplane-kubeadm-config.yaml' is not in or below 'cluster-api-provider-metal3/test/e2e/data/infrastructure-metal3/cluster-template-upgrade-workload'`

lentzi90: The point is you need to move the files so you never have `..` for `patchesStrategicMerge`. So you can move them to `cluster-template-upgrade-workload` and use that as "base" for `cluster-template-ubuntu`.

```
└── infrastructure-metal3
    ├── bases
    │   ├── cluster
    │   └── ippool
    ├── cluster-template-centos
    ├── cluster-template-ubuntu
    └── cluster-template-upgrade-workload
        ├── controlplane-kubeadm-config.yaml
        └── workers-kubeadm-config.yaml
```

```yaml
# cluster-template-upgrade-workload
resources:
- ../bases/cluster
patchesStrategicMerge:
  # NOTE: No double dots here!
  - ./cluster-template-ubuntu/controlplane-kubeadm-config.yaml
  - ./cluster-template-ubuntu/workers-kubeadm-config.yaml
```

Now since cluster-template-upgrade-workload already has everything we need except the IP pool, we just add that in the cluster-template-ubuntu:
```yaml
# cluster-template-ubuntu
resources:
- ../bases/ippool
- ../cluster-template-upgrade-workload
# No patches needed, we just grab the IP pool and the cluster-template-upgrade-workload.
```

mboukhalfa: Semantically upgrade use ubuntu as a base because we can also use upgrade based on centos if we want and then it does not make sense to me it is strange that ubuntu template is based on the upgrade, I prefer that we keep this kustomize folder as it is ?

lentzi90: Then please go with my first suggestion of moving them to bases. Referencing patches outside of the folder is very confusing and makes it impossible to reuse the kustomization. This is why I want to get rid of it, and why there is a flag for it: it is not the "normal" way.

Having the kubeadm-config in the bases is quite nice. Then we have one base for the cluster, one for the IP pool and one each for ubuntu and centos specific clusters. Then in the overlays we just pick what we need, e.g. ubuntu + IP pool.

hack/e2e/environment.sh
```diff
@@ -21,6 +21,24 @@ function os_check() {
...
+    export CAPI_VERSION="v1alpha4"
+    export CAPM3_VERSION="v1alpha5"
+fi
+
+# Override project infra vars that point to 
+# the current branch to build capm3 image and crds
+if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then 
+    export CAPM3RELEASE="v0.5.5"
+    export CAPIRELEASE="v0.4.7"
+    export CAPM3BRANCH="release-0.5"
```
furkatgofurov7, line 35: These are not needed to be set at all, see why:

- CAPM3RELEASE is set to the [latest](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/releases.sh#L47) release for v1a5 API of CAPM3 when you export it on line 27
- Same for CAPI release, it is set to  is set to the [latest](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/releases.sh#L37) release for v1a4 API of CAPI when you export it on line 26
- CAPM3BRANCH is [set](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L146) accordingly based on the CAPM3RELEASE

hack/e2e/environment.sh
furkatgofurov7: This is set to the same location as you are trying to set later, basically, you are unsetting/setting back to the same location IMO.

hack/e2e/environment.sh
```diff
@@ -21,6 +21,24 @@ function os_check() {
...
+    # This var is set in project infra to use the current repo location for 
+    # building CAPM3 image while upgrade needs an old version 
+    unset CAPM3_LOCAL_IMAGE
+    export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
+    export CAPM3PATH="${M3PATH}/cluster-api-provider-metal3"
```
furkatgofurov7, line 40: they are same when exported in m3-dev-env, why to redefine here?

hack/e2e/environment.sh
mboukhalfa: This is hard coded to a fixed release because on the e2e config file I am still had coding release just this that will be conflict between v1beta1 and v1alpha5 release if we parameterize them we can work on parameterize that separately

hack/e2e/environment.sh
mboukhalfa: I am trying to override the values set by [project infra](https://github.com/metal3-io/project-infra/blob/71f1a1fd319ac7210d51d280ed93e9b89ecc2e3e/jenkins/scripts/integration_test_env.sh#L63) not dev-env

test/e2e/data/infrastructure-metal3/cluster-template-upgrade-workload/kustomization.yaml
mboukhalfa: done

lentzi90 (approved):

test/e2e/data/infrastructure-metal3/cluster-template-upgrade-workload/kustomization.yaml
lentzi90: Thanks LGTM now! :slightly_smiling_face:

hack/e2e/environment.sh
furkatgofurov7: Okay, I was suggesting to remove from the file not to parameterize it

furkatgofurov7 (approved):
/retitle 🌱 Add v1alpha5 to v1beta e2e upgrade tests
This looks in quite good shape now, we can follow up on minor improvements later.

/approve

hack/e2e/environment.sh
mboukhalfa: once the config file will be parameterized we can remove them to use always the latest but for now better to keep this fixed here

PR #468, opened 2022-01-14
:seedling: Improve controller unit test syntax part 1

Rozzii:
syntax.

### DIFF
--- a/controllers/metal3machine_controller_test.go
+++ b/controllers/metal3machine_controller_test.go
@@ -77,7 +77,8 @@ func setReconcileNormalExpectations(ctrl *gomock.Controller,
 	// Bootstrap data not ready, we'll requeue, not call anything else
 	m.EXPECT().IsBootstrapReady().Return(!tc.BootstrapNotReady)
 	if tc.BootstrapNotReady {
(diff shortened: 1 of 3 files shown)


macaptain:
Hm, I'm not sure that expanding `tc` to `testCase` or `m` to `mockMachineManager` is desirable.

If you look at the tests in Golang's own source tree, `tc` is standard boilerplate for 'test case': https://github.com/golang/go/search?q=tc

And in gomock, `m` is the standard name for mocks (or `mr` for 'mock recorder): https://github.com/golang/mock/blob/main/sample/mock_user_test.go

We should try to copy the conventions used in Golang itself.

In my opinion, expanding the short variables makes the code less readable. The short names used by convention help to take focus away from what is test boilerplate code.

lentzi90:
> In my opinion, expanding the short variables makes the code less readable. The short names used by convention help to take focus away from what is test boilerplate code, so that we can focus on the logic of the test instead.

I'm torn. I agree that following conventions is good, but honestly when I jump on a new project and start reading through the tests I do find short variables frustrating. It takes time to find where it was defined and understand what it is, and then some cognitive load to remember it when getting back to what I was looking for at first. But of course it gets easier after a while when I'm more familiar with the project and maybe that's the way it should be. :shrug:

PR #471, opened 2022-01-17
debug CONFIG_FILE_PATH value

mboukhalfa:
debug

### DIFF
--- a/test/e2e/pivoting_test.go
+++ b/test/e2e/pivoting_test.go
@@ -46,7 +46,7 @@ func pivoting() {
 	}
 	_, err := targetClusterClientSet.CoreV1().Namespaces().Create(ctx, ironicNamespace, metav1.CreateOptions{})
 	Expect(err).To(BeNil(), "Unable to create the Ironic namespace")
-
+	Logf("os.Getenv: %v", os.Getenv("CONFIG_FILE_PATH"))
 	By("Initialize Provider component in target cluster")
 	clusterctl.Init(ctx, clusterctl.InitInput{
 		KubeconfigPath: targetCluster.GetKubeconfigPath(),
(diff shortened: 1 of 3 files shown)


test/e2e/pivoting_test.go
```diff
@@ -46,7 +46,7 @@ func pivoting() {
 	}
 	_, err := targetClusterClientSet.CoreV1().Namespaces().Create(ctx, ironicNamespace, metav1.CreateOptions{})
 	Expect(err).To(BeNil(), "Unable to create the Ironic namespace")
-
+	Logf("os.Getenv: %v", os.Getenv("CONFIG_FILE_PATH"))
```
furkatgofurov7, line 49: I do not think this is the right way to access this var when using Logf, can we try, i.e:
```suggestion
        var (
		configFilePath = e2eConfig.GetVariable("CONFIG_FILE_PATH")
	)
	Logf("config file path is: %v", configFilePath)
```

furkatgofurov7 (changes requested):

test/e2e/pivoting_test.go
mboukhalfa: Should I use  e2eConfig.GetVariable("CONFIG_FILE_PATH") also in : 
```
By("Initialize Provider component in target cluster")
	clusterctl.Init(ctx, clusterctl.InitInput{
		KubeconfigPath: targetCluster.GetKubeconfigPath(),
		// TODO (furkat), to uplift CAPI to v1.1.0-beta.1 pre-release
		// in pivoting tests we are pointing to overrides folder as ClusterctlConfigPath.
		// We have to revert this to point to e2e ClusterctlConfigPath once there is an
		// actual latest release with v1.1.X tag.
	ClusterctlConfigPath:    e2eConfig.GetVariable("CONFIG_FILE_PATH")
```
To make sure it has the same value printed?

furkatgofurov7: Sure, go ahead

PR #474, opened 2022-01-18
:seedling: Improve baremetal unit test syntax part 1

Rozzii:
syntax.


lentzi90 (approved):
Nice!

PR #475, opened 2022-01-18
:seedling: Improve baremetal unit test syntax part 2

Rozzii:
syntax.

### DIFF
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -618,8 +618,8 @@ var _ = Describe("Metal3Machine manager", func() {
 
 		DescribeTable("Test ChooseHost",
 			func(tc testCaseChooseHost) {
-				c := fakeclient.NewClientBuilder().WithScheme(setupScheme()).WithObjects(tc.Hosts...).Build()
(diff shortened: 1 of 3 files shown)


baremetal/metal3machine_manager_test.go
```diff
@@ -618,8 +618,8 @@ var _ = Describe("Metal3Machine manager", func() {
 
 		DescribeTable("Test ChooseHost",
 			func(tc testCaseChooseHost) {
-				c := fakeclient.NewClientBuilder().WithScheme(setupScheme()).WithObjects(tc.Hosts...).Build()
-				machineMgr, err := NewMachineManager(c, nil, nil, tc.Machine,
+				fakeClient := fakeclient.NewClientBuilder().WithScheme(setupScheme()).WithObjects(tc.Hosts...).Build()
```
lentzi90, line 621: It looks a bit dangerous to me to have both `fakeclient` and `fakeClient`. In some other places I think we have just `fake` for the import. What do you think about doing the same here?
```suggestion
				fakeClient := fake.NewClientBuilder().WithScheme(setupScheme()).WithObjects(tc.Hosts...).Build()
```

PR #483, opened 2022-01-21
 ✨ Add re-inspection API tests to E2E

furkatgofurov7:
[Feature tests in m3-dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/Makefile#L57) contain re-inspection API tests from BMO and e2e misses that test for now. So:

- This adds missing re-inspection API e2e tests to the framework and is called during the remediation tests instead of being called from e2e_test.go separately to save time and not prolong e2e tests running time. To call it separately, we had to do scale up/down operations, since inspection can be tested only with "Available" BMHs'.

More about re-inspection API: https://github.com/metal3-io/baremetal-operator/blob/main/docs/inspectAnnotation.md

### DIFF
--- a/test/e2e/remediation_test.go
+++ b/test/e2e/remediation_test.go
@@ -10,7 +10,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	bmh "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	bmo "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	capm3 "github.com/metal3-io/cluster-api-provider-metal3/api/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


Rozzii (approved):
LGTM

lentzi90 (approved):

smoshiur1237 (approved):
/lgtm

test/e2e/remediation_test.go
```diff
@@ -99,10 +100,14 @@ func remediation() {
...
 		}
-		g.Expect(filterBmhsByProvisioningState(bmhs, bmh.StateAvailable)).To(HaveLen(newReplicaCount + len(workerM3Machines)))
+		g.Expect(filterBmhsByProvisioningState(bmhs, bmo.StateAvailable)).To(HaveLen(newReplicaCount + len(workerM3Machines)))
 		return nil
 	}, e2eConfig.GetIntervals(specName, "wait-machine-remediation")...).Should(Succeed())
 
+	// Calling an inspection tests here for now until we have a parallelism enabled in e2e framework.
+	inspection()
+
+	Logf("Start checking unhealthy annotation")
```
kashifest, line 110: A similar Log entry before you call inspection() ? Something like `Starting re-introspection test` ?

kashifest (commented):
lgtm, a small nit in line.

test/e2e/remediation_test.go
furkatgofurov7: Thanks for reviewing @kashifest . That log is already inside the inspection() call, see here https://github.com/metal3-io/cluster-api-provider-metal3/pull/483/files#diff-968216775903e52091aa51d52ca10af4ec608c73457ce10aaa29a9b05c82bb0bR17

kashifest: Ow. ok. All good then

PR #485, opened 2022-01-24
✨ Add re-inspection API tests to E2E in release-0.5 branch

furkatgofurov7:
[Feature tests in m3-dev-env](https://github.com/metal3-io/metal3-dev-env/blob/master/Makefile#L57) contain re-inspection API tests from BMO and e2e in release-0.5 branch misses that test for now. So:

- This adds missing re-inspection API e2e tests to the framework in the `release-0.5` branch of CAPM3 and is called during the remediation tests instead of being called from e2e_test.go separately to save time and not prolong e2e tests running time. To call it separately, we had to do scale up/down operations, since inspection can be tested only with "Available" BMHs'.

More about re-inspection API: https://github.com/metal3-io/baremetal-operator/blob/main/docs/inspectAnnotation.md

Backport of https://github.com/metal3-io/cluster-api-provider-metal3/pull/483

### DIFF
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -104,6 +104,7 @@ intervals:
   default/wait-cp-available: ["50m", "30s"]
   default/wait-bmh-deprovisioning: ["50m", "10s"]
   default/wait-bmh-available: ["50m", "20s"]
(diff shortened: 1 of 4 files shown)


kashifest (approved):

smoshiur1237 (approved):
/lgtm

PR #486, opened 2022-01-24
✨ Add BMO upgrade e2e test

lentzi90:

Adds a simple BMO upgrade e2e test.

This test is very similar to the ironic upgrade test. So far they do not
really upgrade from an older version to the latest. Instead they upgrade
from no tag (latest) to main. So we at least see that it is possible to
rollout an upgrade.

The idea is to later make the tests start from an older release and upgrade to the latest.


mboukhalfa (approved):
LGTM 
/test-v1b1-centos-e2e

Rozzii (approved):
LGTM

furkatgofurov7 (commented):
/lgtm

PR #488, opened 2022-01-26
 ✨ Allow Capm3 Fast Track in bmh when AutomatedCleaning is enabled

Sunnatillo:
This PR adds:

- Capm3FastTrack variable and sets it as default to false 
- A possibility where  bmh.Spec.Online field is changed to on/off during deprovisioning based on values of     
AutomatedCleaningMode and Capm3FastTrack

```
AutomatedCleaningMode   | Capm3FastTrack |   BMH
	disabled           false            turn off
	disabled           true             turn off
	metadata           false            turn off
	metadata           true             turn on
```

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"os"
 	"strings"
 	"time"
 
@@ -64,6 +65,10 @@ const (
 	PausedAnnotationKey = "metal3.io/capm3"
 	ProviderIDPrefix    = "metal3://"
 	ProviderLabelPrefix = "metal3.io/uuid"
+)
+
+var (
+	Capm3FastTrack = os.Getenv("CAPM3_FAST_TRACK")
 )
 
 // MachineManagerInterface is an interface for a MachineManager
@@ -535,8 +540,29 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			host.Spec.NetworkData = nil
 			bmhUpdated = true
 		}
-		if host.Spec.Online {
+
+		//	Change bmh's online status to on/off  based on AutomatedCleaningMode and Capm3FastTrack values
+		//	AutomatedCleaningMode |	Capm3FastTrack|   BMH
+		//		disabled				false 			turn off
+		//		disabled				true 			turn off
+		//		metadata				false 			turn off
+		//		metadata				true 			turn on
+
+		onlineStatus := host.Spec.Online
+		if Capm3FastTrack == "" {
+			Capm3FastTrack = "false"
+			m.Log.Info("Capm3FastTrack is not set, setting it to default value false")
+		}
+		if host.Spec.AutomatedCleaningMode == "disabled" {
 			host.Spec.Online = false
+		} else if Capm3FastTrack == "true" {
+			host.Spec.Online = true
+		} else if Capm3FastTrack == "false" {
+			host.Spec.Online = false
+		}
+		m.Log.Info(fmt.Sprintf("Host %v AutomatedCleaningMode is %v, setting Online field to %v", host.Name, host.Spec.AutomatedCleaningMode, host.Spec.Online))
+
+		if onlineStatus != host.Spec.Online {
 			bmhUpdated = true
 		}
 
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -484,6 +484,11 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 	// clear an error if one was previously set
 	m.clearError()
 
+	if Capm3FastTrack == "" {
+		Capm3FastTrack = "false"
+		m.Log.Info("Capm3FastTrack is not set, setting it to default value false")
+	}
+
 	host, helper, err := m.getHost(ctx)
 	if err != nil {
 		return err
@@ -549,10 +554,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 		//		metadata				true 			turn on
 
 		onlineStatus := host.Spec.Online
-		if Capm3FastTrack == "" {
-			Capm3FastTrack = "false"
-			m.Log.Info("Capm3FastTrack is not set, setting it to default value false")
-		}
+
 		if host.Spec.AutomatedCleaningMode == "disabled" {
 			host.Spec.Online = false
 		} else if Capm3FastTrack == "true" {
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -145,6 +145,7 @@ func bmhSpec() *bmh.BareMetalHostSpec {
 		Image: &bmh.Image{
 			URL: "myimage",
 		},
+		Online: true,
 	}
 }
 
@@ -174,6 +175,7 @@ func bmhSpecSomeImg() *bmh.BareMetalHostSpec {
 		MetaData:    &corev1.SecretReference{},
 		NetworkData: &corev1.SecretReference{},
 		UserData:    &corev1.SecretReference{},
+		Online:      true,
 	}
 }
 
@@ -1426,7 +1428,6 @@ var _ = Describe("Metal3Machine manager", func() {
 		NodeReuseEnabled                bool
 		MachineIsControlPlane           bool
 		MachineIsNotControlPlane        bool
-		CheckBMHOnlineStatus            bool
 		ExpectedBMHOnlineStatus         bool
 		capm3fasttrack                  string
 	}
@@ -1443,9 +1444,8 @@ var _ = Describe("Metal3Machine manager", func() {
 			if tc.BMCSecret != nil {
 				objects = append(objects, tc.BMCSecret)
 			}
-			if tc.CheckBMHOnlineStatus {
-				Capm3FastTrack = tc.capm3fasttrack
-			}
+
+			Capm3FastTrack = tc.capm3fasttrack
 
 			fakeClient := fake.NewClientBuilder().WithScheme(setupSchemeMm()).WithObjects(objects...).Build()
 
@@ -1561,7 +1561,8 @@ var _ = Describe("Metal3Machine manager", func() {
 				Expect(err).NotTo(HaveOccurred())
 				Expect(m3mTemplate.Spec.NodeReuse).To(BeTrue())
 			}
-			if tc.CheckBMHOnlineStatus {
+
+			if tc.Host != nil {
 				savedbmh := bmh.BareMetalHost{}
 				err = fakeClient.Get(context.TODO(),
 					client.ObjectKey{
@@ -1571,9 +1572,7 @@ var _ = Describe("Metal3Machine manager", func() {
 					&savedbmh,
 				)
 				Expect(err).NotTo(HaveOccurred())
-				if tc.capm3fasttrack != "" {
-					Expect(Capm3FastTrack).To(Equal(tc.capm3fasttrack))
-				} else if tc.capm3fasttrack == "" {
+				if tc.capm3fasttrack == "" {
 					Expect(Capm3FastTrack).To(Equal("false"))
 				}
 				Expect(savedbmh.Spec.Online).To(Equal(tc.ExpectedBMHOnlineStatus))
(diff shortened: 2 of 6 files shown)


config/manager/manager.yaml
```diff
@@ -47,3 +47,11 @@ spec:
...
         - effect: NoSchedule
           key: node-role.kubernetes.io/master
+---
+kind: ConfigMap 
+apiVersion: v1 
+metadata:
+  name: fast-track-cm
+  namespace: system
+data:	
+  fastTrack: "false"
```
dtantsur, line 57: `fastTrack` defaults to true in ironic-image, why false here?

controllers/metal3machine_controller.go
```diff
@@ -68,6 +68,7 @@ type Metal3MachineReconciler struct {
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=kubeadmcontrolplanes,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
```
Xenwar, line 71: Do we need all verbs ? if so, replaced by *
Also, should this scoped to some namespace ?
```suggestion
// +kubebuilder:rbac:groups="",resources=configmaps,verbs=*
```

baremetal/metal3machine_manager.go
```diff
@@ -1707,3 +1728,22 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
 	}
 	return nodes, nodesCount, err
 }
+
+func getValueFromCM(m *MachineManager, ctx context.Context, configmapKey client.ObjectKey) string {
+	original := &corev1.ConfigMap{}
+
+	if err := m.client.Get(ctx, configmapKey, original); err != nil {
+		m.Log.Info("Failed to extract ConfigMap")
```
Xenwar, line 1736: could this be "failed to get configmap" ? to make it different from the error on line 1741

baremetal/metal3machine_manager.go
```diff
@@ -1707,3 +1728,22 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+	}
+
+	fastTrack, ok := original.DeepCopy().Data["fastTrack"]
+	if !ok {
+		m.Log.Info("Failed to extract FastTrack from ConfigMap")
+	}
+
+	if fastTrack == "" {
+		m.Log.Info("Set fastTrack varible value to false")
```
Xenwar, line 1745:
```suggestion
		m.Log.Info("No fastTrack value is given, setting it to false")
```

baremetal/metal3machine_manager.go
```diff
@@ -533,8 +533,29 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+		//	Automated Cleaning | Fast Track	| BMH
+		//	disabled			false 		turn off
+		//	disabled			true 		turn off
+		//	metadata			false 		turn off
+		//	metadata 			true 		turn on
+
+		onlineStatus := host.Spec.Online
+
+		//Extract fastTrack value from configMap
+		configMapKey := client.ObjectKey{Name: "capm3-fast-track-cm", Namespace: "capm3-system"}
```
Xenwar, line 548: parameterize the configmap and namespace parameters

baremetal/metal3machine_manager.go
```diff
@@ -533,8 +533,29 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			host.Spec.NetworkData = nil
 			bmhUpdated = true
 		}
-		if host.Spec.Online {
+
+		//	turn on/off bmh depending on AutomatedCleaningMode and fastTrack values
```
Xenwar, line 537:
```suggestion
		//	Change bmh's online status to on/off  based on AutomatedCleaningMode and fastTrack values
```

baremetal/metal3machine_manager.go
```diff
@@ -533,8 +533,29 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+
+		//	Automated Cleaning | Fast Track	| BMH
+		//	disabled			false 		turn off
+		//	disabled			true 		turn off
+		//	metadata			false 		turn off
+		//	metadata 			true 		turn on
```
Xenwar, line 543: It is a bit difficult to see which values go with which column, fix the alignment.

macaptain, line 543: Can we have unit tests covering these cases? There are already some tests for the `Delete` function, so you should be able to extend these tests and cover all the behaviors: 

https://github.com/metal3-io/cluster-api-provider-metal3/blob/6ccfd3bf9863794e1b74cf408556be6f8809309a/baremetal/metal3machine_manager_test.go#L1420

config/manager/manager.yaml
Sunnatillo: Currently capm3 powers off the BMH after deprovisioning. Not to change the current behaviour we decided that default value to be "false".  
Default value for AutomatedCleaningMode is "metadata".

baremetal/metal3machine_manager.go
Sunnatillo: done

controllers/metal3machine_controller.go
Sunnatillo: We did not specify  namespaces for other resources. If you think this is the right way, I can add

baremetal/metal3machine_manager.go
Sunnatillo: done

baremetal/metal3machine_manager.go
Sunnatillo: done

baremetal/metal3machine_manager.go
Sunnatillo: done

controllers/metal3machine_controller.go
furkatgofurov7: @Xenwar are we sure we can replace verbs with *, if so can you please reference to the example where it has been used? I have not seen it in practice so far and [kubeuilder](https://book.kubebuilder.io/reference/markers/rbac.html) book does not seem to mention that

config/manager/manager.yaml
dtantsur: It doesn't only affect deprovisioning. It also means that you will have 2 additional reboots: after inspection and after initial cleaning (before deployment). This makes provisioning significantly longer.

If CAPM3 powers off BMH itself, the fast track value will not matter: Ironic will obey.

Sunnatillo: fastTrack variable is not tied the one with in Ironic. This fastTrack variable is just to control whether CAPM3 should turn off the BMH after deprovisioning or not.

controllers/metal3machine_controller.go
Xenwar: It is a valid syntax. You can use `* ` for resources or verbs. Here is one example for resources.
https://github.com/kubernetes-sigs/cluster-addons/blob/b8b26d8aa73c917c00ed58871de87bd8a3b6c56a/dashboard/controllers/dashboard_controller.go#L91

But, I agree with making it like other controllers. So, we can keep the list of verbs.

baremetal/metal3machine_manager.go
```diff
@@ -533,8 +534,35 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+		//		disabled				true 			turn off
+		//		metadata				false 			turn off
+		//		metadata				true 			turn on
+
+		//get capm3FastTrack from env
+		capm3FastTrack := os.Getenv("CAPM3_FAST_TRACK")
+		if capm3FastTrack == "" {
+			m.Log.Info("No CAPM3 FAST TRACK environment variable is given, setting it to false")
+			capm3FastTrack = "false"
+		}
```
maelk, line 551: you should get the variable at setup time, since it won't change. So you could have this as a package variable and set before you add the reconciler in main.go

baremetal/metal3machine_manager.go
```diff
@@ -533,8 +534,35 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
+		if host.Spec.AutomatedCleaningMode == "disabled" {
+			m.Log.Info(fmt.Sprintf("Host %v AutomatedCleaningMode is %v, setting Online field to false", host.Name, host.Spec.AutomatedCleaningMode))
 			host.Spec.Online = false
+		} else if capm3FastTrack == "true" {
+			m.Log.Info(fmt.Sprintf("CAPM3 FAST TRACK value is %v, Host %v AutomatedCleaningMode is %v, setting Online field to true", capm3FastTrack, host.Name, host.Spec.AutomatedCleaningMode))
+			host.Spec.Online = true
+		} else if capm3FastTrack == "false" {
+			m.Log.Info(fmt.Sprintf("CAPM3 FAST TRACK value is %v, Host %v AutomatedCleaningMode is %v, setting Online field to false", capm3FastTrack, host.Name, host.Spec.AutomatedCleaningMode))
+			host.Spec.Online = false
+		}
```
maelk, line 564:
```suggestion
		if host.Spec.AutomatedCleaningMode == "disabled" {
			host.Spec.Online = false
		} else if capm3FastTrack == "true" {
			host.Spec.Online = true
		} else if capm3FastTrack == "false" {
			host.Spec.Online = false
		}
		m.Log.Info(fmt.Sprintf("CAPM3 FAST TRACK value is %v, Host %v AutomatedCleaningMode is %v, setting Online field to %s", capm3FastTrack, host.Name, host.Spec.AutomatedCleaningMode, host.Spec.Online))
```

maelk: What about this ? It's more visual for the cases

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,10 @@ kind: Kustomization
 resources:
   - manager.yaml
 
+configMapGenerator:
+- name: fast-track-configmap
+  literals:
+  - CAPM3_FAST_TRACK=${CAPM3_FAST_TRACK}
```
maelk, line 10: you should default to false here to keep the same behaviour as before if it is unset

baremetal/metal3machine_manager.go
Sunnatillo: If we declare camp3FastTrack inside main.go (main package), how we use it in baremetal package? 
Can please you clarify what do you mean by package variable?

baremetal/metal3machine_manager.go
Sunnatillo: added

baremetal/metal3machine_manager.go
Sunnatillo: done

config/manager/manager.yaml
Sunnatillo: @dtantsur, Not to confuse Ironic fast track variable with the one introduced in this patch, I have renamed to Capm3FastTrack

baremetal/metal3machine_manager.go
```diff
@@ -535,8 +545,26 @@ func (m *MachineManager) Delete(ctx context.Context) error {
...
 			bmhUpdated = true
 		}
-		if host.Spec.Online {
+
+		//	Change bmh's online status to on/off  based on AutomatedCleaningMode and Capm3FastTrack values
+		//	AutomatedCleaningMode |	Capm3FastTrack|   BMH
+		//		disabled				false 			turn off
+		//		disabled				true 			turn off
+		//		metadata				false 			turn off
+		//		metadata				true 			turn on
```
Xenwar, line 554: Not critical, but aligning the fields could improve readability.

Sunnatillo: I tried to align lines for both code editor and for github, this was max I came up with. If I make alignment for one, in the other it will be unaligned.

baremetal/metal3machine_manager_test.go
```diff
@@ -1554,34 +1561,54 @@ var _ = Describe("Metal3Machine manager", func() {
...
+				err = fakeClient.Get(context.TODO(),
+					client.ObjectKey{
+						Name:      tc.Host.Name,
+						Namespace: tc.Host.Namespace,
+					},
+					&savedbmh,
+				)
+				Expect(err).NotTo(HaveOccurred())
+				if tc.capm3fasttrack != "" {
+					Expect(Capm3FastTrack).To(Equal(tc.capm3fasttrack))
```
kashifest, line 1576: Trying to understand what are we testing here? We are setting `Capm3FastTrack = tc.capm3fasttrack` few lines back.

Sunnatillo: yes we are setting it earlier, if you look to full condition it is testing the cases when we  set  and we do not set the Capm3fasttrack value

kashifest: My point is this will always be true` Expect(Capm3FastTrack).To(Equal(tc.capm3fasttrack)) ` since we are setting `Capm3FastTrack = tc.capm3fasttrack` few lines back.

kashifest (approved):

PR #511, opened 2022-02-08
🌱 Remove install_kustomize.sh under hack/tools and use ensure_kustomize.sh from hack folder

kashifest:
We have two files currently which installs kustomize. `hack/ensure_kustomize.sh` and `hack/tools/install_kustomize.sh`. This
PR removes hack/tools/install_kustomize.sh and also uplifts kustomize version to 4.4.1 to keep it aligned with CAPI.


kashifest:
I will open a similar PR in IPAM repo once this PR is reviewed and agreed upon.

lentzi90 (commented):
Could we not use the exact same script they use in CAPI? I see there are some differences now when compared to [this](https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/ensure-kustomize.sh).

kashifest:
> Could we not use the exact same script they use in CAPI? I see there are some differences now when compared to [this](https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/ensure-kustomize.sh).

Yes there are some differences, I dont recall why this is like this, however I have tried to keep it as close to CAPI as possible and still keep the rest of the code which seemed useful to me. Do you see any part of the current code is not useful?

lentzi90:
> Yes there are some differences, I dont recall why this is like this, however I have tried to keep it as close to CAPI as possible and still keep the rest of the code which seemed useful to me. Do you see any part of the current code is not useful?

I think we could do the [same check](https://github.com/kubernetes-sigs/cluster-api/blob/d38a4b1c13b5f26ac407eab85b83597662f630bd/hack/ensure-kustomize.sh#L28-L31) for `goos` as they do instead of [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/7dfb5ebecf8a3399f12674a016a6feb177798712/hack/ensure-kustomize.sh#L33). But I'm not sure.

Then the check for minimum version [at the end](https://github.com/metal3-io/cluster-api-provider-metal3/blob/7dfb5ebecf8a3399f12674a016a6feb177798712/hack/ensure-kustomize.sh#L49-L57) I guess is useful for long running environments so we don't get stuck on an old release. So I guess we should keep it. Maybe we could upstream it? :slightly_smiling_face:

kashifest:
> > Yes there are some differences, I dont recall why this is like this, however I have tried to keep it as close to CAPI as possible and still keep the rest of the code which seemed useful to me. Do you see any part of the current code is not useful?
> 
> I think we could do the [same check](https://github.com/kubernetes-sigs/cluster-api/blob/d38a4b1c13b5f26ac407eab85b83597662f630bd/hack/ensure-kustomize.sh#L28-L31) for `goos` as they do instead of [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/7dfb5ebecf8a3399f12674a016a6feb177798712/hack/ensure-kustomize.sh#L33). But I'm not sure.
> 
> Then the check for minimum version [at the end](https://github.com/metal3-io/cluster-api-provider-metal3/blob/7dfb5ebecf8a3399f12674a016a6feb177798712/hack/ensure-kustomize.sh#L49-L57) I guess is useful for long running environments so we don't get stuck on an old release. So I guess we should keep it. Maybe we could upstream it? slightly_smiling_face

I am unsure if we should introduce the same check here for OS, so I am keeping this out of this PR. Also I didn't get your point about upstream, you mean CAPI ?

lentzi90:
> I am also unsure if we should introduce the same check here for OS, so I am keeping this out of this PR. Also I didn't get your point about upstream, you mean CAPI ?

Fair enough, and yes I meant CAPI. But we can take a look at that later.

furkatgofurov7 (approved):
/approve
/retitle 🌱 Remove install_kustomize.sh under hack/tools and use ensure_kustomize.sh from hack folder

PR #517, opened 2022-02-14
🐛 Reconcile Metal3MachineTemplate on Metal3Machine events

fmuyassarov:
Normally, automated cleaning mode (either `disabled` or `metadata`) is replicated from a
Metal3MachineTemplate spec to Metal3Machine spec. It is allowed to patch Metal3Machine's
`automtedCleaningMode` directly, and if that's the case  CAPM3 controller should notice it,
request a reconciliation for the Metal3MachineTemplate and reset the `automatedCleaningMode`
on the Metal3Machine to the same as in Metal3MachineTemplate. This ensures that the source
of the truth is always in the template.

### DIFF
--- a/controllers/metal3machinetemplate_controller.go
+++ b/controllers/metal3machinetemplate_controller.go
@@ -21,15 +21,21 @@ import (
 	"github.com/metal3-io/cluster-api-provider-metal3/baremetal"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
 	templateControllerName = "Metal3MachineTemplate-controller"
+	clonedFromGroupKind    = capi.TemplateClonedFromGroupKindAnnotation
+	clonedFromName         = capi.TemplateClonedFromNameAnnotation
 )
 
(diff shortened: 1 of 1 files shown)


controllers/metal3machinetemplate_controller.go
```diff
@@ -113,6 +119,32 @@ func (r *Metal3MachineTemplateReconciler) reconcileNormal(ctx context.Context,
...
+			handler.EnqueueRequestsFromMapFunc(r.Metal3MachinesToMetal3MachineTemplate),
+		).
 		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(ctrl.LoggerFrom(ctx), r.WatchFilterValue)).
 		Complete(r)
 }
+
+func (m *Metal3MachineTemplateReconciler) Metal3MachinesToMetal3MachineTemplate(o client.Object) []ctrl.Request {
+	result := []ctrl.Request{}
+	// Fetch the Metal3MachineTemplate instance.
+	metal3MachineTemplate := &capm3.Metal3MachineTemplate{}
```
Xenwar, line 133: I see you are retrieving the related/referenced template, but I think the part that make the template the source of truth is missing.

fmuyassarov: That part is already in the main: https://github.com/metal3-io/cluster-api-provider-metal3/blob/13aa6e002eb43c938559a66a7e17328aa3390198/baremetal/metal3machinetemplate_manager.go#L95.
This patch intends to add a watch only.

Xenwar: Got it, thanks.

controllers/metal3machinetemplate_controller.go
```diff
@@ -113,6 +119,32 @@ func (r *Metal3MachineTemplateReconciler) reconcileNormal(ctx context.Context,
...
+	if m3m, ok := o.(*capm3.Metal3Machine); ok {
+		if m3m.Annotations[clonedFromGroupKind] != "Metal3MachineTemplate.infrastructure.cluster.x-k8s.io" && m3m.Annotations[clonedFromName] != metal3MachineTemplate.Name {
```
kashifest, line 135: Isn't this `metal3MachineTemplate.Name` supposed to be empty at this point since it was declared at line 133 and hasn't been initialized since?

kashifest (commented):
/lgtm

PR #520, opened 2022-02-15
🐛 Reconcile Metal3MachineTemplate on Metal3Machine events

fmuyassarov:
Normally, automated cleaning mode (either disabled or
metadata) is replicated from a Metal3MachineTemplate spec
to Metal3Machine spec. It is allowed to patch Metal3Machine's
automtedCleaningMode directly, and if that's the case CAPM3
controller should notice is, request a reconciliation for the
Metal3MachineTemplate and reset the automatedCleaningMode on
the Metal3Machine to the same as in Metal3MachineTemplate.
This ensures that the source of the truth is always in the
template.

### DIFF
--- a/controllers/metal3machinetemplate_controller.go
+++ b/controllers/metal3machinetemplate_controller.go
@@ -21,15 +21,21 @@ import (
 	"github.com/metal3-io/cluster-api-provider-metal3/baremetal"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	capi "sigs.k8s.io/cluster-api/api/v1alpha4"
 	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


controllers/metal3machinetemplate_controller.go
```diff
@@ -113,6 +119,30 @@ func (r *Metal3MachineTemplateReconciler) reconcileNormal(ctx context.Context,
...
 		For(&capm3.Metal3MachineTemplate{}).
+		Watches(
+			&source.Kind{Type: &capm3.Metal3Machine{}},
+			handler.EnqueueRequestsFromMapFunc(r.Metal3MachinesToMetal3MachineTemplate),
+		).
 		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(ctrl.LoggerFrom(ctx), r.WatchFilterValue)).
 		Complete(r)
 }
+
+func (m *Metal3MachineTemplateReconciler) Metal3MachinesToMetal3MachineTemplate(o client.Object) []ctrl.Request {
```
kashifest, line 130: I think we missed the unit tests for this function in main branch. I think we should add it here to test the functionality properly. We can add the same for main branch also later on.

furkatgofurov7 (approved):
/lgtm

kashifest (approved):

kashifest:
Unit tests will be added later.

PR #523, opened 2022-02-15
🌱 Parameterize api folder path in Makefile

furkatgofurov7:
Parameterizes api folder path in the Makefile targets. 

We can use just run: `make modules` instead of running `go mod tidy` in every folder where go.mod file is updated.


kashifest (approved):

fmuyassarov (commented):
/lgtm

PR #524, opened 2022-02-16
🐛 Fix IMAGE_OS case

mboukhalfa:
Since IMAGE_OS changed to lowercase in Dev_Env in this [PR](https://github.com/metal3-io/metal3-dev-env/pull/948), updates in capm3 hack file are need else it will break the test locally.


lentzi90 (approved):
LGTM!
Thanks for fixing. I didn't notice we had this in CAPM3 also...

smoshiur1237 (approved):
/lgtm

fmuyassarov (approved):

PR #527, opened 2022-02-17
🐛 Add live-iso option to Image's DiskFormat in v1a5 and v1b1

macaptain:
The image.format option of 'live-iso' was available in the v1alpha4 API but had gone missing in the v1alpha5 and v1beta1 APIs.

**What this PR does / why we need it**: The live-iso option for the disk format of images for Metal3Machines was supported in v1alpha4 and should be supported in v1alpha5 and v1beta1.

**Which issue(s) this PR fixes**:
Fixes #500

Also, this commit may need cherry-picking into release branches, I'm not sure exactly which branches would need this fix.

PR #529, opened 2022-02-17
🐛Cherrypick live-iso option to Image's DiskFormat for v1a5 in release-0.5

smoshiur1237:
This PR will add the live-iso option for v1a5 in release-0.5 branch and changes are done according to main branch https://github.com/metal3-io/cluster-api-provider-metal3/pull/527


namnx228 (approved):
/lgtm

Xenwar:
I think the yaml(crds) should be generated as well. There should be a make target.

PR #533, opened 2022-02-18
🌱 Replace jetstack from cert-manager URL and update version to v1.5.3

kashifest:


smoshiur1237 (approved):
/lgtm

lentzi90 (commented):
I was surprised to see `tilt_modules` here in the repo. My understanding is that these should most likely be put in `.gitignore` like `node_modules` and similar.

The content in the `cert_manager` tilt module is coming from [here](https://github.com/tilt-dev/tilt-extensions/tree/master/cert_manager). So if we update the files in our repo I guess they will just be overridden when we run `tilt up`. Or maybe I'm missing something?

kashifest:
> I was surprised to see `tilt_modules` here in the repo. My understanding is that these should most likely be put in `.gitignore` like `node_modules` and similar.
> 
> The content in the `cert_manager` tilt module is coming from [here](https://github.com/tilt-dev/tilt-extensions/tree/master/cert_manager). So if we update the files in our repo I guess they will just be overridden when we run `tilt up`. Or maybe I'm missing something?

I do not know TBH how this is working, I am just updating the reference URL and version here.

lentzi90:
> I do not know TBH how this is working, I am just updating the reference URL and version here.

I think it will have no effect but I'm fine with merging it. I have created an [issue upstream](https://github.com/tilt-dev/tilt-extensions/issues/346) so they will hopefully update it there also soon. Will also try to find out if tilt_modules should be included in git or not.

PR #534, opened 2022-02-18
🌱 Remove tilt_modules

lentzi90:

Tilt modules are automatically downloaded and does not need to be
included in source control. The latest release (v0.25) has also changed
to use the XDG data dir for tilt_modules instead of keeping them here
next to the Tiltfile.

See this comment: https://github.com/tilt-dev/tilt/issues/5336#issuecomment-1040808714

It also seems like the cert-manager version setting was ignored in the Tiltfile. This is fixed by this PR.


lentzi90:
I have confirmed locally that tilt still works and also that the cert-manager version is now correctly set.

fmuyassarov (commented):
looks good
/approve

furkatgofurov7:
Running this locally noticed some unwanted changes, holding

Rozzii (approved):

furkatgofurov7:
While running Tilt from this branch locally (`make tilt-settings`, `make tilt-up`), I have noticed that the all files you deleted under **tilt_modules** folder are re-generated. The first impression I had the PR description was we would not need those files at all but I do not like the fact that they will be still re-generated and will be there. 
I also checked other providers, and found out:

- https://github.com/kubernetes-sigs/cluster-api/pull/5785 in CAPI which enabled them to completely remove the files under tilt_modules

lentzi90:
@furkatgofurov7 yes Tilt will re-generate the tilt_modules automatically when you run it. Depending on what version you have they will be in the working directory (older versions) or in the XDG data directory (for latest Tilt).

What they did in CAPI was to get rid of the extensions and instead use [their own way of installing cert-manager](https://github.com/fabriziopandini/cluster-api/blob/ba114a32162652b78a02daac18504377a0f9708a/hack/tools/tilt-prepare/main.go#L325-L338). We could do something similar but that is a bigger change. My point with this was just to not have the tilt_modules in git since they are autogenerated by Tilt.

PR #537, opened 2022-02-21
🌱 Generate templates from e2e

mboukhalfa:
The e2e-tests templates were either generated from dev-env or using `envsubst` so the e2e parameters such as clusterName, replicas .. had not any effects.
This PR adds templates with parameters to the e2e tests and remove the dependency on the dev-env template. Also uses `kustomize` to generate OS customized templates.


mboukhalfa:
/assign @namnx228 
cc @Rozzii

Makefile
```diff
@@ -123,26 +123,35 @@ test-e2e: ## Run e2e tests with capi e2e testing framework
...
 	$(ENVSUBST) < $(E2E_CONF_FILE) > $(E2E_CONF_FILE_ENVSUBST)
 
+## --------------------------------------
+## Templates
+## --------------------------------------
+E2E_TEMPLATES_DIR ?= $(ROOT_DIR)/test/e2e/data/infrastructure-metal3
+.PHONY: cluster-templates
+cluster-templates: $(KUSTOMIZE) ## Generate cluster templates
+	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu --load-restrictor LoadRestrictionsNone > $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu.yaml
+	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-centos --load-restrictor LoadRestrictionsNone > $(E2E_TEMPLATES_DIR)/cluster-template-centos.yaml
```
lentzi90, line 146: We should not need to change the load restrictor. We can instead change the kustomization (see other comments).
```suggestion
	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu > $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu.yaml
	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-centos > $(E2E_TEMPLATES_DIR)/cluster-template-centos.yaml
```

Makefile
```diff
@@ -560,7 +569,6 @@ clean-examples: ## Remove all the temporary files generated in the examples fold
 
 clean-e2e:
 	rm -f $(E2E_CONF_FILE_ENVSUBST)
-	rm -rf $(E2E_ENVSUBST_DIR)
```
lentzi90, line 563: Why is this removed?

test/e2e/data/infrastructure-metal3/cluster-template-centos/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+bases:
+- ../bases/cluster-with-kcp.yaml
+- ../bases/ippool.yaml
+- ../bases/md.yaml
```
lentzi90, line 4: If you put a `kustomize.yaml` file in `bases` with this in it:
```yaml
resources:
- cluster-with-kcp.yaml
- ippool.yaml
- md.yaml
```
we could change this and also get rid of the load restrictor when building.

```suggestion
resources:
- ../base
```

(I changed `bases -> resources` since `bases` is deprecated in kustomize. They basically work the same now.)

test/e2e/data/infrastructure-metal3/cluster-template-ubuntu/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+bases:
+- ../bases/cluster-with-kcp.yaml
+- ../bases/ippool.yaml
+- ../bases/md.yaml
```
lentzi90, line 4: Same here as in the other `kustomize.yaml`.

test/e2e/e2e_test.go
```diff
@@ -34,7 +34,7 @@ var _ = Describe("Workload cluster creation", func() {
 	BeforeEach(func() {
 		osType := strings.ToLower(os.Getenv("OS"))
 		Expect(osType).ToNot(Equal(""))
-		flavorSuffix = "-m3-dev-env"
+		flavorSuffix = osType
```
lentzi90, line 37: It looks like this is no longer a suffix, so maybe change it to just `flavor`?

test/e2e/data/infrastructure-metal3/bases/cluster-with-kcp.yaml
```diff
@@ -0,0 +1,117 @@
+apiVersion: cluster.x-k8s.io/${CAPI_VERSION}
+kind: Cluster
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: ${NAMESPACE}
```
lentzi90, line 5: I think we can remove the namespace here and put it in the `kustomize.yaml` file instead.

lentzi90 (commented):
I like the use of kustomize to handle the duplication between centos and ubuntu templates!
The use of envsubst on the other hand I'm not so fond of. I was thinking that we could maybe use go templates instead? That way we could use the same variables we use in the go code to render the templates. It is not a strong preference but I would like to hear what others think.

test/e2e/data/infrastructure-metal3/cluster-template-centos/kustomization.yaml
mboukhalfa: At some point I think we might not need ippool.yaml that was the case in the upgrade (test workload template)

Makefile
mboukhalfa: What do you mean by changing the kustomization

Makefile
```diff
@@ -123,26 +123,35 @@ test-e2e: ## Run e2e tests with capi e2e testing framework
...
 .PHONY: e2e-substitutions
-e2e-substitutions: $(ENVSUBST) $(subst $(E2E_TEMPLATES_DIR),$(E2E_ENVSUBST_DIR),$(wildcard $(E2E_TEMPLATES_DIR)/*.yaml))
+e2e-substitutions: $(ENVSUBST)
+	mkdir -p $(E2E_ENVSUBST_DIR)
 	$(ENVSUBST) < $(E2E_CONF_FILE) > $(E2E_CONF_FILE_ENVSUBST)
 
+## --------------------------------------
+## Templates
+## --------------------------------------
+E2E_TEMPLATES_DIR ?= $(ROOT_DIR)/test/e2e/data/infrastructure-metal3
```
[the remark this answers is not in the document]
mboukhalfa: Done also `E2E_ENVSUBST_DIR` renamed to `E2E_OUT_DIR`

hack/e2e/environment.sh
```diff
@@ -58,3 +58,7 @@ source "${M3_DEV_ENV_PATH}/scripts/feature_tests/node_reuse/node_reuse_vars.sh"
 # Pin Calico version
 export CALICO_PATCH_RELEASE="v3.21.0"
 export CALICO_MINOR_RELEASE="v3.21"
+
+# Exported to the cluster templates
+SSH_PUB_KEY_CONTENT=$(cat "${HOME}/.ssh/id_rsa.pub")
+export SSH_PUB_KEY_CONTENT
```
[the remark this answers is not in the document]
mboukhalfa: Moved to Makefile

Makefile
mboukhalfa: Thanks, the line `rm -f $(E2E_CONF_FILE_ENVSUBST)` removed instead of `rm -rf $(E2E_ENVSUBST_DIR)` since it is a child file inside the Dir

test/e2e/e2e_test.go
mboukhalfa: changed

test/e2e/data/infrastructure-metal3/bases/cluster-with-kcp.yaml
mboukhalfa: As a result the namespace will be still a parameter do not see the reason for that

test/e2e/data/infrastructure-metal3/cluster-template-ubuntu/kustomization.yaml
mboukhalfa: see the comment

test/e2e/config/e2e_conf.yaml
```diff
@@ -62,16 +62,12 @@ providers:
     type: "url"
     contract: v1beta1
     files:
-    - sourcePath: "${PWD}/metadata.yaml"
+    - sourcePath: "../../../metadata.yaml"
```
namnx228, line 65: This metadata.yaml file that e2e uses is different from the one in the root directory. Please revert this change

namnx228 (changes requested):
This is a really good work. Just one thing to consider. See my comment

test/e2e/config/e2e_conf.yaml
mboukhalfa: if you mean reverting `${PWD}/metadata.yaml` it will end up using the same file

test/e2e/data/infrastructure-metal3/bases/cluster-with-kcp.yaml
lentzi90: It is just simpler. One place instead of many. Now we need to have it for the Cluster, Metal3Cluster, KubeadmControlPlane and Metal3MachineTemplates. If we put it in `kustomize.yaml` it is only in one place and applies to all of them I think.

test/e2e/data/infrastructure-metal3/cluster-template-centos/kustomization.yaml
lentzi90: In that case we should have a separate base for the IP pool. Something like this:
```
├── bases
│   ├── cluster
│   │   ├── cluster-with-kcp.yaml
│   │   ├── kustomization.yaml
│   │   └── md.yaml
│   └── ippool
│       ├── ippool.yaml
│       └── kustomization.yaml
└── overlays
    ├── centos
    │   └── kustomization.yaml
    └── ubuntu
        └── kustomization.yaml
```

Makefile
lentzi90: It is the [other comment](https://github.com/metal3-io/cluster-api-provider-metal3/pull/537#discussion_r814644367)

test/e2e/data/infrastructure-metal3/bases/cluster/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+bases:
```
lentzi90, line 1: Nit.
```suggestion
resources:
```

test/e2e/data/infrastructure-metal3/bases/ippool/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+bases:
```
lentzi90, line 1:
```suggestion
resources:
```

test/e2e/data/infrastructure-metal3/cluster-template-centos/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+bases:
```
lentzi90, line 1:
```suggestion
resources:
```

test/e2e/data/infrastructure-metal3/cluster-template-ubuntu/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+bases:
```
lentzi90, line 1:
```suggestion
resources:
```

lentzi90 (commented):
Nice!
I added suggestions to change `bases` to `resources`, since it [has been deprecated](https://github.com/kubernetes-sigs/kustomize/blob/661743c7e5bd8c3d9d6866b6bc0a6f0e0b0512eb/site/content/en/blog/releases/v2.1.0.md#field-changes--deprecations). We are still on an old version of kustomize where it is supported but it is good to prepare already for the future.

Other than that I think it is good to go in! :slightly_smiling_face:

test/e2e/config/e2e_conf.yaml
```diff
@@ -87,7 +83,8 @@ variables:
   CAPIRELEASE: "${CAPIRELEASE}"
   CAPM3RELEASE: "${CAPM3RELEASE}"
   CONFIG_FILE_PATH: "${HOME}/.cluster-api/overrides/clusterctl.yaml"
-
+  # Default vars for the template, those values could be overridden by the env-vars.
+  SERVICE_CIDR: "10.96.0.0/12"
```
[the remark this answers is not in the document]
mboukhalfa: I do not think it is exported in the env because e2e was complaining about this var while generating the template but I can add possibility to read from env else use the default value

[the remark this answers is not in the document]
mboukhalfa: Well I guess e2e will prioritize the value in the env else it use what is declared in conf file so no need to read from env in the conf file

lentzi90 (approved):

furkatgofurov7 (commented):
Good job, thanks! 
The only thing I would suggest is to rename file names to be specific:

- md-KubeadmConfigTemplate.yaml ==> workers-kubeadm-config.yaml 
- kcp-KubeadmControlPlane.yaml ==> controlplane-kubeadm-config.yaml

test/e2e/data/infrastructure-metal3/bases/ippool/kustomization.yaml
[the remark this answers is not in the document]
mboukhalfa: done

furkatgofurov7 (approved):
/retitle 🌱 Generate templates from e2e
/approve

PR #539, opened 2022-02-22
🐛 Replace update taints with patch node taints

mboukhalfa:
This PR adds: 
- Fix update nodes conflict when untainting by using patch instead of update https://jenkins.nordix.org/view/Metal3%20Main/job/metal3_main_v1a5_e2e_test_ubuntu/22/console

### DIFF
--- a/test/e2e/node_reuse_test.go
+++ b/test/e2e/node_reuse_test.go
@@ -55,7 +55,7 @@ func nodeReuse() {
 
 	By("Untaint all CP nodes before scaling down machinedeployment")
 	controlplaneNodes := getControlplaneNodes(clientSet)
-	untaintNodes(clientSet, controlplaneNodes, controlplaneTaint)
+	untaintNodes(targetClusterClient, controlplaneNodes, controlplaneTaint)
 
 	By("Scale down MachineDeployment to 0")
 	scaleMachineDeployment(ctx, targetClusterClient, clusterName, namespace, 0)
@@ -206,7 +206,7 @@ func nodeReuse() {
 
 	By("Untaint CP nodes after upgrade of two controlplane nodes")
 	controlplaneNodes = getControlplaneNodes(clientSet)
-	untaintNodes(clientSet, controlplaneNodes, controlplaneTaint)
+	untaintNodes(targetClusterClient, controlplaneNodes, controlplaneTaint)
 
 	Byf("Wait until all %v KCP machines become running and updated with new %s k8s version", numberOfControlplane, upgradedK8sVersion)
 	Eventually(
@@ -258,7 +258,7 @@ func nodeReuse() {
 	// We have untainted the 2 first CPs
 	for untaintedNodeCount := 0; untaintedNodeCount < numberOfControlplane-2; {
 		controlplaneNodes = getControlplaneNodes(clientSet)
-		untaintedNodeCount = untaintNodes(clientSet, controlplaneNodes, controlplaneTaint)
+		untaintedNodeCount = untaintNodes(targetClusterClient, controlplaneNodes, controlplaneTaint)
 		time.Sleep(10 * time.Second)
 	}
 
@@ -503,15 +503,15 @@ func updateBootImage(m3machineTemplateName string, clusterClient client.Client, 
 	Expect(helper.Patch(ctx, &m3machineTemplate)).To(Succeed())
 }
 
-func untaintNodes(clientSet *kubernetes.Clientset, nodes *corev1.NodeList, taint *corev1.Taint) (count int) {
+func untaintNodes(targetClusterClient client.Client, nodes *corev1.NodeList, taint *corev1.Taint) (count int) {
 	count = 0
 	for i := range nodes.Items {
 		Logf("Untainting node %v ...", nodes.Items[i].Name)
 		newNode, changed := removeTaint(&nodes.Items[i], taint)
 		if changed {
-			node, err := clientSet.CoreV1().Nodes().Update(ctx, newNode, metav1.UpdateOptions{})
-			Expect(err).To(BeNil(), "Failed to update nodes")
-			Logf("Node %v untainted", node.Name)
+			patchHelper, err := patch.NewHelper(&nodes.Items[i], targetClusterClient)
+			Expect(err).To(BeNil())
+			Expect(patchHelper.Patch(ctx, newNode)).To(Succeed(), "Failed to patch node")
 			count++
 		}
 	}


mboukhalfa:
/test-v1b1-ubuntu-e2e 
Failed in remediation https://jenkins.nordix.org/job/metal3_capm3_v1b1_e2e_test_ubuntu/54/

lentzi90:
I did not understand why we need "workaround to bypass failed machines". Why should we bypass? Is it not an error we need to investigate instead and fix in the controller so it doesn't happen?

mboukhalfa:
> I did not understand why we need "workaround to bypass failed machines". Why should we bypass? Is it not an error we need to investigate instead and fix in the controller so it doesn't happen?

Adding a retry in the waiting interval to make PR tests pass is not blocking the investigation and fixing the issue it just trying to save sometime using auto retry

lentzi90:
> Adding a retry in the waiting interval to make PR tests pass in less time is not blocking the investigation and fixing the issue it just trying to save sometime using auto failed vm retry instead of rerun the full test

But it could hide an actual issue! If we make the test pass by forcing a retry (by deleting the failed Machine), we are actually hiding a flaky test. User should not have to delete a Machine to get node reuse working.

scripts/ci-e2e.sh
```diff
@@ -89,7 +89,7 @@ source "${REPO_ROOT}/hack/ensure-kustomize.sh"
 
 # This will run the tests with env variabls defined in environment.sh
 # or exported by metal3-dev-env scripts
-${M3_DEV_ENV_PATH}/scripts/run_command.sh make e2e-tests
+${M3_DEV_ENV_PATH}/scripts/run_command.sh make e2e-tests || sleep infinity
```
namnx228, line 92: Why should we have sleep infinity here?

mboukhalfa: Oops ! used temporarily for debug, removed thanks

mboukhalfa:
> > Adding a retry in the waiting interval to make PR tests pass in less time is not blocking the investigation and fixing the issue it just trying to save sometime using auto failed vm retry instead of rerun the full test
> 
> But it could hide an actual issue! If we make the test pass by forcing a retry (by deleting the failed Machine), we are actually hiding a flaky test. User should not have to delete a Machine to get node reuse working.

done

test/e2e/node_reuse_test.go
```diff
@@ -509,15 +511,37 @@ func untaintNodes(clientSet *kubernetes.Clientset, nodes *corev1.NodeList, taint
...
+			err := patchNode(clientSet, nodes.Items[i].Name, &nodes.Items[i], newNode)
+			Expect(err).To(BeNil(), "Failed to patch node")
+			Logf("Node %v untainted", nodes.Items[i].Name)
 			count++
 		}
 	}
 	return
 }
 
+func patchNode(clientSet *kubernetes.Clientset, nodeName string, oldNode *corev1.Node, newNode *corev1.Node) error {
```
lentzi90, line 523: I think it would be better to change this function to something more generic, e.g. `generatePatch(oldNode *corev1.Node, newNode *corev1.Node)`. And then call `Patch` instead of `Update` with the generated patch.
As it is now. `patchNode` does something very specific (and a bit surprising I think): it generates a patch for the taints and then applies it, but it ignores all other differences there could be between the Nodes.

test/e2e/node_reuse_test.go
```diff
@@ -509,15 +511,37 @@ func untaintNodes(clientSet *kubernetes.Clientset, nodes *corev1.NodeList, taint
 		Logf("Untainting node %v ...", nodes.Items[i].Name)
 		newNode, changed := removeTaint(&nodes.Items[i], taint)
 		if changed {
-			node, err := clientSet.CoreV1().Nodes().Update(ctx, newNode, metav1.UpdateOptions{})
-			Expect(err).To(BeNil(), "Failed to update nodes")
-			Logf("Node %v untainted", node.Name)
+			err := patchNode(clientSet, nodes.Items[i].Name, &nodes.Items[i], newNode)
```
lentzi90, line 514: See my other comment for more explanations.
```suggestion
                        patchBytes := generatePatch(oldNode *corev1.Node, newNode *corev1.Node)
			err := clientSet.CoreV1().Nodes().Patch(ctx, nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
```

test/e2e/node_reuse_test.go
lentzi90: Well it could be used for patching labels or annotations also, not just taints. All I'm saying is that I would expect a function like `patchNode` to include everything, not just taints.

mboukhalfa: It is considering only taints because we are trying to avoid conflict issues like `Operation cannot be fulfilled on nodes \"test1-wjvvm\": the object has been modified; please apply your changes to the latest version and try again` https://jenkins.nordix.org/view/Metal3%20Main/job/metal3_main_v1a5_e2e_test_ubuntu/22/console 
I would prefer to rename the function to something like `patchNodeTaints`

lentzi90: Ok I'm fine with that :+1:

mboukhalfa: done

test/e2e/node_reuse_test.go
```diff
@@ -509,15 +511,37 @@ func untaintNodes(clientSet *kubernetes.Clientset, nodes *corev1.NodeList, taint
 		Logf("Untainting node %v ...", nodes.Items[i].Name)
 		newNode, changed := removeTaint(&nodes.Items[i], taint)
 		if changed {
-			node, err := clientSet.CoreV1().Nodes().Update(ctx, newNode, metav1.UpdateOptions{})
-			Expect(err).To(BeNil(), "Failed to update nodes")
-			Logf("Node %v untainted", node.Name)
+			err := patchNodeTaints(clientSet, nodes.Items[i].Name, &nodes.Items[i], newNode)
```
kashifest, line 514: why do you need to send to parameters `nodes.Items[i].Name, &nodes.Items[i]`, cant we just send the node and we can derive the name from it on the other side ?

mboukhalfa: done

lentzi90 (approved):

furkatgofurov7 (commented):
/retitle 🐛 Replace update taints with patch node taints
/approve

kashifest:
@mboukhalfa, one thing just struck my mind, can we make use of this https://github.com/kubernetes-sigs/cluster-api/blob/main/util/patch/patch.go instead of writing the code from scratch? Here's one example , https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/machine/machine_controller_noderef.go#L99

test/e2e/node_reuse_test.go
```diff
@@ -503,15 +503,15 @@ func updateBootImage(m3machineTemplateName string, clusterClient client.Client,
...
+func untaintNodes(targetClusterClient client.Client, nodes *corev1.NodeList, taint *corev1.Taint) (count int) {
 	count = 0
 	for i := range nodes.Items {
 		Logf("Untainting node %v ...", nodes.Items[i].Name)
 		newNode, changed := removeTaint(&nodes.Items[i], taint)
 		if changed {
-			node, err := clientSet.CoreV1().Nodes().Update(ctx, newNode, metav1.UpdateOptions{})
-			Expect(err).To(BeNil(), "Failed to update nodes")
-			Logf("Node %v untainted", node.Name)
+			patchHelper, err := patch.NewHelper(&nodes.Items[i], targetClusterClient)
```
kashifest, line 512: shouldn't this be newNode?

mboukhalfa:
/test-v1b1-ubuntu-e2e 
This test should fail

PR #550, opened 2022-03-03
✨ Reduce e2e test dependencies on dev-env

mboukhalfa:
This PR Get rid of (unnecessary) CAPM3 dependence on metal3-dev-env by :

1. Delegate cloning repo to dev-env and remove it from capm3 ci-e2e file
2. Set all variables needed exported explicitly in CAPM3 in ci-e2e.sh instead of sourcing files in metal3-dev-env
3. Get rid of run_command.sh (in metal3-dev-env) usage in CAPM3 e2e tests. Export necessary variables in ci-e2e.sh instead.
4. Remove sourcing of feature tests variables from hack/e2e/environment.sh in CAPM3.


test/e2e/config/e2e_conf.yaml
```diff
@@ -70,21 +70,41 @@ providers:
       targetName: "cluster-template-centos.yaml"
 
 variables:
-  KUBERNETES_VERSION: "${KUBERNETES_VERSION}"
-  UPGRADED_K8S_VERSION: "${UPGRADED_K8S_VERSION}"
```
mboukhalfa, line 74: This declaration e2e_conf.yaml is meaningless since when using the e2e framework to read a file from e2e_config file it will check first if the var exist in variable environment in that case I will not even check the file so in this file we need to have only explicit value that can be overridden by variable environment.

Check the function workflow https://github.com/kubernetes-sigs/cluster-api/blob/3433f7b769b4e7f5cb899b2742a5a8a1a9f51b3e/test/framework/clusterctl/e2e_config.go#L553

scripts/ci-e2e.sh
```diff
@@ -75,10 +76,56 @@ source "${REPO_ROOT}/hack/ensure-kubectl.sh"
...
+export IMAGE_RAW_CHECKSUM=http://$PROVISIONING_URL_HOST/images/${IMAGE_RAW_NAME}.md5sum
+
+# UPGRADE
+if [[ "${IMAGE_OS}" == "ubuntu" ]]; then
+  export UPGRADED_IMAGE_NAME="UBUNTU_20.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
+  export UPGRADED_RAW_IMAGE_NAME="UBUNTU_20.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
+else
+  export UPGRADED_IMAGE_NAME="CENTOS_8_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
+  export UPGRADED_RAW_IMAGE_NAME="CENTOS_8_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
+fi
```
mboukhalfa, line 125: Building the image vars based on different os is process implemented in devenv Should not we have a better way to reuse that instead of duplicate it here https://github.com/metal3-io/metal3-dev-env/blob/master/lib/images.sh

test/e2e/config/e2e_conf.yaml
```diff
@@ -70,21 +70,41 @@ providers:
...
+  KUBERNETES_VERSION: "v1.23.2"
+  CONTROL_PLANE_MACHINE_COUNT: 3
+  WORKER_MACHINE_COUNT: 1
   APIVersion: "infrastructure.cluster.x-k8s.io/${CAPM3_VERSION}"
   NAMEPREFIX: "${NAMEPREFIX:-baremetal-operator}"
   IRONIC_NAMESPACE: "${IRONIC_NAMESPACE:-baremetal-operator-system}"
   CONTAINER_REGISTRY: "${CONTAINER_REGISTRY:-quay.io}"
   IRONIC_IMAGE_TAG: "${IRONIC_IMAGE_TAG:-main}"
   MARIADB_IMAGE_TAG: "${MARIADB_IMAGE_TAG:-main}"
   UPGRADED_BMO_IMAGE_TAG: "${UPGRADED_BMO_IMAGE_TAG:-main}"
```
mboukhalfa, line 83: Can we use here only explicit values check this comment https://github.com/metal3-io/cluster-api-provider-metal3/pull/550/files#r819782996

lentzi90: I'm not sure exactly how this works in all situations. I guess we could just put `main` here and then the e2e framework would still take the value from the environment variable if it is set? If that's the case then yes we should just put `main` here.

mboukhalfa:
/assign @fmuyassarov
cc @Rozzii @namnx228 @furkatgofurov7

test/e2e/config/e2e_conf.yaml
mboukhalfa: I tried to limit the changes to only those affecting the dependency on devenv so I prefer we add these improvements separately

test/e2e/remediation_test.go
```diff
@@ -369,7 +369,7 @@ func listVms(state vmState) []string {
 		flag = "--state-other"
 	}
 
-	cmd := exec.Command("virsh", "list", "--name", flag)
+	cmd := exec.Command("sudo", "virsh", "list", "--name", flag)
```
lentzi90, line 372: How could this work before?

hack/e2e/environment.sh
```diff
@@ -17,62 +32,68 @@ function os_check() {
...
+export CLUSTER_URL_HOST="$CLUSTER_PROVISIONING_IP"
+
+# Vars used in the tests 
+export IRONIC_NAMESPACE=${IRONIC_NAMESPACE:-"baremetal-operator-system"}
+export NAMEPREFIX=${NAMEPREFIX:-"baremetal-operator"}
+export IRONIC_DATA_DIR="$DEV_ENV_WORKING_DIR/ironic"
+export BMOPATH="${BMOPATH:-${M3PATH}/baremetal-operator}"
+export IRONIC_TLS_SETUP=${IRONIC_TLS_SETUP:-"true"}
+export IRONIC_BASIC_AUTH=${IRONIC_BASIC_AUTH:-"true"}
+export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-${IRONIC_TLS_SETUP}}
```
lentzi90, line 99: Could we not set these in the go code or config file instead? Actually some of them are already in the e2e config.

hack/e2e/releases.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+function get_latest_release() {
+  set +x
+  if [ -z "${GITHUB_TOKEN:-}" ]; then
```
lentzi90, line 5: I'm not sure if the `GITHUB_TOKEN` will actually be set when we do this in CAPM3. Worth checking!

mboukhalfa: I think this var always exported from project infra https://github.com/metal3-io/project-infra/blob/9470543342be88a38a94e120b81ee04c90a9576b/jenkins/scripts/files/run_integration_tests.sh#L6

hack/e2e/environment.sh
mboukhalfa: moved some to e2e config and others still needed in the var environment because they are used in other scrips

hack/e2e/environment.sh
```diff
@@ -17,62 +32,64 @@ function os_check() {
...
-source "${M3_DEV_ENV_PATH}/scripts/feature_tests/feature_test_vars.sh"
-# shellcheck disable=SC1091
-# shellcheck disable=SC1090
-source "${M3_DEV_ENV_PATH}/scripts/feature_tests/node_reuse/node_reuse_vars.sh"
 
-# Pin Calico version
-export CALICO_PATCH_RELEASE="v3.21.0"
-export CALICO_MINOR_RELEASE="v3.21"
+# Exported to the cluster templates
+SSH_PUB_KEY_CONTENT=$(cat "$HOME/.ssh/id_rsa.pub")
```
mboukhalfa, line 83: Fails in the file does not exist when testing on a clean machine

mboukhalfa: fixed

Rozzii (approved):

test/e2e/remediation_test.go
lentzi90: Makes sense! No need to investigate more.

scripts/ci-e2e.sh
```diff
@@ -81,10 +49,19 @@ source "${REPO_ROOT}/hack/ensure-kubectl.sh"
...
-# or exported by metal3-dev-env scripts
-${M3_DEV_ENV_PATH}/scripts/run_command.sh make e2e-tests
+# shellcheck source=./hack/e2e/images.sh
+source "${REPO_ROOT}/hack/e2e/images.sh"
+# shellcheck source=./hack/e2e/releases.sh
+source "${REPO_ROOT}/hack/e2e/releases.sh"
+# shellcheck source=./hack/e2e/ironic_basic_auth.sh
+source "${REPO_ROOT}/hack/e2e/ironic_basic_auth.sh"
+# shellcheck source=./hack/e2e/ironic_tls_setup.sh
+source "${REPO_ROOT}/hack/e2e/ironic_tls_setup.sh"
```
namnx228, line 59: Because cloning metal3-dev-env is mandatory for now, these scripts should be sourced from the repo. Keeping track of two copies of the same scripts in two different repos makes it challenging to maintain e2e tests.

namnx228 (approved):
/lgtm

Rozzii (commented):
LGTM

mboukhalfa:
> There is still some failing test but the concept LGTM

Failing because it still waiting for new release image

PR #551, opened 2022-03-03
🌱 Uplift containerd CRI plugin

furkatgofurov7:
Fixes https://github.com/containerd/containerd/security/advisories/GHSA-crp2-qrr5-8pq7
As a result of this other packages have been also auto-updated.


kashifest:
/approve
/lgtm
Taking this in for security reasons.

PR #561, opened 2022-03-10
🐛 Fix mockgen path in tools.go

honza:
This has to be consistent with the build command in the Makefile in
order for vendoring to work properly (builds with `-mod=vendor` fail without this fix because the correct files aren't vendored).

### DIFF
--- a/hack/tools/tools.go
+++ b/hack/tools/tools.go
@@ -22,7 +22,7 @@ package tools
 
 import (
 	_ "github.com/drone/envsubst"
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (approved):
Thanks for the fix @honza! This indeed is not in sync with Makefile and was failing when running locally. 
/retitle 🐛 Fix mockgen path in tools.go
/approve
/cc @kashifest @fmuyassarov @namnx228

PR #563, opened 2022-03-11
✨ Introduce an additional providerID format and set providerID from nodes when noCloudProvider is set to false

Xenwar:
This PR has three main goals.

**(1)providerID from nodes when noCloudProvider is set to false**

With `Metal3Cluster.Spec.NoCloudProvider` set to `false`.  The source of truth for the `providerID` is the node.
And, the `providerID` field on the `Metal3Machine` should be updated accordingly.

**(2) A new providerID format**

This PR also introduced a new providerID format. The are now two supported providerID formats.
Suppose, we have a bmh named node-0, then the two providerIDs formats are shown below.
```
metal3://d668eb95-5df6-4c10-a01a-fc69f4299fc6
metal3://metal3/node-0/test-controlplane-xyz
```

**(3) preventing  duplicate usage of a providerID**
This PR prevents a providerID from being used by multiple nodes at the same time. To this end, both formats of a given providerID are treated as one and should not be used simultaneously.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -91,7 +91,7 @@ type MachineManagerInterface interface {
 	Update(context.Context) error
 	HasAnnotation() bool
 	GetProviderIDAndBMHID() (string, *string)
-	SetNodeProviderID(context.Context, string, string, ClientGetter) error
+	SetNodeProviderID(context.Context, string, *string, ClientGetter) error
 	SetProviderID(string)
 	SetPauseAnnotation(context.Context) error
 	RemovePauseAnnotation(context.Context) error
@@ -1317,40 +1317,76 @@ func (m *MachineManager) GetProviderIDAndBMHID() (string, *string) {
 type ClientGetter func(ctx context.Context, c client.Client, cluster *clusterv1.Cluster) (clientcorev1.CoreV1Interface, error)
 
 // SetNodeProviderID sets the metal3 provider ID on the kubernetes node.
-func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerID string, clientFactory ClientGetter) error {
-	if !m.Metal3Cluster.Spec.NoCloudProvider {
-		return nil
-	}
+func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID string, providerIDOnM3M *string, clientFactory ClientGetter) error {
+	// todo: bmhID should not be trusted and needs to removed from the signature.
 	corev1Remote, err := clientFactory(ctx, m.client, m.Cluster)
 	if err != nil {
 		return errors.Wrap(err, "Error creating a remote client")
 	}
+	namespace := m.Metal3Machine.GetNamespace()
+	m3mName := m.Metal3Machine.GetName()
+	bmhName, err := m.getBmhNameFromM3Machine()
+	if err != nil {
+		m.Log.Info("unable to retrieve BMH name from Metal3Machine")
+		return &RequeueAfterError{RequeueAfter: requeueAfter}
+	}
 
-	nodeLabel := fmt.Sprintf("%s=%v", ProviderLabelPrefix, bmhID)
-	nodes, nodesCount, err := m.getNodesWithLabel(ctx, nodeLabel, clientFactory)
+	providerIDLegacy := fmt.Sprintf("metal3://%s", bmhID)
+	providerIDNew := fmt.Sprintf("metal3://%s/%s/%s", namespace, bmhName, m3mName)
+
+	if !m.Metal3Cluster.Spec.NoCloudProvider {
+		matchingNodesCount, err := m.getMatchingNodesWithoutLabelCount(ctx, providerIDLegacy, providerIDNew, providerIDOnM3M, clientFactory)
+		if matchingNodesCount > 1 {
+			m.Log.Info("More than one node using the same providerID")
+			return errors.Wrap(err, "More than one node using the same providerID")
+		}
+		if err != nil {
+			m.Log.Info("error retrieving node, requeuing")
+			return &RequeueAfterError{RequeueAfter: requeueAfter}
+		}
+		if matchingNodesCount == 0 {
+			// The node could either be still running cloud-init or
+			// kubernetes has not set the node.spec.ProviderID field yet.
+			m.Log.Info("Some target nodes do not have spec.providerID field set yet, requeuing")
+			return &RequeueAfterError{RequeueAfter: requeueAfter}
+		}
+		return nil
+	}
+	nodeLabel := fmt.Sprintf("%s=%s", ProviderLabelPrefix, bmhID)
+	nodes, countNodesWithLabel, err := m.getNodesWithLabel(ctx, nodeLabel, clientFactory)
 	if err != nil {
 		m.Log.Info("error retrieving node, requeuing")
 		return &RequeueAfterError{RequeueAfter: requeueAfter}
 	}
-	if nodesCount == 0 {
+	if countNodesWithLabel == 0 {
 		// The node could either be still running cloud-init or have been
 		// deleted manually. TODO: handle a manual deletion case.
-		m.Log.Info("Target node is not found, requeuing")
+		m.Log.Info("requeuing, could not find node with label: ", nodeLabel)
 		return &RequeueAfterError{RequeueAfter: requeueAfter}
 	}
+	if countNodesWithLabel > 1 {
+		return errors.Wrap(err, fmt.Sprintf("Found multiple target nodes with the same label: (%s)", nodeLabel))
+	}
+	var nodeVar corev1.Node
 	for _, node := range nodes.Items {
-		node := node
-		if node.Spec.ProviderID == providerID {
-			continue
+		providerIDOnNode := node.Spec.ProviderID
+		if providerIDOnNode == "" {
+			node.Spec.ProviderID = providerIDNew
+		} else if providerIDOnNode == providerIDNew {
+			*providerIDOnM3M = providerIDNew
+		} else if providerIDOnNode == providerIDLegacy {
+			*providerIDOnM3M = providerIDLegacy
+		} else {
+			m.Log.Info("node using unsupported providerID format", providerIDOnNode)
+			return errors.Wrap(err, "node using unsupported providerID format")
 		}
-		node.Spec.ProviderID = providerID
-		_, err = corev1Remote.Nodes().Update(ctx, &node, metav1.UpdateOptions{})
+		nodeVar = node
+		_, err = corev1Remote.Nodes().Update(ctx, &nodeVar, metav1.UpdateOptions{})
 		if err != nil {
-			return errors.Wrap(err, "unable to update the target node")
+			return errors.Wrap(err, "unable to update the target node with providerID")
 		}
 	}
 	m.Log.Info("ProviderID set on target node")
-
 	return nil
 }
 
@@ -1722,6 +1758,18 @@ func (m *MachineManager) getMachineSet(ctx context.Context) (*clusterv1.MachineS
 	return nil, errors.New("MachineSet is not found")
 }
 
+// getBmhNameFromM3Machine retrieves bmhName from m3m annotations.
+func (m *MachineManager) getBmhNameFromM3Machine() (string, error) {
+	annotationValue := m.Metal3Machine.ObjectMeta.GetAnnotations()[HostAnnotation]
+	valueParts := strings.Split(annotationValue, "/")
+	if (len(valueParts) < 2) || (valueParts[0] != m.Metal3Machine.GetNamespace()) {
+		errMessage := fmt.Sprintf("unable to retrieve bmh name from metal3machine: %s , using annotation: %s", m.Metal3Machine.GetName(), annotationValue)
+		return "", errors.New(errMessage)
+	}
+	bmhName := valueParts[1]
+	return bmhName, nil
+}
+
 // getNodesWithLabel gets kubernetes nodes with a given label.
 func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string, clientFactory ClientGetter) (*corev1.NodeList, int, error) {
 	corev1Remote, err := clientFactory(ctx, m.client, m.Cluster)
@@ -1729,9 +1777,11 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
 		return nil, 0, errors.Wrap(err, "Error creating a remote client")
 	}
 	nodesCount := 0
-	nodes, err := corev1Remote.Nodes().List(ctx, metav1.ListOptions{
+	filter := metav1.ListOptions{
 		LabelSelector: nodeLabel,
-	})
+	}
+
+	nodes, err := corev1Remote.Nodes().List(ctx, filter)
 	if err != nil {
 		m.Log.Info(fmt.Sprintf("error while retrieving nodes with label (%s): %v", nodeLabel, err))
 		return nil, 0, err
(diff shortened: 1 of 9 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -1729,9 +1748,14 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
 	nodesCount := 0
-	nodes, err := corev1Remote.Nodes().List(ctx, metav1.ListOptions{
+	filter := metav1.ListOptions{
 		LabelSelector: nodeLabel,
-	})
+	}
+	// When no labels are in use, retrieve all nodes
+	if !m.Metal3Cluster.Spec.NoCloudProvider {
+		filter = metav1.ListOptions{}
+	}
```
kashifest, line 1757: Lines 1751-1757 isn't clear to me, is it so that if NoCloudProvier is false, we return all the nodes ? and Why is that ?

Xenwar: Yes. 
If NoCloudProvier=false, we get all nodes. The nodes could have labels or not, it should not matter. 
However, 
If NoCloudProvier=true, then we just get the nodes with labels.

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
...
+	providerIDLegacy := *providerID
+	providerIDNew := fmt.Sprintf("metal3://%s/%s/%s", namespace, bmhName, m3mName)
+	if !m.Metal3Cluster.Spec.NoCloudProvider {
+		// Settig providerID on nodes takes time. Continuous monitoring is required.
+		nodesWithProviderIDset := 0
+		for _, node := range nodes.Items {
+			m.Log.Info("PP0: ", node.Spec.ProviderID)
+			if node.Spec.ProviderID == "" {
+				m.Log.Info("PP1: NoCloudProvider set to false. But, no providerID is set on the node either", node.GetName())
+				nodesWithProviderIDset++
```
maelk, line 1354: no need to count, since you use exit in the other cases. if you reach line 1366, it means you've gone through all nodes. If you still want to count the nodes, then you should also count in an else condition

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
...
+			} else if node.Spec.ProviderID == providerIDNew {
+				m.Log.Info("PP2: Found a node with matching node providerID in new format")
+				*providerID = providerIDNew
+				return nil
+			} else if node.Spec.ProviderID == providerIDLegacy {
+				m.Log.Info("PP3: Found a node with matching node providerID in legacy format")
+				*providerID = providerIDLegacy
+				return nil
+			}
```
maelk, line 1363: missing a case here, where the PID is set but to something different (all other provisioned nodes)

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
...
+		}
+		// We have considered all nodes, and no matching node was found.
+		if nodesWithProviderIDset == nodesCount {
```
maelk, line 1366: you don't need this condition

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
...
+	namespace := m.Metal3Machine.GetNamespace()
+	m3mName := m.Metal3Machine.GetName()
+	bmhName, err := m.getBmhNameFromM3Machine()
+	if err != nil {
+		m.Log.Info("unable to retrieve BMH from Metal3Cluster")
+		return &RequeueAfterError{RequeueAfter: requeueAfter}
+	}
+	providerIDLegacy := *providerID
+	providerIDNew := fmt.Sprintf("metal3://%s/%s/%s", namespace, bmhName, m3mName)
+	if !m.Metal3Cluster.Spec.NoCloudProvider {
```
maelk, line 1347: you could separate the content of this if into another function. It would be more readable

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
 		m.Log.Info("Target node is not found, requeuing")
 		return &RequeueAfterError{RequeueAfter: requeueAfter}
 	}
```
maelk, line 1337: indentation issue above here

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
...
+		}
+		// We have considered all nodes, and no matching node was found.
+		if nodesWithProviderIDset == nodesCount {
+			return errors.New("No matching node with the providerID found: ")
+		}
+		m.Log.Info("Waiting for more nodes to have their providerID property set before declaring failure")
+		return &RequeueAfterError{RequeueAfter: requeueAfter}
+	}
+	var nodeVar corev1.Node
 	for _, node := range nodes.Items {
```
maelk, line 1373: at this point you should check there is only one node (it should have used the label matching). If not, error out

baremetal/metal3machine_manager.go
```diff
@@ -1338,15 +1335,58 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerI
...
-		node := node
-		if node.Spec.ProviderID == providerID {
-			continue
+		if node.Spec.ProviderID == "" {
+			m.Log.Info("missing node providerID, setting value from label")
+			node.Spec.ProviderID = providerIDLegacy
```
maelk, line 1376: here we could set the new format, no ?

baremetal/metal3machine_manager.go
Xenwar: That was my assumption as well. However, even though all nodes are retrieved, all nodes do not have their providerID set. 
So, the reason I am counting is to make sure that the decision is based on all nodes(not just the nodes that have providerID)

baremetal/metal3machine_manager.go
Xenwar: thanks, done

baremetal/metal3machine_manager.go
Xenwar: I think we do as explained above.

baremetal/metal3machine_manager.go
Xenwar: thanks, will do that.

baremetal/metal3machine_manager.go
Xenwar: A bit unclear. can you elaborate ?

baremetal/metal3machine_manager.go
Xenwar: I think that is implicit in the logic, since it will return the moment it finds it.

Xenwar: I have thought about it as well, but I was assuming the user is responsible for providing uniq nodes. But, if we are to verify as well, I should probably move the logic to a different function so that uniqueness is checked.

baremetal/metal3machine_manager.go
Xenwar: I was not sure on that, but I assumed this:
There are two sources or providerID (node level) and (code level). If the former is missing, the latter is used. 
But, I can change it to the new format as well.

baremetal/metal3machine_manager.go
maelk: it does not matter, whether they have the providerID set or not, as long as you don't return during the loop, it means that you didn't find what you were looking for

baremetal/metal3machine_manager.go
maelk: if you expand the block just above, there is some indentation issues linked to the change apparently

baremetal/metal3machine_manager.go
maelk: There's no need to make it a whole thing, just a small length check is enough

baremetal/metal3machine_manager.go
maelk: here, the providerID is unset, and you identified the node based on the label. It is really up to you what providerID you set there (as long as it is the same as on the m3m) . Since we introduce a new format, let's use it by default.

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1787,36 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+			*providerIDonM3M = providerIDNew
+			matchingNodesCount = 1
+			return matchingNodesCount, nil
+		} else if providerIDOnNode == providerIDLegacy {
+			*providerIDonM3M = providerIDLegacy
+			matchingNodesCount = 1
+			return matchingNodesCount, nil
+		} else {
+			m.Log.Info("The node: ", node.GetName(), " does not match expected providerID: ", providerIDOnNode, "Considering other nodes ...")
+		}
```
maelk, line 1819: This logic is correct, but not fully resilient to the case where we would have a duplicated PID. This PR is supposed to fix this issue, but we should probably ensure that if two nodes match either the new or old providerID, we stop there and error out. This is needed for the transition phase

Xenwar: Identifying the duplicates can be done. But, what the corrective measure in this case ? Try again ?

Xenwar:
/test-v1b1-centos-e2e runs v1b1 e2e tests on CentOS
/test-v1b1-upgrade-e2e

kashifest:
/approve
I will really appreciate a documentation on this new provider id and the introduced change in functionalities later on.

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1791,83 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+	err = m.duplicateProviderIDsExist(validNodes, providerIDLegacy, providerIDNew)
+	if err != nil {
+		// There are, at least, two nodes. Details are in err
+		return 2, err
+	}
+	if matchingNodeProviderID != "" {
+		*providerIDonM3M = matchingNodeProviderID
+		return 1, nil
+	}
+	return matchingNodesCount, nil
```
namnx228, line 1836: Is this matchingNodesCount variable always 0? Because there is no assignment and modification other than its initialization?

namnx228 (commented):
Great work. Left some comments.

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1791,83 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
 	}
 	return nodes, nodesCount, err
 }
+
+// getNodesWithoutLabel gets kubernetes nodes based on their Spec.providerID field.
+func (m *MachineManager) getNodesWithoutLabel(ctx context.Context, providerIDLegacy, providerIDNew string, providerIDonM3M *string, clientFactory ClientGetter) (int, error) {
```
furkatgofurov7, line 1796: should we rename this into something else, since we are not getting nodes without labels, but more of a counting matching nodes? maybe to `getMatchingNodesCount` or something similar

baremetal/metal3machine_manager.go
```diff
@@ -1317,40 +1317,76 @@ func (m *MachineManager) GetProviderIDAndBMHID() (string, *string) {
...
+		return errors.Wrap(err, fmt.Sprintf("Found multiple target nodes with the same label: (%s)", nodeLabel))
+	}
+	var nodeVar corev1.Node
 	for _, node := range nodes.Items {
-		node := node
-		if node.Spec.ProviderID == providerID {
-			continue
+		providerIDOnNode := node.Spec.ProviderID
+		if providerIDOnNode == "" {
+			node.Spec.ProviderID = providerIDNew
```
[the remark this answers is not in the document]
Xenwar: In this case, we have found the node using label, but it does not have a providerID. so,
(if we are using labels) AND (node has no providerID) THEN the m3m becomes the source of truth.

baremetal/metal3machine_manager.go
Xenwar: Yes, it is always zero.
There are multiple locations that should return 0. It is just paramerization.

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1791,83 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+		m.Log.Info(fmt.Sprintf("error while retrieving nodes: %v", err))
+		return matchingNodesCount, err
+	}
+	// kubernetes nodes with their providerID and name fields set
+	validNodes := make(map[string][]string)
+	matchingNodeProviderID := ""
+	for _, node := range nodes.Items {
+		providerIDOnNode := node.Spec.ProviderID
+		if node.Spec.ProviderID == "" {
+			m.Log.Info("no providerID value found on node: ", node.GetName())
```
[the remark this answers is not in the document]
Xenwar: That is handled on line 1823:
```
		if providerIDOnNode != "" && node.GetName() != "" {
			validNodes[providerIDOnNode] = append(validNodes[providerIDOnNode], node.GetName())
		}
```

baremetal/metal3machine_manager.go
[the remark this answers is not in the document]
namnx228: So in this case, m3m has already had the new providerID, is that true? But `providerIDNew` is just created in this function, so how M3m can have this providerID?

baremetal/metal3machine_manager.go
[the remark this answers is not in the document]
namnx228: I mean, if the node has no name, the log will be "no providerID value found on node: " without any names. Should we have a new log for this case?

baremetal/metal3machine_manager.go
namnx228: I think if it is always zero, we should not use a variable, but directly return 0. Variable means for something that can be changed.

baremetal/metal3machine_manager.go
```diff
@@ -1317,40 +1317,76 @@ func (m *MachineManager) GetProviderIDAndBMHID() (string, *string) {
 type ClientGetter func(ctx context.Context, c client.Client, cluster *clusterv1.Cluster) (clientcorev1.CoreV1Interface, error)
 
 // SetNodeProviderID sets the metal3 provider ID on the kubernetes node.
-func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID, providerID string, clientFactory ClientGetter) error {
-	if !m.Metal3Cluster.Spec.NoCloudProvider {
-		return nil
-	}
+func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID string, providerIDOnM3M *string, clientFactory ClientGetter) error {
+	// todo: bmhID should not be trusted and needs to removed from the signature.
```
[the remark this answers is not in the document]
Xenwar: This is for future work. 
Just to make clear, the problem is on https://github.com/metal3-io/cluster-api-provider-metal3/blob/b1acf226c5382081079c906219c16da0fc16c0b7/baremetal/metal3machine_manager.go#L1304-L1314

The providerID is taken from Metal3Machine. Then the bmh is extracted from it. This is wrong. 
What should happen in either of these ways.

(1) Get all BMHs and filter by the related(owner Ref) M3M
(2) Get the BMH from that M3M using annotation

That way we make the m3m resource the source of truth for the BMH-id

baremetal/metal3machine_manager_test.go
```diff
@@ -2285,9 +2305,11 @@ var _ = Describe("Metal3Machine manager", func() {
...
 					},
 				},
-				HostID:             string(bmhuid),
-				ExpectedError:      false,
-				ExpectedProviderID: fmt.Sprintf("%s%s", ProviderIDPrefix, string(bmhuid)),
+				HostID:               string(bmhuid),
+				ExpectedError:        false,
+				ExpectedProviderID:   fmt.Sprintf("metal3://%s/%s/%s", namespaceName, "myhost", "abc"),
+				M3MHasHostAnnotation: true,
+				// here
```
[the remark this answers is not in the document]
Xenwar: It is separated.
`matchingNodesCount, err := m.getNodesWithoutLabel`
Here, we are just evaluating the returned result.

controllers/metal3machine_controller_integration_test.go
```diff
@@ -579,6 +579,16 @@ var _ = Describe("Reconcile metal3machine", func() {
 					newMetal3Cluster(metal3ClusterName, nil, nil, nil, nil, false),
 					newBareMetalHost(nil, nil, nil, false),
 				},
+				TargetObjects: []runtime.Object{
+					&corev1.Node{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "bmh-0",
```
lentzi90, line 585: I'm a bit confused by the name here. This is a Node, right? Is the name to indicate that it is using the BMH with the same name?

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1791,83 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+	if err != nil {
+		m.Log.Info(fmt.Sprintf("error while retrieving nodes: %v", err))
+		return matchingNodesCount, err
+	}
+	// kubernetes nodes with their providerID and name fields set
+	validNodes := make(map[string][]string)
+	matchingNodeProviderID := ""
+	for _, node := range nodes.Items {
+		providerIDOnNode := node.Spec.ProviderID
+		if node.Spec.ProviderID == "" {
```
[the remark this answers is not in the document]
Xenwar: True, thanks.

baremetal/metal3machine_manager.go
```diff
@@ -1317,40 +1317,76 @@ func (m *MachineManager) GetProviderIDAndBMHID() (string, *string) {
...
+		matchingNodesCount, err := m.getNodesWithoutLabel(ctx, providerIDLegacy, providerIDNew, providerIDOnM3M, clientFactory)
+		if matchingNodesCount > 1 {
+			m.Log.Info("More than one node using the same providerID")
+			return errors.Wrap(err, "More than one node using the same providerID")
+		}
+		if err != nil {
+			m.Log.Info("error retrieving node, requeuing")
+			return &RequeueAfterError{RequeueAfter: requeueAfter}
+		}
+		if matchingNodesCount == 0 {
```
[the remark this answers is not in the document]
Xenwar: Yes. 
We can only make relevant decision after retrieving and evaluating the providerIDs. Any other error will result in a retry.

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1791,83 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+		// duplicates due to the same, at least, two instances of legacy OR new OR unknown formats being used in multiple nodes.
+		if len(nodes) > 1 {
+			errMsg := fmt.Sprintf("providerID %s is in use by multiple nodes: (%s)", providerID, matchingNodes)
+			m.Log.Info(errMsg)
+			return errors.New(errMsg)
+		}
+	}
+	// duplicates due to both the legacy AND new providerIDs being used by multiple nodes.
+	if duplicateUsageCounter > 1 {
+		errMsg := fmt.Sprintf("both providerIDs (%s and %s) cannot be used at the same time by node: (%s)", providerIDLegacy, providerIDNew, duplicateNodes)
```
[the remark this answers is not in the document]
Xenwar: Should be duplicateNodes as it is an array of nodes that are duplicates.

baremetal/metal3machine_manager.go
```diff
@@ -1741,3 +1791,83 @@ func (m *MachineManager) getNodesWithLabel(ctx context.Context, nodeLabel string
...
+			m.Log.Info("The node: ", node.GetName(), " does not match expected providerID: ", providerIDOnNode, "Considering other nodes ...")
+		}
+		if providerIDOnNode != "" && node.GetName() != "" {
+			validNodes[providerIDOnNode] = append(validNodes[providerIDOnNode], node.GetName())
+		}
+	}
+	err = m.duplicateProviderIDsExist(validNodes, providerIDLegacy, providerIDNew)
+	if err != nil {
+		// There are, at least, two nodes. Details are in err
+		return 2, err
```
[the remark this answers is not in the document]
Xenwar: The threshold is 2. 
`duplicateProviderIDsExist` is returning the actual nodes in the `err`.
We can make `duplicateProviderIDsExist` return the number of nodes, but that information is not relevant with respect to the caller. So, `at least` is the threshold.

baremetal/metal3machine_manager.go
Xenwar: I agree. 
I was trying to emphasise  that fact that labels are not in user. I should probably rename it and add comments about the label.

Xenwar:
/hold
Addressing the comments

baremetal/metal3machine_manager.go
lentzi90: I agree with @namnx228 , just returning 0 would be less confusing. If we still want to have a descriptive name let's make it a constant and name it something like `zeroMatchingNodes`.

baremetal/metal3machine_manager.go
Xenwar: Changed it to `getMatchingNodesWithoutLabelCount`

controllers/metal3machine_controller_integration_test.go
Xenwar: Thanks. 
changed it to node-0, node-1 ...

baremetal/suite_test.go
```diff
@@ -55,7 +55,7 @@ const (
 	bmhuid            = "22222222-9845-4c48-9e49-c74be387f57c"
 )
 
-var providerid = fmt.Sprintf("%s/%s/%s", namespaceName, bmhuid, m3muid)
+var providerid = fmt.Sprintf("%s/%s/%s", namespaceName, "abc", "abc")
```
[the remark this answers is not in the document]
Xenwar: I agree with another PR. 
our test cases are full of such hard coded values.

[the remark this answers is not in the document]
Xenwar: Tried that with this PR as well, but was affecting many other tests that are unrelated to this one.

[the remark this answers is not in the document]
lentzi90: Thanks, let's leave it to another PR then :slightly_smiling_face:

Xenwar:
/test-v1b1-centos-e2e runs v1b1 e2e tests on CentOS
/test-v1b1-upgrade-e2e
/test-v1b1-ubuntu-integration

baremetal/metal3machine_manager_test.go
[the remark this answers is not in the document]
furkatgofurov7: I was pointing to the comment

baremetal/metal3machine_manager.go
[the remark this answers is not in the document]
furkatgofurov7: Okay, thanks

baremetal/metal3machine_manager.go
Xenwar: Here the results are evaluated. The actual content has moved to the end of the file

kashifest:
test-v1b1-ubuntu-e2e

furkatgofurov7 (commented):
/retitle ✨ Introduce an additional providerID format and set providerID from nodes when noCloudProvider is set to false
/lgtm

PR #566, opened 2022-03-11
:bug: Fix recursion in deleteOwnerRefFromList

### LINKED ISSUE #565: Wrong parameter for recursion in deleteOwnerRefFromList
If you use a refList as input where one owner ref is deleted, then the recursion is called. In this recursion, the last two elements of the array are being cut off.

In the recursion we should only cut off one element, the one that is supposed to be deleted.

/kind bug

janiskemper:
In the recursive call of deleteOwnerRefFromList there is used the
variable refListLen := len(refList) - 1. However, in the recursive call
we use refListLen - 1, so that we have len(refList) - 2 in the end. It
should only be len(refList) - 1 though.

Fixes #565

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -1394,7 +1394,7 @@ func deleteOwnerRefFromList(refList []metav1.OwnerReference,
 	}
 	refListLen := len(refList) - 1
 	refList[index] = refList[refListLen]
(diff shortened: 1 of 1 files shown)


Rozzii:
Could you please help?
/cc @Xenwar

mboukhalfa:
/lgtm
Thanks @janiskemper for the fix! 
Would be great to write an issue to add unit test cases covering this

PR #568, opened 2022-03-15
🌱Add e2e tests for ephemeral cluster

smoshiur1237:
This PR will introduce e2e test for ephemeral cluster such as ironic upgrade, bmo upgrade, cert rotation and node reuse tests.

### DIFF
--- a/test/e2e/cert_rotation_test.go
+++ b/test/e2e/cert_rotation_test.go
@@ -12,22 +12,20 @@ import (
 	appv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"k8s.io/apimachinery/pkg/labels"
-	framework "sigs.k8s.io/cluster-api/test/framework"
 )
 
-func certRotation() {
+func certRotation(clientSet *kubernetes.Clientset, clusterClient client.Client) {
 	Logf("Start the certificate rotation test")
 	By("Check if Ironic pod is running")
-	targetClusterClientSet := targetCluster.GetClientSet()
 	ironicNamespace := os.Getenv("NAMEPREFIX") + "-system"
 	ironicDeploymentName := os.Getenv("NAMEPREFIX") + "-ironic"
-	ironicDeployment, err := getDeployment(targetCluster, ironicDeploymentName, ironicNamespace)
+	ironicDeployment, err := getDeployment(clusterClient, ironicDeploymentName, ironicNamespace)
 	Eventually(func() error {
-		ironicPod, err := getPodFromDeployment(targetCluster, ironicDeployment, ironicNamespace)
+		ironicPod, err := getPodFromDeployment(clientSet, ironicDeployment, ironicNamespace)
 		if err != nil {
 			return err
 		}
@@ -42,11 +40,10 @@ func certRotation() {
 
 	By("Get the current number of time containers were restarted")
 	containerNumRestart := make(map[string]int32)
-	containerNumRestart["ironic-api"] = 0
-	containerNumRestart["ironic-conductor"] = 0
+	containerNumRestart["ironic-httpd"] = 0
 	containerNumRestart["mariadb"] = 0
 	Expect(err).To(BeNil())
-	ironicPod, err := getPodFromDeployment(targetCluster, ironicDeployment, ironicNamespace)
+	ironicPod, err := getPodFromDeployment(clientSet, ironicDeployment, ironicNamespace)
 	Expect(err).To(BeNil())
 	for _, container := range ironicPod.Status.ContainerStatuses {
 		if _, exist := containerNumRestart[container.Name]; exist {
@@ -61,17 +58,29 @@ func certRotation() {
 		"mariadb-cert",
 	}
 	for _, secretName := range secretList {
-		err := targetClusterClientSet.CoreV1().Secrets(ironicNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
+		err := clientSet.CoreV1().Secrets(ironicNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
 		Expect(err).To(BeNil(), "Cannot detele this secret: %s", secretName)
 	}
 
 	By("Wait for containers in the ironic pod to be restarted")
 	Eventually(func() error {
-		ironicPod, err := getPodFromDeployment(targetCluster, ironicDeployment, ironicNamespace)
+		ironicPod, err := getPodFromDeployment(clientSet, ironicDeployment, ironicNamespace)
 		if err != nil {
 			return err
 		}
-
+		// check for container in containerNumRestart list
+		for container := range containerNumRestart {
+			notFound := true
+			for _, ironicContainer := range ironicPod.Status.ContainerStatuses {
+				if ironicContainer.Name == container {
+					notFound = false
+					break
+				}
+			}
+			if notFound {
+				return fmt.Errorf("%s container does not exist in Ironic pod", container)
+			}
+		}
 		if ironicPod.Status.Phase == corev1.PodRunning {
 			for _, container := range ironicPod.Status.ContainerStatuses {
 				if oldNumRestart, exist := containerNumRestart[container.Name]; exist {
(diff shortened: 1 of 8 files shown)


test/e2e/cert_rotation_test.go
```diff
@@ -42,11 +40,10 @@ func certRotation() {
 
 	By("Get the current number of time containers were restarted")
 	containerNumRestart := make(map[string]int32)
-	containerNumRestart["ironic-api"] = 0
-	containerNumRestart["ironic-conductor"] = 0
+	containerNumRestart["ironic-httpd"] = 0
 	containerNumRestart["mariadb"] = 0
 	Expect(err).To(BeNil())
-	ironicPod, err := getPodFromDeployment(targetCluster, ironicDeployment, ironicNamespace)
+	ironicPod, err := getPodFromDeployment(clientSet, ironicDeployment, ironicNamespace)
```
namnx228, line 46:
```suggestion
	ironicPod, err := getPodFromDeployment(clusterClient, ironicDeployment, ironicNamespace)
```

test/e2e/cert_rotation_test.go
```diff
@@ -61,17 +58,28 @@ func certRotation() {
...
 	for _, secretName := range secretList {
-		err := targetClusterClientSet.CoreV1().Secrets(ironicNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
+		err := clientSet.CoreV1().Secrets(ironicNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
 		Expect(err).To(BeNil(), "Cannot detele this secret: %s", secretName)
 	}
 
 	By("Wait for containers in the ironic pod to be restarted")
 	Eventually(func() error {
-		ironicPod, err := getPodFromDeployment(targetCluster, ironicDeployment, ironicNamespace)
+		ironicPod, err := getPodFromDeployment(clientSet, ironicDeployment, ironicNamespace)
```
namnx228, line 67: The same here

test/e2e/cert_rotation_test.go
```diff
@@ -61,17 +58,28 @@ func certRotation() {
...
+		for container := range containerNumRestart {
+			notFound := true
+			for _, ironicContainer := range ironicPod.Status.ContainerStatuses {
+				if ironicContainer.Name == container {
+					notFound = false
+					break
+				}
+			}
+			if notFound {
+				return fmt.Errorf("%s Ironic pod does not exits", container)
```
namnx228, line 80:
```suggestion
				return fmt.Errorf("%s container does not exist in Ironic pod", container)
```

namnx228 (commented):
Good work! I left some comments. Please trigger the e2e test after fixing these comments. Thanks

test/e2e/cert_rotation_test.go
smoshiur1237: `clientSet` is the correct one.

test/e2e/cert_rotation_test.go
smoshiur1237: `clientSet` should be right.

test/e2e/cert_rotation_test.go
smoshiur1237: Done

test/e2e/cert_rotation_test.go
```diff
@@ -61,17 +58,28 @@ func certRotation() {
...
+			for _, ironicContainer := range ironicPod.Status.ContainerStatuses {
+				if ironicContainer.Name == container {
+					notFound = false
+					break
+				}
+			}
+			if notFound {
+				return fmt.Errorf("%s container does not exist in Ironic pod", container)
+			}
+		}
```
mboukhalfa, line 82: I do not think we need this since already included in the next loop we can check `exist` than return `error` if `false`

test/e2e/cert_rotation_test.go
```diff
@@ -42,11 +40,10 @@ func certRotation() {
 
 	By("Get the current number of time containers were restarted")
 	containerNumRestart := make(map[string]int32)
-	containerNumRestart["ironic-api"] = 0
-	containerNumRestart["ironic-conductor"] = 0
+	containerNumRestart["ironic-httpd"] = 0
```
mboukhalfa, line 43: should not we add ironic container also here ?

hack/e2e/environment.sh
```diff
@@ -22,6 +22,7 @@ export CAPI_VERSION=${CAPI_VERSION:-"v1beta1"}
 export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
 export NUM_NODES=${NUM_NODES:-"4"}
 export UPGRADE_TEST=${UPGRADE_TEST:-false}
+export EPHEMERAL_TEST=${EPHEMERAL_TEST:-false}
```
lentzi90, line 25: I think we don't need to have the variable in this file. If I recall correctly, the only reason we have `UPGRADE_TEST` here is because we set CAPI/CAPM3 versions depending on if it is an upgrade test or not. But the file should basically only be for metal3-dev-env, which does not know about this `EPHEMERAL_TEST` variable.

test/e2e/e2e_suite_test.go
```diff
@@ -76,6 +79,7 @@ func init() {
 	flag.StringVar(&artifactFolder, "e2e.artifacts-folder", "", "folder where e2e test artifact should be stored")
 	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
 	flag.BoolVar(&upgradeTest, "e2e.trigger-upgrade-test", false, "if true, the e2e upgrade test will be triggered and other tests will be skipped")
+	flag.BoolVar(&ephemeralTest, "e2e.trigger-ephemeral-test", false, "if true, the e2e ephemeral test will be triggered and other tests will be skipped")
```
mboukhalfa, line 82:
```suggestion
	flag.BoolVar(&ephemeralTest, "e2e.trigger-ephemeral-test", false, "if true, all e2e tests run in the ephemeral cluster without pivoting to the target cluster")
```

mboukhalfa (commented):
Please check some suggestions

test/e2e/e2e_test.go
```diff
@@ -51,6 +51,36 @@ var _ = Describe("Workload cluster creation", func() {
 			By("Creating a high available cluster")
 			if upgradeTest {
 				upgradeManagementCluster()
+			} else if ephemeralTest {
```
lentzi90, line 54: It would be nice to avoid duplicating the cluster creation here I think. What if we change the logic like this?

```go
if upgradeTest {
    upgradeManagementCluster()
} else {
    // ApplyClusterTemplateAndWait... same for both ephemeral and "normal" tests
    if ephemeralTest {
        // Run tests in ephemeral cluster
    } else {
        // Run tests with pivot
    }
}
```

test/e2e/cert_rotation_test.go
smoshiur1237: ironic container is not needed as ironic-httpd is taking care of the TLS

lentzi90 (commented):
Great work! I really like how easy the tests could be reused!
Added some questions and suggestions below.

test/e2e/e2e_test.go
smoshiur1237: Yes I am working on it.

test/e2e/upgrade_ironic_test.go
```diff
@@ -32,12 +32,8 @@ func upgradeIronic() {
 	for i, container := range deploy.Spec.Template.Spec.Containers {
 		switch container.Name {
 		case
-			// TODO(dtantsur): remove api and conductor once
-			// they're fully replaced by the all-in-one ironic
 			"ironic",
-			"ironic-api",
 			"ironic-dnsmasq",
-			"ironic-conductor",
```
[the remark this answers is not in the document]
smoshiur1237: We had discussion on it.  ironic-api and ironic-conductor container is not needed on the list and that was the reason I deleted those name from here.

test/e2e/cert_rotation_test.go
smoshiur1237: As ironic-httpd is taking care of TLS, we don't need ironic container

test/e2e/cert_rotation_test.go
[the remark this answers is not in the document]
smoshiur1237: This is for checking if all the container exist in the containerNumRestart list.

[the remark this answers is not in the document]
lentzi90: Thanks for explaining! As discussed offline, let's add a comment so it is easier to understand what it is doing.

test/e2e/upgrade_ironic_test.go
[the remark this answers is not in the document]
lentzi90: I see, fine by me! :heavy_check_mark:

test/e2e/cert_rotation_test.go
mboukhalfa: True that's different check then

PR #573, opened 2022-03-17
🌱 Don't write hardware details into statusAnnotation

fmuyassarov:
Don't write hardware details into `statusAnnotation`, because we don't need to move
inspection data to the target cluster via annotation anymore but do it instead via
hardwareData CRD, which is implemented in https://github.com/metal3-io/baremetal-operator/pull/1099.
Partially fixes https://github.com/metal3-io/cluster-api-provider-metal3/issues/266

### DIFF
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -864,6 +864,16 @@ var _ = Describe("Metal3Machine manager", func() {
 				Expect(statusPresent).To(BeTrue())
 				annotation, err := json.Marshal(&tc.Host.Status)
 				Expect(err).To(BeNil())
+				// (Note) manager code marshals the inspection data stored in annotation,
+				// which causes alphabetically reordering of keys. Since we are marshaling
+				// only the annotation, the status value here doesn't match the marshaled
+				// annotation data, because it wasn't re-ordered by the JSON marshaller.
+				// That's why we are marshaling status data as well so that its fields are
+				// also alphabetically reordered to match the annotation keys style..
+				obj := map[string]interface{}{}
+				err = json.Unmarshal(annotation, &obj)
+				Expect(err).To(BeNil())
+				annotation, _ = json.Marshal(obj)
 				Expect(status).To(Equal(string(annotation)))
 			} else {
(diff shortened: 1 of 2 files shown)


lentzi90 (approved):

baremetal/metal3machine_manager_test.go
```diff
@@ -744,6 +744,16 @@ var _ = Describe("Metal3Machine manager", func() {
...
 				annotation, err := json.Marshal(&tc.Host.Status)
 				Expect(err).To(BeNil())
+				// (Note) manager code marshals the inspection data stored in annotation,
+				// which causes alphabetically reordering of keys. Since we are marshaling
+				// only the annotation, the status value here doesn't match the marshaled
+				// annotation data, because it wasn't re-ordered by the JSON marshaller.
+				// That's why we are marshaling status data as well so that its fields are
+				// also alphabetically reordered to match the annotation keys style..
+				obj := map[string]interface{}{}
+				err = json.Unmarshal(annotation, &obj)
```
kashifest, line 754: This is a bit unclear to me , why are you unmarshalling an empty object ?

fmuyassarov:
This PR is ready, adding hold just to not merge it before getting green light on https://github.com/metal3-io/baremetal-operator/pull/1099
/hold

fmuyassarov:
/hold cancel
@kashifest this should be good to go know as BMO PR got merged.

fmuyassarov:
Just leaving a note here @kashifest. Feature wise we have tested it with corresponding BMO, dev-env patches. E2E ubuntu test is failing here due to a common issue, not related to this patch. However, we should run E2E at least once. Not to block the feature landing and to avoid introducing a potential bug related to this feature, let's merge it and once E2E main is fixed we can verify the test with these additions.
Just a heads up, integration test is already checking the functionality of HardwareData. The only difference between E2E and integration tests is that we run some extra tests like node reuse etc during E2E. But since the feature is mainly concerned about pivoting, which is tested during integration, we should be safe to assume that patch is okay to land.

smoshiur1237 (commented):
/lgtm

PR #577, opened 2022-03-22
🌱 Remove golint linter from golangci.yaml and rename golint hack script to golangci-lint

ren-yifei:


furkatgofurov7 (approved):
/retitle 🌱 Remove golint linter from golangci.yaml and rename golint hack script to golangci-lint 
/approve

PR #580, opened 2022-03-23
Parameterize hard coded object/resource names in tests

Xenwar:
As discussed in https://github.com/metal3-io/cluster-api-provider-metal3/pull/563#discussion_r831825842, the resource names in the unit tests are hard coded. 

This PR parameterizes these values. It also removes the centrally defined objects such as the following. 
```
    testObjectMeta = metav1.ObjectMeta{
        Name:     "",
        Namespace: namespaceName,
    }
```
The problem we are facing is that `testObjectMeta` is used for any resource, such as metal3Machine, Machine, cluster and so on. This creates a problem when a specific test cases needs to change the name or namespace.

### DIFF
--- a/baremetal/metal3datatemplate_manager_test.go
+++ b/baremetal/metal3datatemplate_manager_test.go
@@ -301,10 +301,7 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 			},
 			dataClaims: []*infrav1.Metal3DataClaim{
 				{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "abc",
-						Namespace: namespaceName,
-					},
+					ObjectMeta: testObjectMeta("abc", namespaceName, ""),
 					Spec: infrav1.Metal3DataClaimSpec{
 						Template: corev1.ObjectReference{
 							Name:      "abc",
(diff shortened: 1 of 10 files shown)


Xenwar:
/hold
Breaking the PR into smaller ones in order to make reviewing easier.

baremetal/metal3data_manager_test.go
```diff
@@ -155,7 +150,11 @@ var _ = Describe("Metal3Data manager", func() {
 		}),
 		Entry("requeue error", testCaseReconcile{
 			m3d: &infrav1.Metal3Data{
-				ObjectMeta: testObjectMeta,
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "abc",
```
mboukhalfa, line 154: Based on the linked [discussion](https://github.com/metal3-io/cluster-api-provider-metal3/pull/563#discussion_r831825842) it was recommended to use a constant instead of the "abc"

Xenwar: I agree. 
That will come next. For now, we are trying to have each test to have its own object so that we can override the name, namespace and uid values

Xenwar:
Most of the content is handled by previous PRs. 
And, the remaining is handled by a new PR https://github.com/metal3-io/cluster-api-provider-metal3/pull/617

PR #581, opened 2022-03-24
Define ObjectMeta `metav1.ObjectMeta` at Test case level

Xenwar:
This PR is specific to the controller package and is required before parameterizing hard coded resource names.
It mainly removes objects such as the following.
```
    testObjectMeta = metav1.ObjectMeta{
        Name:     "",
        Namespace: namespaceName,
    }
```
The problem we are facing is that objects such as `testObjectMeta` are used for any resource, such as metal3Machine, Machine, cluster and so on. This creates a problem when a specific test cases needs to change the name or namespace.


kashifest (approved):

smoshiur1237 (approved):
/lgtm

PR #583, opened 2022-03-24
🌱 Leave BMO and Ironic configmaps alone

lentzi90:

These configmaps are created/edited from metal3-dev-env when setting up
the environment. It is better to let them stay as they are to avoid
conflicts and mistakes. They are cleaned/reset every time the bootstrap
cluster is lanuched anyway.

Wait for https://github.com/metal3-io/metal3-dev-env/pull/975 before merging this.


smoshiur1237 (commented):
/lgtm

kashifest (approved):

PR #586, opened 2022-03-28
🐛 Fix condition Reasons values for Metal3Machine

schrej:
Replaces the value of `SettingProviderIDOnNodeFailedReason` with `SettingProviderIDOnNodeFailed` so it's easier to figure out what's going when that error occurs.

### DIFF
--- a/api/v1beta1/condition_consts.go
+++ b/api/v1beta1/condition_consts.go
@@ -48,7 +48,7 @@ const (
 	// KubernetesNodeReadyCondition documents the transition of a Metal3Machine into a Kubernetes Node.
 	KubernetesNodeReadyCondition clusterv1.ConditionType = "KubernetesNodeReady"
 	// Could not find the BMH associated with the Metal3Machine
-	MissingBMHReason = "MissingBMHReason"
+	MissingBMHReason = "MissingBMH"
 	// Could not set the ProviderID on the target cluster's Node object.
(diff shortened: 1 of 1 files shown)


api/v1beta1/condition_consts.go
```diff
@@ -50,5 +50,5 @@ const (
 	// Could not find the BMH associated with the Metal3Machine
 	MissingBMHReason = "MissingBMHReason"
```
furkatgofurov7, line 51: This also looks not accurate to me actually, should be i.e:

```suggestion
	MissingBMHReason = "MissingBMH"
```

so it is aligned with other reasons.

furkatgofurov7 (commented):
Thanks for sending a fix @schrej!

furkatgofurov7:
> /test all

Looks like we have some problems with prow, it is not running automatically any of the default prow jobs.

api/v1beta1/condition_consts.go
furkatgofurov7: @schrej would you like to fix this as well?

furkatgofurov7 (approved):
More discussion on this change on this slack thread https://kubernetes.slack.com/archives/CHD49TLE7/p1647969847894829 
/retitle 🐛 Fix condition Reasons values for Metal3Machine
/approve
/cc @kashifest @Xenwar @fmuyassarov

PR #598, opened 2022-04-08
🌱 Reuse capi upgrade test function

mboukhalfa:
This PR remove the custom upgrade test and reuse [clusterctl_upgrade](https://github.com/kubernetes-sigs/cluster-api/blob/release-1.2/test/e2e/clusterctl_upgrade.go) test from capi.

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -58,19 +58,11 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
 # shellcheck disable=SC1091,SC1090
 source "${M3_DEV_ENV_PATH}/lib/ironic_tls_setup.sh"
 
-if [[ ${GINKGO_FOCUS:-} == "upgrade" ]]; then
-  export CAPI_FROM_RELEASE="${CAPIRELEASE}"
-  export CAPI_TO_RELEASE="${CAPI_TO_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.2.")}"
-
-  export CAPM3_FROM_RELEASE="${CAPM3RELEASE}"
-  export CAPM3_TO_RELEASE="${CAPM3_TO_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v1.2.")}"
-else
-  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
-  export CAPI_TO_RELEASE="${CAPIRELEASE}"
-
-  export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
-  export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
-fi
+# Parameterize e2e_config
+export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
+export CAPI_TO_RELEASE="${CAPIRELEASE}"
+export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
+export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
 
 # image for live iso testing
 export LIVE_ISO_IMAGE="https://artifactory.nordix.org/artifactory/metal3/images/iso/minimal_linux_live-v2.iso"
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -61,23 +61,13 @@ export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
 export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
 export CAPM3_LOCAL_IMAGE="${CAPM3PATH}"
 
+export PATH=$PATH:$HOME/.krew/bin
+
 # Upgrade test environment vars and config
 if [[ ${GINKGO_FOCUS:-} == "upgrade" ]]; then
-    export CAPI_VERSION="v1alpha4"
-    export CAPM3_VERSION="v1alpha5"
-    # Ironic and BMO images to start from. They will then upgrade to main/latest
-    export IRONIC_TAG="capm3-v0.5.5"
-    export BAREMETAL_OPERATOR_TAG="capm3-v0.5.5"
+  export NUM_NODES=${NUM_NODES:-"5"} 
 fi
-# Override project infra vars that point to
-# the current branch to build capm3 image and crds
-if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then
-    export CAPM3BRANCH="release-0.5"
-    # This var is set in project infra to use the current repo location for
-    # building CAPM3 image while upgrade needs an old version
-    unset CAPM3_LOCAL_IMAGE
-    export CAPM3PATH="${M3PATH}/cluster-api-provider-metal3"
-fi
+
 if [[ "${IMAGE_OS}" == "ubuntu" ]]; then
   export UPGRADED_IMAGE_NAME="UBUNTU_22.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
   export UPGRADED_RAW_IMAGE_NAME="UBUNTU_22.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -110,6 +110,9 @@ variables:
   CNI: "/tmp/calico.yaml"
   KUBERNETES_VERSION: "v1.24.0"
   UPGRADED_K8S_VERSION: "v1.24.1"
+  # INIT_WITH_KUBERNETES_VERSION will be used here
+  # https://github.com/kubernetes-sigs/cluster-api/blob/bb377163f141d69b7a61479756ee96891f6670bd/test/e2e/clusterctl_upgrade.go#L170
+  INIT_WITH_KUBERNETES_VERSION: ${FROM_K8S_VERSION}
   CONTROL_PLANE_MACHINE_COUNT: 3
   WORKER_MACHINE_COUNT: 1
   APIVersion: "infrastructure.cluster.x-k8s.io/${CAPM3_VERSION}"
@@ -129,6 +132,8 @@ variables:
   # Pin Calico version
   CALICO_PATCH_RELEASE: "v3.21.0"
   CALICO_MINOR_RELEASE: "v3.21"
+  # Pin CertManager for upgrade tests
+  CERT_MANAGER_RELEASE: v1.5.3
   # Default vars for the template, those values could be overridden by the env-vars.
   CAPI_VERSION: "v1beta1"
   CAPM3_VERSION: "v1beta1"
@@ -149,6 +154,7 @@ variables:
   IMAGE_CHECKSUM_TYPE: "md5"
   IMAGE_USERNAME: "metal3"
   NODE_DRAIN_TIMEOUT: "0s"
+
 intervals:
   default/wait-controllers: ["5m", "10s"]
   default/wait-cluster: ["20m", "30s"] # The second time to check the availibility of the cluster should happen late, so kcp object has time to be created
--- a/test/e2e/upgrade_management_cluster_test.go
+++ b/test/e2e/upgrade_management_cluster_test.go
@@ -1,32 +1,21 @@
 package e2e
 
 import (
-	"context"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 	"strings"
 
-	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/pointer"
-	clusterv1alpha4 "sigs.k8s.io/cluster-api/api/v1alpha4"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
-	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/config"
+	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
 	framework "sigs.k8s.io/cluster-api/test/framework"
-	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var _ = Describe("When testing cluster upgrade [upgrade]", func() {
-	upgradeManagementCluster()
-})
+const workDir = "/opt/metal3-dev-env/"
 
-func upgradeManagementCluster() {
+var _ = Describe("When testing cluster upgrade v1alpha5 > current [upgrade]", func() {
 	BeforeEach(func() {
 		osType := strings.ToLower(os.Getenv("OS"))
 		Expect(osType).ToNot(Equal(""))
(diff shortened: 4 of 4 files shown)


furkatgofurov7:

/hold

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -332,3 +350,97 @@ func printImages(clusterProxy framework.ClusterProxy) {
 		}
 	}
 }
+func getPreCreateCluster(ctx context.Context) func(proxy framework.ClusterProxy, managementClusterNamespace string) {
+	return func(clusterProxy framework.ClusterProxy, managementClusterNamespace string) {
+		Logf("Apply the available BMHs CRs")
+		cmd := exec.Command("bash", "-c", "kubectl krew update; kubectl krew install split-yaml")
```
mboukhalfa, line 356: we can add as a workaround to the path /home/metal3ci/.krew/bin

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+func splitBMHostFile() {
+	cmd := exec.Command("bash", "-c", "cat bmhosts_crs.yaml | kubectl split-yaml -t {{.name}}.yaml -p.")
+	cmd.Dir = workDir
+	output, err := cmd.CombinedOutput()
+	Logf("splitting bmhs:n %v", string(output))
+	Expect(err).To(BeNil())
+}
 
-		Expect(bootstrapClusterProxy.Apply(ctx, upgradeClusterTemplate)).To(Succeed())
+func getPreWaitForCluster(ctx context.Context) func(proxy framework.ClusterProxy, clusterNamespace string, clusterName string) {
```
lentzi90, line 54: This needs a doc string and a better name. What is it doing?
Do you think it would be possible to define this as a variable instead before the `Describe`? I mean something like this:

```go
createBMHs := func(clusterProxy framework.ClusterProxy, clusterNamespace string) {...}
// ...
// Then use it like this
PreInit: createBMHs
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
-		}, e2eConfig.GetIntervals(specName, "wait-controllers")...)
-
-		/*-------------------------------------------------*
-		| Pivot to run ironic/BMO and resources on target |
-		*--------------------------------------------------*/
-		By("Start pivoting")
-		listBareMetalHosts(ctx, bootstrapClusterProxy.GetClient(), client.InNamespace(namespace))
-		listNodes(ctx, upgradeClusterClient)
+		}
+		os.Setenv("PROVIDER_ID_FORMAT", "metal3://{{ ds.meta_data.uuid }}")
```
lentzi90, line 94: Why is it set here instead of before we start the test? Is it because of the older version? In any case it needs an explanation.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+	}
+}
 
+func getPreInitFunc(ctx context.Context) func(proxy framework.ClusterProxy) {
+	return func(clusterProxy framework.ClusterProxy) {
+		// install certmanager
+		err := downloadFile("/tmp/certManager.yaml", "https://github.com/cert-manager/cert-manager/releases/download/v1.5.3/cert-manager.yaml")
```
lentzi90, line 101: Version is hard coded. We should parameterize this.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
-					exist = true
-					break
-				}
-			}
-			if !exist {
-				images = append(images, c.Image)
-				Logf("%v", c.Image)
-			}
-		}
+func getPreUpgrade(ctx context.Context) func(proxy framework.ClusterProxy) {
```
lentzi90, line 133: Same. This also needs a better name and doc string.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+		applyBMH := func(node_index int) {
+			node := fmt.Sprintf("node-%v", node_index)
+			cmd := exec.Command("kubectl", "apply", "-f", fmt.Sprintf("/opt/metal3-dev-env/%v-bmc-secret.yaml", node), "-n", clusterNamespace, "--kubeconfig", clusterProxy.GetKubeconfigPath())
+			cmd.Dir = workDir
+			output, err := cmd.CombinedOutput()
+			Logf(string(output))
+			Expect(err).To(BeNil())
+			cmd = exec.Command("kubectl", "apply", "-f", fmt.Sprintf("/opt/metal3-dev-env/%v.yaml", node), "-n", clusterNamespace, "--kubeconfig", clusterProxy.GetKubeconfigPath())
+			cmd.Dir = workDir
+			output, err = cmd.CombinedOutput()
```
lentzi90, line 69: If we change the `splitBMHostFile` like suggested, we could make this easier to understand by looping over the file names directly. I would also want to use `clusterProxy.Apply` like for cert-manager below.

lentzi90 (changes requested):
This will be nice when finished!
/hold
https://github.com/metal3-io/cluster-api-provider-metal3/pull/754 should go in first so we don't need the uplift in this one.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+var _ = Describe("When testing cluster upgrade v1alpha5 > current [upgrade]", func() {
+	capi_e2e.ClusterctlUpgradeSpec(ctx, func() capi_e2e.ClusterctlUpgradeSpecInput {
+		return capi_e2e.ClusterctlUpgradeSpecInput{
+			E2EConfig:                 e2eConfig,
+			ClusterctlConfigPath:      clusterctlConfigPath,
+			BootstrapClusterProxy:     bootstrapClusterProxy,
+			ArtifactFolder:            artifactFolder,
+			SkipCleanup:               true, // parameterize after merging https://github.com/kubernetes-sigs/cluster-api/pull/7373
+			InitWithProvidersContract: "v1alpha4",
+			InitWithBinary:            "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/clusterctl-{OS}-{ARCH}",
```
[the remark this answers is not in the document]
mboukhalfa: I think here we are testing upgrade from `v1alpha4` (capi) so setting this in the environment somewhere then propagate it to here with some logic doesn't seem needed so far there is not case where `InitWithProvidersContract` will be `v1beta1` and since the support for this api versoin already dropped I do not think we will get patches for `v0.4.x` so `v0.4.7` is the latest.

It is hardcoded but it makes debugging the input easier although we can move them to e2e read them from there

scripts/ci-e2e.sh
```diff
@@ -58,19 +58,11 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
...
-  export CAPM3_FROM_RELEASE="${CAPM3RELEASE}"
-  export CAPM3_TO_RELEASE="${CAPM3_TO_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v1.2.")}"
-else
-  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
-  export CAPI_TO_RELEASE="${CAPIRELEASE}"
-
-  export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
-  export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
-fi
+# Prameterize e2e_config
```
furkatgofurov7, line 61: nit: typo

scripts/environment.sh
```diff
@@ -61,22 +61,10 @@ export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
 export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
 export CAPM3_LOCAL_IMAGE="${CAPM3PATH}"
 
+export PATH=$PATH:$HOME/.krew/bin
```
furkatgofurov7, line 64: is this needed?

scripts/environment.sh
```diff
@@ -61,22 +61,10 @@ export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
...
-fi
-# Override project infra vars that point to
-# the current branch to build capm3 image and crds
-if [[ "${CAPM3_VERSION}" == "v1alpha5" ]]; then
-    export CAPM3BRANCH="release-0.5"
-    # This var is set in project infra to use the current repo location for
-    # building CAPM3 image while upgrade needs an old version
-    unset CAPM3_LOCAL_IMAGE
-    export CAPM3PATH="${M3PATH}/cluster-api-provider-metal3"
+export NUM_NODES=${NUM_NODES:-"5"}
```
furkatgofurov7, line 67: I am not understanding why we would need all of these

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+const workDir = "/opt/metal3-dev-env/"
+
+var _ = Describe("When testing cluster upgrade v1alpha5 > current [upgrade]", func() {
+	capi_e2e.ClusterctlUpgradeSpec(ctx, func() capi_e2e.ClusterctlUpgradeSpecInput {
+		return capi_e2e.ClusterctlUpgradeSpecInput{
+			E2EConfig:                 e2eConfig,
+			ClusterctlConfigPath:      clusterctlConfigPath,
+			BootstrapClusterProxy:     bootstrapClusterProxy,
+			ArtifactFolder:            artifactFolder,
+			SkipCleanup:               true, // parameterize after merging https://github.com/kubernetes-sigs/cluster-api/pull/7373
```
furkatgofurov7, line 25: can we make this a bit clearer, with TODO note, i.e:  `// TODO(mboukhalfa):  parameterize after merging https://github.com/kubernetes-sigs/cluster-api/pull/7373`

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
-			WorkerMachineCount:       pointer.Int64Ptr(1),
-			InfrastructureProvider:   clusterctl.DefaultInfrastructureProvider,
-			// setup clusterctl logs folder
-			LogFolder: clusterctlLogFolder,
-		})
+})
 
-		Expect(upgradeClusterTemplate).ToNot(BeNil(), "Failed to get the cluster template")
+// Install the split-yaml tool to split our bmhosts_crs.yaml file.
+func installSplitYaml() {
```
furkatgofurov7, line 38: with respect to Go naming conventions:
```suggestion
func installSplitYAML() {
```

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+	Expect(err).To(BeNil())
+}
 
-		Logf("Applying the cluster template yaml to the cluster")
+// Split the bmhosts_crs.yaml file to make it easy to apply limited number on BMH in each namespace.
+func splitBMHostFile() {
+	cmd := exec.Command("bash", "-c", "cat bmhosts_crs.yaml | kubectl split-yaml -t {{.name}}.yaml -p.")
+	cmd.Dir = workDir
+	output, err := cmd.CombinedOutput()
+	Logf("splitting bmhs:n %v", string(output))
```
furkatgofurov7, line 50: this has to be reworded definitely

test/e2e/config/e2e_conf.yaml
```diff
@@ -149,6 +149,7 @@ variables:
   IMAGE_CHECKSUM_TYPE: "md5"
   IMAGE_USERNAME: "metal3"
   NODE_DRAIN_TIMEOUT: "0s"
+  INIT_WITH_KUBERNETES_VERSION: "v1.24.1"
```
furkatgofurov7, line 152: how is this specific version was chosen? based on `UPGRADED_K8S_VERSION`? do we need an extra variable at all?
Edit: I don't see it being used anywhere

test/e2e/upgrade_management_cluster_test.go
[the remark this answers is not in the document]
furkatgofurov7: @mboukhalfa you already have env var for them: `CAPI_FROM_RELEASE`

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -2,340 +2,137 @@ package e2e
...
+			BootstrapClusterProxy:     bootstrapClusterProxy,
+			ArtifactFolder:            artifactFolder,
+			SkipCleanup:               true, // parameterize after merging https://github.com/kubernetes-sigs/cluster-api/pull/7373
+			InitWithProvidersContract: "v1alpha4",
+			InitWithBinary:            "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/clusterctl-{OS}-{ARCH}",
+			PreInit:                   getPreInitFunc(ctx),
+			PreWaitForCluster:         getPreWaitForCluster(ctx),
+			PreUpgrade:                getPreUpgrade(ctx),
+			MgmtFlavor:                osType,
+			WorkloadFlavor:            "ubuntu",
```
furkatgofurov7, line 32: parameterize would be better

furkatgofurov7 (commented):
This looks like a nice improvement, however, I would like to understand better the workflow with these changes introduced:

test/e2e/upgrade_management_cluster_test.go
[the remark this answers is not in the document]
lentzi90: Ok maybe v1alpha4 is fine but don't we already have the release here: https://github.com/metal3-io/cluster-api-provider-metal3/blob/6b3b651f81b54a48c9f39d5d402baf58c26af225/scripts/ci-e2e.sh#L62

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: added doc I prefer to keep the generic name as it may do more than just creating bmh for example exporting some vars and to be consistent with other hooks

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: if exported before it breaks creating the management because it uses v1beta1 templates

mboukhalfa: WIP comming soon :relaxed:

scripts/environment.sh
mboukhalfa: yeah for some reason when devenv install krew it will not be available in the path here ! I had to add it

scripts/environment.sh
mboukhalfa: capi test add scale up check after upgrade and for this we need additional bmh https://github.com/kubernetes-sigs/cluster-api/blob/4d8c88f24430d4fd15109e53fba57ad2acf5ddc8/test/e2e/clusterctl_upgrade.go#L393

test/e2e/config/e2e_conf.yaml
mboukhalfa: hmm for now we are using what ever capi would define in their e2e config file but we can override that from the environment variable

scripts/environment.sh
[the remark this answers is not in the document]
mboukhalfa: THANKS

test/e2e/upgrade_management_cluster_test.go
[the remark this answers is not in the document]
mboukhalfa: done

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: done

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: fixed

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: I still want to keep same generic names for call hooks added docs

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: done

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: done

test/e2e/upgrade_management_cluster_test.go
lentzi90: Ok my point is not the outer function. It is clear that it just returns the inner function. My point is the inner function has no description for what it is doing and also no name so there is no way to understand it other than reading the code. It is not necessary to do it in this way. You can define the function separately and give it a name, then just pass it to the upgrade spec directly.
Like this:

```go
func myFancyPreWaitForClusterFunction(clusterProxy framework.ClusterProxy, clusterNamespace string, clusterName string) {
...
}

var _ = Describe("When testing cluster upgrade v1alpha5 > current [upgrade]", func() {
	capi_e2e.ClusterctlUpgradeSpec(ctx, func() capi_e2e.ClusterctlUpgradeSpecInput {
		return capi_e2e.ClusterctlUpgradeSpecInput{
			E2EConfig:                 e2eConfig,
			ClusterctlConfigPath:      clusterctlConfigPath,
			BootstrapClusterProxy:     bootstrapClusterProxy,
			ArtifactFolder:            artifactFolder,
			SkipCleanup:               true, // parameterize after merging https://github.com/kubernetes-sigs/cluster-api/pull/7373
			InitWithProvidersContract: "v1alpha4",
			InitWithBinary:            "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.7/clusterctl-{OS}-{ARCH}",
			PreInit:                   getPreInitFunc(ctx),
			PreWaitForCluster:         myFancyPreWaitForClusterFunction,
			PreUpgrade:                getPreUpgrade(ctx),
			MgmtFlavor:                osType,
			WorkloadFlavor:            "ubuntu",
		}
	})
})
```

test/e2e/config/e2e_conf.yaml
mboukhalfa: in fact this will be used in the in the capi upgrade test since we are passing our e2e config as input and and the test look for a var with this name although it has same value as `UPGRADED_K8S_VERSION`

lentzi90 (commented):
Thanks for working on this!

Now I'm happy with how we pass the functions to the upgrade spec input. But I still want better descriptions and names for them. It is obvious that they are used for `Pre{Init|WaitForCluster|Upgrade}` since we pass them to the spec, but we need an explanation for what they are actually doing.

So for PreInit for example, ok we pass a function to this so it is obvious we want to do something before initializing. What are we doing? That is the question. On a high level maybe `SetupIronicBMO` could work?

mboukhalfa:
> Thanks for working on this!
> 
> Now I'm happy with how we pass the functions to the upgrade spec input. But I still want better descriptions and names for them. It is obvious that they are used for `Pre{Init|WaitForCluster|Upgrade}` since we pass them to the spec, but we need an explanation for what they are actually doing.
> 
> So for PreInit for example, ok we pass a function to this so it is obvious we want to do something before initializing. What are we doing? That is the question. On a high level maybe `SetupIronicBMO` could work?

I think the current names are better to easily find which function runs in which step and as the outer function just decorate inner functions that have specific names I think that's better

lentzi90 (commented):
I would still want more descriptive names for the `pre*` functions, but I can live with the descriptions. Let's see what others say.

/lgtm

mboukhalfa:
> I would still want more descriptive names for the `pre*` functions, but I can live with the descriptions. Let's see what others say.
> 
> /lgtm

I guess the current way is more conform with CAPI since pre is common with all hooks and differentiate them from other functions

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -35,307 +24,128 @@ func upgradeManagementCluster() {
...
-			ObjectMeta: metav1.ObjectMeta{
-				Name: e2eConfig.GetVariable("IRONIC_NAMESPACE"),
-			},
+	capi_e2e.ClusterctlUpgradeSpec(ctx, func() capi_e2e.ClusterctlUpgradeSpecInput {
+		return capi_e2e.ClusterctlUpgradeSpecInput{
+			E2EConfig:                 e2eConfig,
+			ClusterctlConfigPath:      clusterctlConfigPath,
+			BootstrapClusterProxy:     bootstrapClusterProxy,
+			ArtifactFolder:            artifactFolder,
+			SkipCleanup:               true, // TODO(mboukhalfa): Parameterize after merging https://github.com/kubernetes-sigs/cluster-api/pull/7373
```
furkatgofurov7, line 33: linked PR is merged, can you do the parameterization now?

test/e2e/config/e2e_conf.yaml
furkatgofurov7: Let's set it to `UPGRADED_K8S_VERSION` not to change it in two places in the future?

scripts/environment.sh
furkatgofurov7: I am not sure got this right

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -35,307 +24,128 @@ func upgradeManagementCluster() {
...
+				Expect(clusterProxy.Apply(ctx, resource, []string{"-n", clusterNamespace}...)).ShouldNot(HaveOccurred())
 			}
-			g.Expect(upgradeClusterClient.List(ctx, machineList, opts...)).To(Succeed())
-			for _, machine := range machineList.Items {
-				if machine.Status.GetTypedPhase() == clusterv1alpha4.MachinePhaseRunning {
-					n++
-				}
+		} else {
+			// Apply secrets and bmhs for [node_2, node_3 and node_4] in the management cluster to host workload cluster
+			for i := 4; i < 10; i++ {
```
furkatgofurov7, line 90: how this is iterating through if you have only two/there BMHs here/below?

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -35,307 +24,128 @@ func upgradeManagementCluster() {
...
+		}
+	}
 
-		By("Get the bootstrap cluster images after upgrading")
-		printImages(bootstrapClusterProxy)
+	// Create bmhs in the in the namespace that will host the new cluster
+	createBMH(clusterProxy, clusterNamespace, clusterName)
+	// This is exported here because we need this only when creating the workload
+	// if exported before it breaks creating the management because it uses v1beta1 templates
+	os.Setenv("PROVIDER_ID_FORMAT", "metal3://{{ ds.meta_data.uuid }}")
```
furkatgofurov7, line 102: why this is set at the end?

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -35,307 +24,128 @@ func upgradeManagementCluster() {
...
+
+	// Export capi/capm3 versions
+	os.Setenv("CAPI_VERSION", "v1alpha4")
+	os.Setenv("CAPM3_VERSION", "v1alpha5")
+	// override default IPs for the workload cluster
+	os.Setenv("CLUSTER_APIENDPOINT_HOST", "192.168.111.250")
+	os.Setenv("BAREMETALV4_POOL_RANGE_START", "192.168.111.201")
+	os.Setenv("BAREMETALV4_POOL_RANGE_END", "192.168.111.240")
+	os.Setenv("PROVISIONING_POOL_RANGE_START", "172.22.0.201")
+	os.Setenv("PROVISIONING_POOL_RANGE_END", "172.22.0.240")
```
furkatgofurov7, line 143: same here, after all why do not use set them in the e2e conf file?

furkatgofurov7 (commented):
This is good already, my second pass on it:

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: We should wait for the new patch release

furkatgofurov7:
> I guess the current way is more conform with CAPI since pre is common with all hooks and differentiate them from other functions

+1 to keep it sync with CAPI, unless we have a proper description for them

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: this specific to the workload if we exported then before we will override the first management cluster

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: we can move this to pre-init with the IP override it is the same at the end what is important is to export this after applying the first target cluster template

mboukhalfa: or even export is in the if block `if isBootstrapProxy {`

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: we know that file are ordered `secret_node_0.yaml, bmh_node_0.yaml, ...,secret_node_n.yaml, bmh_node_n.yaml..` so the first iteration is to apply the secret then the second to apply bmh for each node

mboukhalfa:
/test-ubuntu-integration-main
/test-upgrade-e2e-main

@lentzi90  can you please re LGTM if the recent changes are ok for you

lentzi90:
> I guess the current way is more conform with CAPI since pre is common with all hooks and differentiate them from other functions

I don't agree. CAPI only provide the spec. They don't even use these functions. To name our custom functions the same as the spec is not the same as conforming with CAPI. We use their spec and that is enough to make it conformant. What we name the input is up to us, and IMO should describe what it actually does, not "when it is called".
Just for the record :wink: 

But I'm still ok with merging if others are happy with this naming.
/lgtm

test/e2e/config/e2e_conf.yaml
```diff
@@ -129,6 +132,8 @@ variables:
   # Pin Calico version
   CALICO_PATCH_RELEASE: "v3.21.0"
   CALICO_MINOR_RELEASE: "v3.21"
+  # Pin CertManager for upgrade tests
+  CERT_MANAGER_RELEASE: v1.5.3
```
furkatgofurov7, line 136: @mboukhalfa what is the reason behind pinning it to this specific version? due to v1a5? please keep in mind https://github.com/metal3-io/cluster-api-provider-metal3/pull/755 is bumping it to 1.9.1

mboukhalfa: yes this is the version used by CAPI `v0.4.8` I am not sure if we uplifted to a newer version what will happen after `clusterctl init ` anyway I prefer to merge this PR then #755 will be rebased and can use the version the more adequate

furkatgofurov7 (approved):

test/e2e/config/e2e_conf.yaml
furkatgofurov7: 755 bumps it in Tiltfile, so this should be fine.

PR #599, opened 2022-04-08
🌱 e2e conf parameterization

adilGhaffarDev:
Parameterizes the CAPI and CAPM3 versions in e2e conf.

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -24,7 +24,7 @@ clone_repo "${M3_DEV_ENV_REPO}" "${M3_DEV_ENV_BRANCH}" "${M3_DEV_ENV_PATH}"
 cat <<-EOF > "${M3_DEV_ENV_PATH}/config_${USER}.sh"
 export CAPI_VERSION=${CAPI_VERSION:-"v1beta1"}
 export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
-export NUM_NODES=${NUM_NODES:-"4"} 
+export NUM_NODES=${NUM_NODES:-"4"}
 export KUBERNETES_VERSION=${FROM_K8S_VERSION}
 export IMAGE_OS=${IMAGE_OS}
 export FORCE_REPO_UPDATE="false"
@@ -58,7 +58,21 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
 # shellcheck disable=SC1091,SC1090
 source "${M3_DEV_ENV_PATH}/lib/ironic_tls_setup.sh"
 
-# run e2e tests 
+if $UPGRADE_TEST; then
+  export CAPI_FROM_RELEASE="${CAPIRELEASE}"
+  export CAPI_TO_RELEASE="${CAPI_TO_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.1.")}"
+
+  export CAPM3_FROM_RELEASE="${CAPM3RELEASE}"
+  export CAPM3_TO_RELEASE="${CAPM3_TO_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v1.1.")}"
+else
+  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
+  export CAPI_TO_RELEASE="${CAPIRELEASE}"
+
+  export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
+  export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
+fi
+
+# run e2e tests
 make e2e-tests
 
 # Clean devenv
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -18,8 +18,8 @@ providers:
 - name: cluster-api
   type: CoreProvider
   versions:
-  - name: v0.4.8
-    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.8/core-components.yaml"
+  - name: ${CAPI_FROM_RELEASE}
+    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CAPI_FROM_RELEASE}/core-components.yaml"
     type: "url"
     contract: v1alpha4
     replacements:
@@ -27,9 +27,9 @@ providers:
         new: --metrics-addr=:8080
     files:
       - sourcePath: "../data/shared/v1alpha4/metadata.yaml"
-  - name: v1.1.3
+  - name: ${CAPI_TO_RELEASE}
     # Use manifest from source files
-    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.1.3/core-components.yaml"
+    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CAPI_TO_RELEASE}/core-components.yaml"
     type: "url"
     contract: v1beta1
     files:
@@ -40,8 +40,8 @@ providers:
 - name: kubeadm
   type: BootstrapProvider
   versions:
-  - name: v0.4.8 # latest published release in the v1alpha4 series; this is used for v1alpha4 --> v1beta1 clusterctl upgrades test only.
-    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.8/bootstrap-components.yaml"
+  - name: ${CAPI_FROM_RELEASE} # latest published release in the v1alpha4 series; this is used for v1alpha4 --> v1beta1 clusterctl upgrades test only.
+    value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CAPI_FROM_RELEASE}/bootstrap-components.yaml"
     type: "url"
     contract: v1alpha4
     replacements:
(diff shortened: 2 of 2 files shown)


scripts/environment.sh
```diff
@@ -57,6 +57,10 @@ export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
 export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
 export CAPM3_LOCAL_IMAGE="${CAPM3PATH}"
 
+export CAPM3_RELEASE = "v1.1.1"
```
mboukhalfa, line 60: You can use `CAPM3RELEASE` directly since it is already exported from the sourced **devenv path** _`lib/releases.sh`_

scripts/environment.sh
```diff
@@ -57,6 +57,14 @@ export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
 export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
 export CAPM3_LOCAL_IMAGE="${CAPM3PATH}"
 
+export CAPIRELEASE="v1.1.3"
```
mboukhalfa, line 60: in the feature test case the script release will export the latest release based on the version in this case it is v1beta1 so you do not need to hard code that I think you can remove it safely from here.
for the other `v0.4.8` yeah it is not exported but still better to not hard code it but use something like in the release script to get the latest based on the version v1alpha5.

Remember! The goal of this work is to not hard code releaseversions

scripts/ci-e2e.sh
```diff
@@ -58,7 +84,27 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
...
-# run e2e tests 
+# CAPM3 and CAPI release path
+CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
+CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+
+# CAPI releases
+export CAPI_TO_RELEASE="${CAPI_TO_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.1.")}"
+if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
+  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.3.")}"
+else
```
kashifest, line 95: I guess this should be 
```suggestion
else if [ "${CAPI_VERSION}" == "v1alpha4" ]; then
```

scripts/ci-e2e.sh
```diff
@@ -58,7 +84,27 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
...
+  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
+fi
+
+# CAPM3 releases
+export CAPM3_TO_RELEASE="${CAPM3_TO_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v1.1.")}"
+if [ "${CAPM3_VERSION}" == "v1alpha4" ]; then
+  export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.4.")}"
+else
```
kashifest, line 103: I guess this should be 
```suggestion
else if [ "${CAPM3_VERSION}" == "v1alpha5" ]; then
```

kashifest (changes requested):

test/e2e/config/e2e_conf.yaml
```diff
@@ -18,7 +18,7 @@ providers:
 - name: cluster-api
   type: CoreProvider
   versions:
-  - name: v0.4.8
+  - name: ${CAPI_FROM_RELEASE} #v0.4.8
```
furkatgofurov7, line 21: It is better not to specify the version (i.e #v0.4.8) even as a comment here and in all other cases in this file since they will get outdated quickly and each time needs to be changed when CAPI makes a new release.

furkatgofurov7 (changes requested):
On line 127, there is a `CAPM3_REL_TO_VERSION` set to hardcoded value but not used anywhere, please remove it. Also, on line 129, `INIT_WITH_BINARY` needs to be parameterized with `CAPI_FROM_RELEASE` variable

test/e2e/config/e2e_conf.yaml
```diff
@@ -71,7 +71,7 @@ providers:
         new: --metrics-addr=:8080
     files:
       - sourcePath: "../data/shared/v1alpha4/metadata.yaml"
-  - name: v1.1.3
+  - name: ${CAPI_TO_RELEASE} #v1.1.3
     # Use manifest from source files
     value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.1.3/control-plane-components.yaml"
```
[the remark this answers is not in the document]
adilGhaffarDev: Fixed.

scripts/ci-e2e.sh
adilGhaffarDev: Fixed it in latest commit, kindly check.

scripts/ci-e2e.sh
adilGhaffarDev: Fixed it in latest commit, kindly check.

scripts/ci-e2e.sh
```diff
@@ -58,7 +84,27 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
...
+CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
+CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+
+# CAPI releases
+export CAPI_TO_RELEASE="${CAPI_TO_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.1.")}"
+if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
+  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.3.")}"
+elif [ "${CAPI_VERSION}" == "v1alpha4" ] || [ "${CAPI_VERSION}" == "v1beta1" ]; then
+  export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
+fi
```
furkatgofurov7, line 97: In reality, we would never hit this check since we go at least from CAPI v0.4.X ==> CAPI v1.1.X and I believe always in an ascending fashion in the future, so _v1alpha3_ check can be dropped, meaning there would be no need for further checking:


```suggestion
export CAPI_FROM_RELEASE="${CAPI_FROM_RELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v0.4.")}"
```

scripts/ci-e2e.sh
```diff
@@ -58,7 +84,27 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
...
+
+# CAPM3 releases
+export CAPM3_TO_RELEASE="${CAPM3_TO_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v1.1.")}"
+if [ "${CAPM3_VERSION}" == "v1alpha4" ]; then
+  export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.4.")}"
+elif [ "${CAPM3_VERSION}" == "v1alpha5" ] || [ "${CAPM3_VERSION}" == "v1beta1" ]; then
+  export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
+fi
```
furkatgofurov7, line 105: The same logic applies to this check as well, please see the above comment

furkatgofurov7 (changes requested):
This is looking good, one last thing I forgot to mention in my previous review:

scripts/ci-e2e.sh
```diff
@@ -1,5 +1,31 @@
...
+
+  if [[ "$release_tag" == "null" ]]; then
+    set -x
+    exit 1
+  fi
+  set -x
+  # shellcheck disable=SC2005
+  echo "$release_tag"
+}
+
```
furkatgofurov7, line 28: I would suggest to put this to environment.sh since there we have lots of similar high level functions defined there

scripts/ci-e2e.sh
```diff
@@ -1,5 +1,31 @@
...
 set -eux pipefail
```
furkatgofurov7, line 29: this should be top level definition in the script, so if you move get_latest_release than this should be fine as is, otherwise move it before get_latest_release

scripts/ci-e2e.sh
```diff
@@ -52,13 +78,41 @@ source "${REPO_ROOT}/hack/ensure-kustomize.sh"
 # shellcheck disable=SC1091,SC1090
 source "${M3_DEV_ENV_PATH}/lib/images.sh"
 # shellcheck disable=SC1091,SC1090
-source "${M3_DEV_ENV_PATH}/lib/releases.sh"
-# shellcheck disable=SC1091,SC1090
```
furkatgofurov7, line 56: After going through this change once again, I noticed that we are already sourcing other scripts here (images.sh, ironic related etc), what is the benefit of not sourcing release.sh the way it was and just set _FROM and _TO vars based on the already populated vars from m3-dev-env make run? That would avoid duplicate code and reduce maintenance of same scripts at 2 different locations.
In short, I am suggesting remove all these release fetching and setting CAPI/CAPM3 versions based on it, and rather  source releases.sh as before, and just set _FROM and _TO  based on the vars from `make` ,

furkatgofurov7 (approved):
Looks much more cleaner, thanks!

/approve

PR #602, opened 2022-04-20
:seedling: Increase judgment to avoid panic

zhouhao3:

Avoid panic when host is nil.

Fixes #

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -1052,7 +1052,7 @@ func (m *MachineManager) nodeReuseLabelExists(ctx context.Context, host *bmov1al
 
 // getBMCSecret will return the BMCSecret associated with BMH.
 func (m *MachineManager) getBMCSecret(ctx context.Context, host *bmov1alpha1.BareMetalHost) (*corev1.Secret, error) {
-	if host.Spec.BMC.CredentialsName == "" {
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (approved):
/approve
/cc @kashifest @smoshiur1237

baremetal/metal3machine_manager.go
```diff
@@ -1052,7 +1052,7 @@ func (m *MachineManager) nodeReuseLabelExists(ctx context.Context, host *bmov1al
 
 // getBMCSecret will return the BMCSecret associated with BMH.
 func (m *MachineManager) getBMCSecret(ctx context.Context, host *bmov1alpha1.BareMetalHost) (*corev1.Secret, error) {
-	if host.Spec.BMC.CredentialsName == "" {
+	if host == nil || host.Spec.BMC.CredentialsName == "" {
```
kashifest, line 1055: I believe this condition is already checked before you call this function no?

zhouhao3: Yse, from the current code, the `getBMCSecret` function is called after verifying that the host is not nil.
But I think `getBMCSecret` still needs this verification to avoid future calling it without verifying that the host is not nil.

kashifest: ok, for safety purpose, I agree
/lgtm

PR #604, opened 2022-04-22
:seedling: Add url validation to the subfield of image

zhouhao3:

Verify that the values of Image.URL and Image.Checksum are valid urls.

Fixes #

### DIFF
--- a/api/v1beta1/metal3machine_types.go
+++ b/api/v1beta1/metal3machine_types.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"net/url"
+
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


api/v1beta1/metal3machine_types.go
```diff
@@ -91,6 +93,19 @@ func (s *Metal3MachineSpec) IsValid() error {
...
+	_, err := url.ParseRequestURI(s.Image.URL)
+	if err != nil {
+		invalid = append(invalid, "Image.URL")
+	}
+	_, err = url.ParseRequestURI(s.Image.Checksum)
+	if err != nil {
+		invalid = append(invalid, "Image.Checksum")
+	}
+	if len(invalid) > 0 {
+		return errors.Errorf("Invalid URL: %v", invalid)
```
furkatgofurov7, line 107: Is this log accurate considering the case where there is an invalid CheckSum as well?

zhouhao3: Because the above verifies whether the URL and Checksum are in the correct url format, the URL here is the representation format instead of Image.URL.
Maybe my log information is not well expressed.
What do you think about this change:
```
return errors.Errorf("The following fields are invalid URLs: %v", invalid)
```

furkatgofurov7: To keep it consistent with the previous validation (missing), it could be:
```
return errors.Errorf("Invalid fields from ProviderSpec: %v", invalid)
```

zhouhao3: updated, thanks.

furkatgofurov7 (approved):
Thanks @zhouhao3, this looks good 
/approve
/cc @kashifest @smoshiur1237

smoshiur1237 (approved):
Thanks @zhouhao3  for adding the url validation. 
/lgtm

PR #608, opened 2022-04-26
:seedling: Fix nil point error in metal3remediation_manager

zhouhao3:

Fixes #

### DIFF
--- a/baremetal/metal3remediation_manager_test.go
+++ b/baremetal/metal3remediation_manager_test.go
@@ -574,6 +574,12 @@ var _ = Describe("Metal3Remediation manager", func() {
 			},
 			RemediationTypeSet: false,
 		}),
+		Entry("Remediation strategy is nil should return empty string", testCaseGetRemediationType{
+			Metal3Remediation: &infrav1.Metal3Remediation{
+				Spec: infrav1.Metal3RemediationSpec{},
+			},
+			RemediationTypeSet: false,
+		}),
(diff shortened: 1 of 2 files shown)


baremetal/metal3remediation_manager_test.go
```diff
@@ -574,6 +574,12 @@ var _ = Describe("Metal3Remediation manager", func() {
 			},
 			RemediationTypeSet: false,
 		}),
+		Entry("Remediation strategy is nil should return empty string", testCaseGetRemediationType{
+			Metal3Remediation: &infrav1.Metal3Remediation{
+				Spec: infrav1.Metal3RemediationSpec{},
+			},
+			RemediationTypeSet: false,
+		}),
```
furkatgofurov7, line 582: This looks like a duplicate of the previous entry eventually (one omitting `Strategy` field at all and the other the `Type` inside the `Strategy` )?

zhouhao3: This test case is added to test the handling of the case where `Strategy` is nil. If there is no code added in the last commit, there will be a panic here.

furkatgofurov7: That is true, sorry I was wrong, this is a different test case actually

furkatgofurov7 (approved):
/approve

kashifest (commented):
/lgtm

PR #612, opened 2022-04-29
🌱 Add CentOS stream 9 in e2e test scripts

kashifest:


scripts/environment.sh
```diff
@@ -24,7 +24,7 @@ function os_check() {
     export DISTRO="${ID}${VERSION_ID%.*}"
     export OS="${ID}"
     export OS_VERSION_ID=$VERSION_ID
-    export SUPPORTED_DISTROS=(centos8 rhel8 ubuntu20)
+    export SUPPORTED_DISTROS=(centos8 centos9 rhel8 ubuntu20)
 
     if [[ ! "${SUPPORTED_DISTROS[*]}" =~ $DISTRO ]]; then
         echo "Supported OS distros for the host are: CentOS Stream 8 or RHEL8 or Ubuntu20.04"
```
furkatgofurov7, line 30:
```suggestion
        echo "Supported OS distros for the host are: CentOS Stream 8/9 or RHEL8/9 or Ubuntu20.04"
```

kashifest: Aha, true, I have to run the tests again though, I will run the integration test only when I change this and once e2e passes on this PR. By the way why are you suggesting RHEL 9 also, does it mean supporting CentOS 9 also supports RHEL 9?

furkatgofurov7: Once e2e passes and pushing, we can just re-run integration tests, agree

kashifest: Ok, I will change it to add RHEL 9 as well

furkatgofurov7 (approved):
/approve

Please fix the echo output as agreed.

smoshiur1237 (approved):
/lgtm

Rozzii (approved):

PR #618, opened 2022-05-06
:seedling: Improve the verification of Image

zhouhao3:

Add verification processing when DiskFormat is live-iso.
Checksum is not required for live-iso.

### DIFF
--- a/api/v1beta1/metal3machine_types.go
+++ b/api/v1beta1/metal3machine_types.go
@@ -87,7 +87,8 @@ func (s *Metal3MachineSpec) IsValid() error {
 	if s.Image.URL == "" {
 		missing = append(missing, "Image.URL")
 	}
-	if s.Image.Checksum == "" {
+	// Checksum is not required for live-iso.
+	if (s.Image.DiskFormat == nil || *s.Image.DiskFormat != "live-iso") && s.Image.Checksum == "" {
 		missing = append(missing, "Image.Checksum")
 	}
 	if len(missing) > 0 {
@@ -99,9 +100,11 @@ func (s *Metal3MachineSpec) IsValid() error {
 	if err != nil {
 		invalid = append(invalid, "Image.URL")
 	}
-	_, err = url.ParseRequestURI(s.Image.Checksum)
-	if err != nil {
-		invalid = append(invalid, "Image.Checksum")
+	if s.Image.DiskFormat == nil || *s.Image.DiskFormat != "live-iso" {
+		_, err = url.ParseRequestURI(s.Image.Checksum)
+		if err != nil {
+			invalid = append(invalid, "Image.Checksum")
+		}
 	}
 	if len(invalid) > 0 {
 		return errors.Errorf("Invalid fields from ProviderSpec: %v", invalid)
--- a/api/v1beta1/metal3machine_types_test.go
+++ b/api/v1beta1/metal3machine_types_test.go
@@ -27,6 +27,7 @@ import (
 )
 
 func TestSpecIsValid(t *testing.T) {
+	diskFormat := "live-iso"
 	cases := []struct {
 		Spec          Metal3MachineSpec
 		ErrorExpected bool
@@ -173,6 +174,19 @@ func TestSpecIsValid(t *testing.T) {
 			},
 			ErrorExpected: true,
 			Name:          "Invalid URL Image.Checksum",
+		},
+		{
+			Spec: Metal3MachineSpec{
+				Image: Image{
+					URL:        "http://172.22.0.1/images/rhcos.iso",
+					DiskFormat: &diskFormat,
+				},
+				UserData: &corev1.SecretReference{
+					Name: "worker-user-data",
+				},
+			},
+			ErrorExpected: false,
+			Name:          "Valid spec with live-iso diskFormat",
 		},
 	}
 


zhouhao3:
@kashifest @furkatgofurov7 @smoshiur1237 PTAL

zhouhao3:
PTAL

zhouhao3:
@kashifest @furkatgofurov7 @smoshiur1237 PTAL

zhouhao3:
@kashifest @furkatgofurov7 @smoshiur1237  PTAL

api/v1beta1/metal3machine_types.go
```diff
@@ -99,9 +99,11 @@ func (s *Metal3MachineSpec) IsValid() error {
 	if err != nil {
 		invalid = append(invalid, "Image.URL")
 	}
-	_, err = url.ParseRequestURI(s.Image.Checksum)
-	if err != nil {
-		invalid = append(invalid, "Image.Checksum")
+	if s.Image.DiskFormat == nil || *s.Image.DiskFormat != "live-iso" {
```
furkatgofurov7, line 102: I am not quite sure about this, if the DiskFormat is nil, do we even have to validate the image.Checksum?

kashifest:
What are we trying to achieve here, can you pleas give a bit more context of the issue?

zhouhao3:
>  What are we trying to achieve here, can you pleas give a bit more context of the issue?

Checksum can be ignored when the value of DiskFormat is live-iso.
I try to increase this part of the verification. Added comments to the code

api/v1beta1/metal3machine_types.go
zhouhao3: This is a bad code, I changed it, PTAL.

zhouhao3: I checked the document again, and the value of `DiskFormat` can be unset, and the value of `DiskFormat` is not set in the corresponding UT. The validation of nil is added here to avoid null pointer errors in *s.Image.DiskFormat.

api/v1beta1/metal3machine_types.go
```diff
@@ -87,7 +87,8 @@ func (s *Metal3MachineSpec) IsValid() error {
 	if s.Image.URL == "" {
 		missing = append(missing, "Image.URL")
 	}
-	if s.Image.Checksum == "" {
+	// Checksum is not required for live-iso.
+	if (s.Image.DiskFormat == nil || *s.Image.DiskFormat != "live-iso") && s.Image.Checksum == "" {
```
mboukhalfa, line 91: Is not `s.Image.DiskFormat == nil || *s.Image.DiskFormat != "live-iso" <=> *s.Image.DiskFormat != "live-iso" `since `s.Image.DiskFormat == nil `included in `*s.Image.DiskFormat != "live-iso"`

mboukhalfa:
cc @adilGhaffarDev @Xenwar

zhouhao3:
@kashifest @furkatgofurov7 Please see if there are any other questions.

api/v1beta1/metal3machine_types.go
zhouhao3: All I want to add is to skip Checksum validation when DiskFormat == live-iso. That is verify Checksum only when DiskFormat != live-iso.

api/v1beta1/metal3machine_types_test.go
```diff
@@ -174,6 +175,19 @@ func TestSpecIsValid(t *testing.T) {
 			ErrorExpected: true,
 			Name:          "Invalid URL Image.Checksum",
 		},
+		{
+			Spec: Metal3MachineSpec{
+				Image: Image{
+					URL:        "http://172.22.0.1/images/rhcos-ootpa-latest.qcow2",
```
adilGhaffarDev, line 181: Shouldn't we use live-iso image here?

furkatgofurov7:

/hold

api/v1beta1/metal3machine_types_test.go
zhouhao3: updated, thanks.

furkatgofurov7 (approved):

furkatgofurov7 (commented):
/approve

zhouhao3:
@mboukhalfa PTAL

api/v1beta1/metal3machine_types.go
```diff
@@ -99,9 +100,11 @@ func (s *Metal3MachineSpec) IsValid() error {
 	if err != nil {
 		invalid = append(invalid, "Image.URL")
 	}
-	_, err = url.ParseRequestURI(s.Image.Checksum)
-	if err != nil {
-		invalid = append(invalid, "Image.Checksum")
+	if s.Image.DiskFormat == nil || *s.Image.DiskFormat != "live-iso" {
+		_, err = url.ParseRequestURI(s.Image.Checksum)
```
schrej, line 104: This change requires the checksum to be a URL, which is a breaking change and not in line with the description of the field as it also allows to provide a checksum directly.

schrej: Also, why is this not checked on the webhook as well?

zhouhao3: > This change requires the checksum to be a URL

This is the original code,  I haven't changed the way `s.Image.Checksum` is validated.

schrej: You are right, the change was introduced with this PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/604
It is still breaking though, but was introduced with Release 1.1.2.

PR #619, opened 2022-05-06
:seedling: fix ControlPlaneEndpoint in metal3cluster_manager

zhouhao3:

Avoid returning err with nil in error cases


zhouhao3:
cc @kashifest @furkatgofurov7 @smoshiur1237

zhouhao3:
PTAL

Xenwar:
would it be possible to squash the commits

zhouhao3:
I think these two commits fix two different problems. Should not be squashed.

zhouhao3:
@furkatgofurov7 @smoshiur1237 PTAL

zhouhao3:
@kashifest @furkatgofurov7 @smoshiur1237 PTAL

furkatgofurov7 (commented):
Thanks!

/lgtm

PR #620, opened 2022-05-10
🌱 Add live ISO e2e test

adilGhaffarDev:
e2e test with live iso.

Discussion on slack regarding live iso:
https://kubernetes.slack.com/archives/CHD49TLE7/p1653654187046809

### DIFF
--- a/test/e2e/live_iso_test.go
+++ b/test/e2e/live_iso_test.go
@@ -1,0 +1,70 @@
+package e2e
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("When testing live iso [live-iso]", func() {
+	liveIsoTest()
+})
+
+func liveIsoTest() {
+	Logf("Starting live ISO test")
+	var (
+		liveISOImageURL = e2eConfig.GetVariable("LIVE_ISO_IMAGE")
+	)
+	BeforeEach(func() {
+		validateGlobals(specName)
+		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+	})
+	It("Should update the bmh with live iso", func() {
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+
+		bmhs, err := getAllBmhs(ctx, bootstrapClient, namespace, specName)
+		Expect(err).NotTo(HaveOccurred(), "Error getting BMHs")
+
+		isoBmh := bmhs[0]
+		Expect(isoBmh).ToNot(BeNil())
+
+		isoBmh.Spec.Image = &bmov1alpha1.Image{
+			URL:          liveISOImageURL,
+			Checksum:     "",
+			ChecksumType: "",
+			DiskFormat:   pointer.StringPtr("live-iso"),
+		}
+
+		Expect(bootstrapClient.Update(ctx, &isoBmh)).NotTo(HaveOccurred())
+
+		By("waiting for live ISO image booted host to be in provisioned state")
+		Eventually(func(g Gomega) {
+			g.Expect(bootstrapClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: bmhs[0].Name}, &isoBmh)).To(Succeed())
+			g.Expect(isoBmh.Status.Provisioning.State).To(Equal(bmov1alpha1.StateProvisioned))
+		}, e2eConfig.GetIntervals(specName, "wait-object-provisioned")...).Should(Succeed())
+
+		vmName := bmhToVMName(isoBmh)
+		serialLogFile := fmt.Sprintf("/var/log/libvirt/qemu/%s-serial0.log", vmName)
+
+		By("Reading serial logs to verify the node was booted from live ISO image")
+		Eventually(func(g Gomega) {
+			cmd := fmt.Sprintf("sudo cat %s | grep '#  Welcome'", serialLogFile)
+			output, err := exec.Command("/bin/sh", "-c", cmd).Output()
+			g.Expect(err).To(BeNil())
+			g.Expect(output).ToNot(BeNil(), fmt.Sprintf("Failed to read serial logs from %s", serialLogFile))
+		}, e2eConfig.GetIntervals(specName, "wait-job")...).Should(Succeed())
+		By("LIVE ISO TEST PASSED!")
+	})
+
+	AfterEach(func() {
+		dumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, artifactFolder, namespace, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
+	})
+}
--- a/docs/e2e-test.md
+++ b/docs/e2e-test.md
@@ -54,6 +54,7 @@ The e2e tests currently include three different sets:
 1. pivoting_based_feature_tests
 1. remediation_feature_tests
 1. upgrade_tests
+1. live_iso_tests
 
 `pivoting_based_feature_tests`: Because these tests run mainly in the target cluster, they are dependent on the pivoting test and need to run in the following order:
 
@@ -80,6 +81,8 @@ The remediation and inspection tests are then run with the controllers still in 
 - Upgrade Ironic
 - Upgrade CAPI/CAPM3
 
+`live_iso_test`: independent from the previous tests and can run independently. This is testing the booting of target cluster's nodes with the live ISO.
+
 Guidelines to follow when adding new E2E tests:
 
 - Tests should be defined in a new file and separate test spec, unless the new test depends on existing tests.
@@ -87,6 +90,7 @@ Guidelines to follow when adding new E2E tests:
 
 `[upgrade]` => runs only existing upgrade tests including CAPI, CAPM3, Ironic and Baremetal Operator.
 `[remediation]` => runs only remediation and inspection tests.
+`[live-iso]` => runs only live ISO test.
 
 For instance, to skip the upgrade E2E tests set `GINKGO_SKIP="[upgrade]"`
 
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -72,5 +72,8 @@ else
   export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
 fi
 
+# image for live iso testing
+export LIVE_ISO_IMAGE="https://artifactory.nordix.org/artifactory/metal3/images/iso/minimal_linux_live-v2.iso"
+
 # run e2e tests
 make e2e-tests


adilGhaffarDev:
test-v1b1-e2e-ubuntu

scripts/environment.sh
```diff
@@ -83,6 +83,8 @@ else
   export UPGRADED_RAW_IMAGE_NAME="CENTOS_9_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
 fi
 
+export LIVE_ISO_IMAGE="https://artifactory.nordix.org/artifactory/metal3/images/iso/minimal_linux_live-v2.iso"
```
mboukhalfa, line 86: I would prefer this to be in e2e-config file since it is needed only by the e2e. I am usually export vars here only if it is necessary but if the var is used only by e2e framework would be better to be in the e2e file

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var _ = Describe("When testing live iso [live-iso]", func() {
+	liveIsoTest()
+})
+
+func liveIsoTest() {
+	It("Should create a target cluster nodes with live iso", func() {
+		controlPlaneMachineCount = int64(*e2eConfig.GetInt32PtrVariable("CONTROL_PLANE_MACHINE_COUNT"))
+		workerMachineCount = int64(*e2eConfig.GetInt32PtrVariable("WORKER_MACHINE_COUNT"))
```
mboukhalfa, line 17: I think only 1 worker should be enough or even spinning only the kcp

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,45 @@
...
+			WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
+			WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
+			WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
+		}, result)
+		targetCluster = bootstrapClusterProxy.GetWorkloadCluster(ctx, namespace, clusterName)
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+		controlplaneM3Machines, workerM3Machines := getMetal3Machines(ctx, bootstrapClient, clusterName, namespace)
+		Expect(controlplaneM3Machines).To(HaveLen(int(controlPlaneMachineCount)))
+		Expect(workerM3Machines).To(HaveLen(int(workerMachineCount)))
+	})
```
mboukhalfa, line 44: add check here by reading `/var/log/libvirt/qemu/"{{ node name }}-serial0.log` and verify it contains `"#  Welcome" `

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,103 @@
...
+				for _, machine := range machineList.Items {
+					if isWorkerNode(machine) && strings.EqualFold(machine.Status.Phase, "provisioning") {
+						n++
+					} else if strings.EqualFold(machine.Status.Phase, "running") {
+						n++
+					}
+				}
+			}
+			return n, nil
+		}, e2eConfig.GetIntervals(specName, "wait-worker-nodes")...).Should(Equal(2))
```
mboukhalfa, line 66: Since you are only verifying the kcp machine is running I believe we can simplify the condition and remove the fuction isWorker as follow:
```suggestion
		Eventually(func() (bool) {
			machineList := &clusterv1alpha4.MachineList{}
			if err := bootstrapClusterProxy.GetClient().List(ctx, machineList, client.InNamespace(namespace), client.MatchingLabels{clusterv1.ClusterLabelName: clusterName}); err == nil {
				for _, machine := range machineList.Items {
					if strings.EqualFold(machine.Status.Phase, "running") {
						return true
					}
				}
			}
			return false
		}, e2eConfig.GetIntervals(specName, "wait-worker-nodes")...).To(Succeed())
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+		getBmhFromM3Machine := func(m3Machine infrav1.Metal3Machine) (result bmov1alpha1.BareMetalHost) {
+			Expect(bootstrapClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: metal3MachineToBmhName(m3Machine)}, &result)).To(Succeed())
+			return result
+		}
+
+		workerBmh := getBmhFromM3Machine(workerM3Machines[0])
+		vmName := bmhToVMName(workerBmh)
+		serialLogFile := fmt.Sprintf("/var/log/libvirt/qemu/%s-serial0.log", vmName)
+		time.Sleep(5 * time.Minute)
```
mboukhalfa, line 80: Why sleeping here ? can replace that directly with a loop check is the logs reached what we are expecting ?

mboukhalfa (commented):
looks good check some suggestions

test/e2e/live_iso_test.go
adilGhaffarDev: I tried the test by removing the check for worker node, but it started failing when I tried to get worker node after this loop, so I think we have to keep it there. Let me know if it can be improved in some other way.

Makefile
```diff
@@ -161,6 +162,8 @@ else
 GINKGO_SKIP := upgrade
 endif
 
+GINKGO_FOCUS := live-iso
```
mboukhalfa, line 165: /hold we cannot merge that else the other tests will not get triggered since ginkgo here focusing only on live-iso without condition, we should wait until the jjb https://gerrit.nordix.org/c/infra/cicd/+/13707 get merged then remove this line

test/e2e/live_iso_test.go
mboukhalfa: can you please describe what was the issue ?

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,95 @@
...
+			LogFolder:                clusterctlLogFolder,
+		})
+
+		Expect(workloadClusterTemplate).ToNot(BeNil(), "Failed to get the cluster template")
+		Logf("Applying the cluster template yaml to the cluster")
+		Expect(bootstrapClusterProxy.Apply(ctx, workloadClusterTemplate)).To(Succeed())
+
+		var controlplaneM3Machines, workerM3Machines []infrav1.Metal3Machine
+
+		By("Waiting for the control plane and worker node to exist")
```
lentzi90, line 52:
```suggestion
		By("Waiting for the control plane and worker metal3machines to exist")
```

lentzi90 (commented):
What is the reason we involve a full target cluster here? If I understand correctly we only check that the BMH can boot from the live-iso. Not that it can form a cluster (it probably cannot?). So I'm wondering about what the goal is.

Do we want to make sure that live-iso an be used also on the CAPM3 side (not just BMO) or was it just the easiest way to implement it since it is similar to other tests? It would be good to add some comment about this I think. The way it is now it can be quite confusing when seeing IP pools, metal3machines and clusters if we only check the boot log.

Alternatively, we could just apply/patch a BMH to use live-iso to boot and be done with it.

test/e2e/live_iso_test.go
adilGhaffarDev: I would like to have @Xenwar input on this. To my knowledge for this test in cluster we need at least one node to be using live iso.

Xenwar: I agree with @lentzi90 on
`Alternatively, we could just apply/patch a BMH to use live-iso to boot and be done with it.`
But, does the e2e code allows us to just provisioning a node, without setting up a cluster ?

lentzi90: I think it should be possible. What I can see now in the live-iso test we start by creating a target cluster. So this means the bootstrap cluster is already there and it should also have the BMHs in available state.
Now we should be able to manipulate the BMH in a similar way like it is done in the remediation test where they [add annotations](https://github.com/metal3-io/cluster-api-provider-metal3/blob/548790c0acba5c10cdaf537e1e7fdbd9e80501ca/test/e2e/common.go#L109).

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+	})
+	It("Should update the bmh with live iso", func() {
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+
+		bmhs, err := getAllBmhs(ctx, bootstrapClient, namespace, specName)
+		Expect(err).NotTo(HaveOccurred(), "Error getting bmhs")
+
+		var isoBmh bmov1alpha1.BareMetalHost
+		Expect(bootstrapClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: bmhs[0].Name}, &isoBmh)).To(Succeed())
```
lentzi90, line 36: I think we can simplify this a bit.
```suggestion
		isoBmh := bmhs[0]
```
Maybe add an Èxpect` to check that it is not `nil`. But the point is that we don't have to fetch the BMH twice.

lentzi90 (approved):
Great job!
One small nit below, otherwise I'm happy with this.

lentzi90 (approved):
LGTM!

furkatgofurov7 (commented):
@adilGhaffarDev can you squash your commits before merging?

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+	. "github.com/onsi/gomega"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("When testing live iso [live-iso]", func() {
+	liveIsoTest()
+})
+
+func liveIsoTest() {
```
furkatgofurov7, line 20: could you add a log indicating live ISO tests starting to keep it consistent with other tests. You can check this as an example: https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/pivoting_test.go#L33

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+		isoBmh.Spec.Image = &bmov1alpha1.Image{
+			URL:          liveIsoImageURL,
+			Checksum:     "",
+			ChecksumType: "",
+			DiskFormat:   pointer.StringPtr("live-iso"),
+		}
+
+		Expect(bootstrapClient.Update(ctx, &isoBmh)).NotTo(HaveOccurred())
+
+		By("waiting for live iso host to be in provisioned state")
```
furkatgofurov7, line 47:
```suggestion
		By("waiting for live ISO image booted host to be in provisioned state")
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+		Eventually(func(g Gomega) {
+			g.Expect(bootstrapClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: bmhs[0].Name}, &isoBmh)).To(Succeed())
+			g.Expect(isoBmh.Status.Provisioning.State).To(Equal(bmov1alpha1.StateProvisioned))
+		}, e2eConfig.GetIntervals(specName, "wait-machine-remediation")...).Should(Succeed())
```
furkatgofurov7, line 51: I would suggest to use: `wait-object-provisioned`
```suggestion
		}, e2eConfig.GetIntervals(specName, "wait-object-provisioned")...).Should(Succeed())
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+		vmName := bmhToVMName(isoBmh)
+		serialLogFile := fmt.Sprintf("/var/log/libvirt/qemu/%s-serial0.log", vmName)
+
+		By("Reading serial logs to verify live iso")
```
furkatgofurov7, line 56:
```suggestion
		By("Reading serial logs to verify the node was booted from live ISO image")
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("When testing live iso [live-iso]", func() {
+	liveIsoTest()
+})
+
+func liveIsoTest() {
+	var (
+		liveIsoImageURL = e2eConfig.GetVariable("LIVE_ISO_IMAGE")
```
furkatgofurov7, line 22: I would stick to capitalizing the acronym (ISO) in the code as it is a best practice in Go. 
```suggestion
		liveISOImageURL = e2eConfig.GetVariable("LIVE_ISO_IMAGE")
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+
+		bmhs, err := getAllBmhs(ctx, bootstrapClient, namespace, specName)
+		Expect(err).NotTo(HaveOccurred(), "Error getting bmhs")
+
+		isoBmh := bmhs[0]
+		Expect(isoBmh).ToNot(BeNil())
+
+		isoBmh.Spec.Image = &bmov1alpha1.Image{
+			URL:          liveIsoImageURL,
```
furkatgofurov7, line 39:
```suggestion
			URL:          liveISOImageURL,
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+	BeforeEach(func() {
+		validateGlobals(specName)
+		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+	})
+	It("Should update the bmh with live iso", func() {
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+
+		bmhs, err := getAllBmhs(ctx, bootstrapClient, namespace, specName)
+		Expect(err).NotTo(HaveOccurred(), "Error getting bmhs")
```
furkatgofurov7, line 33:
```suggestion
		Expect(err).NotTo(HaveOccurred(), "Error getting BMHs")
```

test/e2e/live_iso_test.go
```diff
@@ -0,0 +1,69 @@
...
+		vmName := bmhToVMName(isoBmh)
+		serialLogFile := fmt.Sprintf("/var/log/libvirt/qemu/%s-serial0.log", vmName)
+
+		By("Reading serial logs to verify live iso")
+		Eventually(func(g Gomega) {
+			cmd := fmt.Sprintf("sudo cat %s | grep '#  Welcome'", serialLogFile)
+			output, err := exec.Command("/bin/sh", "-c", cmd).Output()
+			g.Expect(err).To(BeNil())
+			g.Expect(output).ToNot(BeNil(), fmt.Sprintf("Failed to read serial logs from %s", serialLogFile))
+		}, e2eConfig.GetIntervals(specName, "wait-worker-nodes")...).Should(Succeed())
```
furkatgofurov7, line 62: using `wait-worker-nodes` is confusing here, is not it? how about using: `wait-service` or `wait-job` if the task can be completed in 10 minutes?

furkatgofurov7 (changes requested):
This is great, few suggestions inline:

furkatgofurov7:
> test-v1b1-ubuntu-e2e-parallel

this is not triggered for some reason. 

/test-v1b1-ubuntu-e2e-parallel

furkatgofurov7:
/retitle 🌱 Add live ISO e2e test

kashifest:
@dtantsur since this test was requested by you, and its passing all the tests , if its looks good to you just unhold the PR.

/hold

kashifest:
/hold cancel
Taking this in since there was no comment/objection from @dtantsur

dtantsur:
Late to the party, but looks good to me, thanks!

PR #635, opened 2022-06-06
 🌱 Downgrade to Ubuntu 20.04 and Centos 8 

kashifest:
**Downgrade to Ubuntu 20.04**
Ubuntu 22.04 is facing issues in network configurations where provisioning bridge is unable to communicate over provisioning bridge in target cluster and this is blocking the feature/e2e tests. We will fall back to 20.04 until we investigate and find a solution for this.

**Downgrade to Centos stream 8**
We have a suspicion that after upgrading to Centos 8 stream we started seeing the client loop disconnect issue too many times in the CI. We will temporarily switch back to Centos 8 and investigate more until we find an asnwer


smoshiur1237 (approved):
LGTM

furkatgofurov7 (approved):

PR #636, opened 2022-06-06
🐛 Fix e2e pivoting by adding capi labels on hardware data crd

mboukhalfa:
Add CAPI labels on hardwareData CRD (added in https://github.com/metal3-io/baremetal-operator/pull/1099) so that hadwareData object is also pivoted as part of clusterctl move operation. Ref: https://github.com/metal3-io/metal3-dev-env/pull/969

### DIFF
--- a/test/e2e/pivoting_test.go
+++ b/test/e2e/pivoting_test.go
@@ -67,6 +67,8 @@ func pivoting() {
 
 	By("Add labels to BMO CRDs")
 	labelBMOCRDs(nil)
+	By("Add Labels to hardwareData CRDs")
+	labelHDCRDs(nil)
 
 	By("Install BMO")
 	installIronicBMO(targetCluster, "false", "true")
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
looks good to me. Although I noticed that for BMH we are not setting move label but rather provider name, which should be okay.
```
"clusterctl.cluster.x-k8s.io=",
"cluster.x-k8s.io/provider=metal3",
```

lentzi90 (commented):
Nice find @mboukhalfa ! Thanks for fixing.

I see you are testing something with nft, are you seeing issues in the tests or what is this for?

scripts/ci-e2e.sh
```diff
@@ -71,7 +71,7 @@ else
   export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
   export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
 fi
-
+sudo nft flush ruleset
```
kashifest, line 74: why do we need this?

Xenwar: could this be one of the commands to fix the broken pipe issue ?
I do not think we need it in this PR.

mboukhalfa:
/hold cancel 
Merging this PR to fix e2e centos test
e2e-centos tested but we will not run ubuntu due to the bmo/ironic issue that will be fixed afterwards

mboukhalfa:
> Nice find @mboukhalfa ! Thanks for fixing.
> 
> I see you are testing something with nft, are you seeing issues in the tests or what is this for?

Was trying to debug ironic bmo communication issue on ubuntu those are removed now

PR #640, opened 2022-06-14
Update template with the new providerID field

Xenwar:
Update template with the new providerID field


test/e2e/data/infrastructure-metal3/bases/centos-kubeadm-config/centos-kubeadm-config.yaml
```diff
@@ -126,7 +126,11 @@ spec:
           container-runtime-endpoint: unix:///var/run/crio/crio.sock
           feature-gates: AllAlpha=false
           node-labels: metal3.io/uuid={{ ds.meta_data.uuid }}
-          provider-id: metal3://{{ ds.meta_data.uuid }}
+{% if CAPM3_VERSION == "v1alpha5" %}
+          provider-id: "metal3://{{ '{{ ds.meta_data.uuid }}' }}"
+{% else %}
+          provider-id: "metal3://{{ '{{ ds.meta_data.providerid }}' }}"
+{% endif %}
```
mboukhalfa, line 133: I do not think the `if else` will be interpreted here, else I would suggest to have the logic somewhere in the code and exports the results to a unique var that should be replaced here just like we are doing with `capi_version` for example

scripts/ci-e2e.sh
```diff
@@ -71,6 +71,11 @@ else
   export CAPM3_FROM_RELEASE="${CAPM3_FROM_RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v0.5.")}"
   export CAPM3_TO_RELEASE="${CAPM3RELEASE}"
 fi
+if [[ $CAPM3_VERSION == "v1alpha5" ]]; then
+  export PROVIDER_ID_FORMAT="metal3://{{ '{{ ds.meta_data.uuid }}' }}"
+else
+  export PROVIDER_ID_FORMAT="metal3://{{ '{{ ds.meta_data.providerid }}' }}"
+fi
```
mboukhalfa, line 78: I would prefer to put this in environment.sh

Xenwar: Thanks, moved it to environment.sh

zhouhao3:
LGTM

kashifest (approved):

PR #641, opened 2022-06-15
WIP 🌱 Adding bmh from the cluster template

mboukhalfa:
wip


furkatgofurov7:

/hold

test/e2e/data/infrastructure-metal3/cluster-template-ubuntu/kustomization.yaml
```diff
@@ -1,3 +1,7 @@
 resources:
+- ../bases/bmhs/node-0-bmc-secret.yaml
+- ../bases/bmhs/node-0.yaml
+- ../bases/bmhs/node-1-bmc-secret.yaml
+- ../bases/bmhs/node-1.yaml
```
mboukhalfa, line 5: we need a way to call dynamic resources the number of bmh nodes is not fixed. The best way would be to load what ever files available in bmhs folder

Rozzii:
/hold
This is titled as WIP but at some point lost the "hold" tag I am putting that back.

PR #644, opened 2022-06-17
🐛 Move dev-env cleaning to the makefile so it get triggered from the pipeline cleaning stage

mboukhalfa:
Currently the e2e CI tarball does not contain the serial-logs because they got cleaned before the logs collection stage.
This PR fix this issue by making the dev-env cleaning triggered from the pipeline cleaning stage after collecting the logs


mboukhalfa:
/test-v1b1-ubuntu-integration
/assign @kashifest 
cc @adilGhaffarDev

kashifest (approved):

PR #648, opened 2022-06-21
Back-port to v1alpha5 forward

Xenwar:
On v1alpha5, we noticed some tests fail due to providerID related issues, which are usually visible on upgrade and node reuse tests. 

In the main branch, we already have a fix that solves such problems. The purpose of this PR is to back port the fix from main to v1alpha5. 

Co-authored-by: [furkatgofurov7](https://github.com/furkatgofurov7) | <furkat.gofurov@est.tech>

### DIFF
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -41,12 +41,6 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-const (
-	m3machine   = "metal3machine"
-	host        = "baremetalhost"
-	capimachine = "machine"
-)
-
 // DataManagerInterface is an interface for a DataManager
 type DataManagerInterface interface {
 	SetFinalizer()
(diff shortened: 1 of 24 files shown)


Xenwar:
Note, the same tests should run with the legacy providerID as well. 
Then, after fixing any issues with both formats, the providerID can be parameterized.

Xenwar:
All tests have passed. 
I am not pushing a new commit to verify the legacy format works as usual.

Xenwar:
Upon a successful run or multiple runs, please remove the following two commits before merging. 
```
f5c31171759e5d0b4dd538232a09be7712b48ba2
3088026c34432d250935a7a54438a392ead7ba36
```

furkatgofurov7:
This is a breaking change, so we won't be backporting it to release branch. 

/close

PR #650, opened 2022-06-22
🐛 Validate providerID for proper prefix

Xenwar:
Validate providerID for proper prefix


baremetal/metal3machine_manager.go
```diff
@@ -1326,6 +1326,9 @@ func (m *MachineManager) GetProviderIDAndBMHID() (string, *string) {
 	bmhID := *providerID
 	if strings.Contains(bmhID, ProviderIDPrefix) {
 		bmhID = strings.TrimPrefix(bmhID, ProviderIDPrefix)
+	} else {
```
Xenwar, line 1329: This would not solve the problem we are facing. The prefix was there and was trimmed correctly.
The issue seems to be related to failure to identify the new and legacy formats.
Closing this issue.

Issue #392, opened 2021-11-18
Reboot remediation is incomplete


slintes:
**What steps did you take and what happened:**
The reboot remediation currently does not delete the related Node, which prevents the cluster from moving workloads from uhealthy nodes to healthy nodes in time.

**What did you expect to happen:**
Node deletion & quick workload relocation.

**Anything else you would like to add:**
See comments here: 
https://github.com/metal3-io/metal3-docs/pull/118#issuecomment-655326761
https://github.com/metal3-io/cluster-api-provider-metal3/pull/367#issuecomment-967150207

The proposal for[ adding the reboot remediation](https://github.com/metal3-io/metal3-docs/blob/master/design/cluster-api-provider-metal3/capm3-remediation-controller-proposal.md) does not specify how remediation should look like exactly. This leaves room for discussion if the current implementtion has a bug, or if adding node deletion is a new feature. That also results in discussion if node deletion should be added to the existing implementation, or if the fix should be implemented as a new remediation strategy. While [on Slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1634217997210700?thread_ts=1634199536.209500&cid=CHD49TLE7) the consensus was extending the existing remediation, the[ implementation PR](https://github.com/metal3-io/cluster-api-provider-metal3/pull/367) triggered new discussion... 🤷🏼‍♂️  

/kind bug

macaptain:
To summarize the reasons for this change given in other discussions:

> if we don't delete the node the total downtime for that workload would be:
> time to power off + time to power on + time for a node to come up + ERC timeout + time for Machine deletion (5 minutes) =~ > 20 minutes
> Comparing to node deletion where the downtime would be - time to power off + time to delete the node =~ 10 seconds

So, looks like fixing this will reduce downtime for workloads which were on the unhealthy BMH.

> The purpose of the 'reboot' remediation strategy is to allow us to not delete the Machine (which would mean reprovisioning the Host) and instead just reboot the Host. Because the Machine itself remains, this means that the CAPI will not delete the Node for us, which is an essential part of remediation, and thus the remediation controller (all remediation controllers) must do it.

Furthermore, since we use a custom remediation controller to override CAPI's usual behavior to delete the Machine (which in turn deletes the node), we should be deleting the node ourselves since CAPI doesn't do that for us.

zaneb:
[...]

PR #367, opened 2021-11-06
:sparkles: Add node deletion to reboot remediation

slintes:
**What this PR does / why we need it**:

This PR adds node deletion to the reboot remediation strategy, because it ensures that unhealthy nodes are fenced, and workloads will be rescheduled to other nodes.
This was proposed on Slack, and it was agreed to add this to the existing reboot remediation strategy, instead of adding a new strategy.

Unfortunately this increases complexity of remediation, because instead of having 1 step only (set reboot annotation), we now have several steps, plus waiting for certain cluster conditions:

1. request power off and wait until done
2. Backup node annotations and labels
3. Delete node and wait until gone
4. Power on and wait until done
5. Restore node annottaions and labels

This is WIP, tests need to be updated / extended. Initial review would be welcome though, to see if this is going into an approvable direction, or something different is expected.

PR #668, opened 2022-07-11
🐛 Fix reboot remediation by adding node deletion

slintes:

Note: Replaces https://github.com/metal3-io/cluster-api-provider-metal3/pull/367 because of broken CI. See discussions over there.

This PR adds node deletion to the reboot remediation strategy, because it ensures that unhealthy nodes are fenced, and workloads will be rescheduled to other nodes.
This was proposed on Slack, and it was agreed to add this to the existing reboot remediation strategy, instead of adding a new strategy.

Unfortunately this increases complexity of remediation, because instead of having 1 step only (set reboot annotation), we now have several steps, plus waiting for certain cluster conditions:

- Request power off and wait until done
- Backup node annotations and label
- Delete node and wait until gone
- Power on and wait until done
- Restore node annotations and labels

Fixes https://github.com/metal3-io/cluster-api-provider-metal3/issues/392


mboukhalfa:
/lgtm 
cc @furkatgofurov7

furkatgofurov7:
> Actually, I don't see any details there about how remediation will happen, it jut says "RC uses BMO APIs to get hosts back into a healthy or manageable state"

> If you like me to add implementation details (since there none yet I guess a new proposal does not make sense), I can do that as follow up.

@slintes as agreed, we will look forward to follow-up PR(s) (completely up to you, either single PR updating both or separate for each) with implementation details in the:
1. already merged proposal: https://github.com/metal3-io/metal3-docs/blob/main/design/cluster-api-provider-metal3/capm3-remediation-controller-proposal.md
2. as well as a user guide for the remediation feature here: https://github.com/metal3-io/metal3-docs/blob/main/docs/user-guide/src/capm3/remediaton.md 

since both need to be synced to avoid confusion in the future for the consumers/users. Meanwhile, I am merging this, thanks for all your hard work :tada: 

/approve

PR #670, opened 2022-07-14
🐛 Fix remediation cleanup

slintes:
The updated remediation flow introduced by https://github.com/metal3-io/cluster-api-provider-metal3/pull/668 has an issue, the remediations CRs can't be deleted because the finalizer isn't removed.

The fix is: Initialize patch helper after getting current CR!!! Otherwise nothing will deleted by patch, only added.

Some more improvements done along the way deugging this:

- Requeue CR when patch fails
- CRs with deletion timestamp are handled in normal CR flow (finalizer needs to be removed)
- add CR deleted check to e2e test
- slightly better logging
- use controller utils for finalizer handling

### DIFF
--- a/baremetal/metal3remediation_manager.go
+++ b/baremetal/metal3remediation_manager.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


furkatgofurov7 (approved):
/approve
/lgtm

Taking this in since the main [feature tests](https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_daily_main_feature_tests_ubuntu/) are broken and this should fix it

PR #676, opened 2022-08-02
:bug: Fix initial target cluster K8s version

Rozzii:
Until the e2e upgrade test starts with CAPI 0.4.x + CAPM 0.5.x + K8s 1.23.x it is not possible to set
the starting version of the target cluster (FROM_K8S_VERSION) higher than 1.23.x

### DIFF
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -48,7 +48,11 @@ else
   export EPHEMERAL_CLUSTER="minikube"
 fi
 
-export FROM_K8S_VERSION="v1.24.0"
+# Until the upgrade process starts with CAPI 0.4.x
+# CAPM 0.5.x and K8s 1.23.x it is not possible to set
+# the starting version of the target cluster (FROM_K8S_VERSION)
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
I might be asking naive question, but how did upgrade run until now with 1.24.0?

fmuyassarov:
> I might be asking naive question, but how did upgrade run until now with 1.24.0?

Do'h, probably it never passed and this is the fix for that?

mboukhalfa:
> > I might be asking naive question, but how did upgrade run until now with 1.24.0?
> 
> Do'h, probably it never passed and this is the fix for that?

Yeah the test start failing after the commit https://github.com/metal3-io/cluster-api-provider-metal3/commit/b4f8a4b93d2d73c27ad46ade0c748e7272406d9d that uplifted `FROM_K8S_VERSION` from 1.23.5 to 1.24.0

fmuyassarov (approved):

PR #674, opened 2022-07-21
🌱Uplift capi to v1.2

adilGhaffarDev:
Follwing go modules uplifted
- sigs.k8s.io/controller-runtime: v0.11.x => v0.12.3
- sigs.k8s.io/controller-tools: v0.8.x => v0.9.1
- sigs.k8s.io/kind: v0.11.x => v0.14.0
- k8s.io/*: v0.23.5 => v0.24.3 (derived from controller-runtime)
- github.com/onsi/gomega: v0.17.0 => v0.18.1 (derived from controller-runtime)
- k8s.io/kubectl: v0.23.5 => 0.24.0

PR #677, opened 2022-08-05
🌱 update golangci-lint to 1.47.3

schrej:
Updates golangci-lint to version 1.47.3
The previous version was leaking memory, maybe because of Go 1.19. The issue seems to be resolved with version 1.47. Since the latest version requires Go 1.19 we should probably wait until the main package switches to that as well.
Changelog: https://github.com/golangci/golangci-lint/releases


schrej:
already done in #674 
/close

Issue #671, opened 2022-07-14
Add Support for CAPI IPAM Contract


schrej:
**User Story**

As a operator I would like to use the new CAPI IPAM contract with metal3 in order to integrate with different IPAM solutions.

**Detailed Description**

The CAPI IPAM contract is now [implemented](https://github.com/kubernetes-sigs/cluster-api/pull/6313) and released as part of CAPI 1.2.0-rc.0.

In order to support it, we'll need to be able to reference IP Pools implemented by various providers. Looking at the current API, I think we can easily do so by extending [`FromPool`](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/api/v1beta1/metal3datatemplate_types.go#L128-L134) in the DataTemplate with `apiGroup` and `kind` parameters. The new fields should either default to the metal3-ip-address-manager types, or just stay empty. A new reference would then look like this:
```yaml
- key: "something"
  name: "some-pool"
  apiGroup: "ipam.cluster.x-k8s.io"
  kind: "InClusterIPPool"
```
Since only optional fields are added, the change would be fully backwards compatible and doesn't require a new API version.

In code we can then differenciate based on `apiGroup` and `kind`, and either create a metal3-ip-address `IPAddressClaim`, or a CAPI one (ipam.cluster.x-k8s.io).

**Anything else you would like to add:**

I'm happy to work on this.

/kind feature

Rozzii:
/triage accepted
Thank you for opening this issue @schrej we would really like to include this feature!

fmuyassarov:
As discussed in the last community meeting, I'm +1 for the change.
For the new fields discussed in the description, perhaps defaulting them to metal3 ipam types would help to understand it better which ipam is being used instead of leaving it empty and assuming that metal3 ipam is in use.
/cc @kashifest any thoughts on the addition? I don't see any blockers and @schrej is willing to contribute.

kashifest:
>  /cc @kashifest any thoughts on the addition? I don't see any blockers and @schrej is willing to contribute.

+1 for this. @schrej Thanks a lot for working on this.

schrej:
The current IPAM contract does not support specifying DNS servers. What was the reason to specify that on the IPPool/IPAddress resources? Do we still need that?

From what I can see, the current implementation would allow to reference a pool to get a Gateway from, even if the server machine has no interface that references the same pool. An IP would still be allocated from the pool, and only the Gateway would be used. Same for DNS.
[...]

PR #678, opened 2022-08-05
🌱 share logic between ip address allocation and release

schrej:
clean up return parameters and update comments to reflect current behaviour

I've refactored the IP address allocation and release functions to share the logic used to determine from which pools to allocate.

I've also cleaned up a few things in the process
- `releaseAddressFromM3Pool` no longer returns whether it was successful or if reconciliation should be requeued, both were unused
- both `getAddressFromM3Pool` and `releaseAddressFromM3Pool` no longer interact with the addresses map, as that was not necessary. They also no longer check whether a pool was processed already, as each pool will only be processed once.

This is in preparation for implementing CAPI IPAM Contract support (#671), so I've renamed `getAddressFromM3Pool` and `releaseAddressFromM3Pool` to include `M3`.
I've also used a `corev1.TypedLocalObjectReference` because I plan to use that to differentiate between pools. If both group and kind are unset, or set to M3 IPAM, the current logic will be used. If a different group/kind is referenced, the new IPAM contract will be used instead.

### DIFF
--- a/baremetal/metal3data_manager_test.go
+++ b/baremetal/metal3data_manager_test.go
@@ -1102,16 +1102,16 @@ var _ = Describe("Metal3Data manager", func() {
 	)
 
 	type testCaseGetAddressFromPool struct {
-		m3d               *infrav1.Metal3Data
-		poolName          string
-		poolAddresses     map[string]addressFromPool
-		ipClaim           *ipamv1.IPClaim
-		ipAddress         *ipamv1.IPAddress
-		expectError       bool
-		expectRequeue     bool
-		expectedAddresses map[string]addressFromPool
-		expectDataError   bool
-		expectClaim       bool
+		m3d             *infrav1.Metal3Data
+		poolName        string
+		poolRef         corev1.TypedLocalObjectReference
+		ipClaim         *ipamv1.IPClaim
+		ipAddress       *ipamv1.IPAddress
+		expectError     bool
+		expectRequeue   bool
+		expectedAddress addressFromPool
+		expectDataError bool
+		expectClaim     bool
 	}
 
 	DescribeTable("Test GetAddressFromPool",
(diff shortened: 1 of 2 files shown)


lentzi90 (approved):
Awesome! I was struggling with this code just the other day, getting confused by the comments that didn't reflect the code... Very nice refactor also.

baremetal/metal3data_manager_test.go
```diff
@@ -1307,13 +1288,11 @@ var _ = Describe("Metal3Data manager", func() {
 	)
 
 	type testCaseReleaseAddressFromPool struct {
```
kashifest, line 1290: now that we dont have expectAddresses and ReleaseAddressFromPool is only returning errors, these testcases are not making so much sense because we are not checking any negative testcases. I would expect a testcase here where expectError is true and probably we should mention the false cases also explicitly.

baremetal/metal3data_manager_test.go
```diff
@@ -1353,39 +1326,19 @@ var _ = Describe("Metal3Data manager", func() {
...
 		},
-		Entry("Already processed", testCaseReleaseAddressFromPool{
-			m3d: &infrav1.Metal3Data{
-				ObjectMeta: testObjectMeta(metal3DataName, namespaceName, ""),
-			},
-			poolName: testPoolName,
-			poolAddresses: map[string]bool{
-				testPoolName: true,
-			},
-			expectedAddresses: map[string]bool{
```
kashifest, line 1364: I am thinking whats the harm if we keep this test, with the current code would it be a no-op ?

kashifest (commented):
/approve
two comments in line.
@furkatgofurov7 PTAL

baremetal/metal3data_manager_test.go
schrej: It does not happen anymore. Each pool will only be processed once.

baremetal/metal3data_manager_test.go
schrej: Testing negative cases requires to inject errors into the api client, which fakeclient does not support. I think I've wrapped it to do so in the past, let me see if I can find that.

lentzi90: Maybe using mocks? Like here in the [metal3machine_controller_test.go](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/controllers/metal3machine_controller_test.go#L94)

baremetal/metal3data_manager_test.go
kashifest: ok. thanks for clarifying. Marking this comment as resolved.

baremetal/metal3data_manager_test.go
schrej: I've cleaned up the tests now and added one that injects an error on deletion.

furkatgofurov7 (commented):
Thanks for the great cleanup @schrej 

/lgtm

PR #685, opened 2022-08-16
🌱 Use finalizer on IPClaim to avoid premature deletion

lentzi90:

- Add a finalizer on IPClaims to protect them from accidental deletion while still in use.
- Also remove the owner reference on the IPClaim before marking for deletion so that IPAM can use it to determine if the IPClaim is still in use or not.
- Edited some confusing comments

I'm a bit unsure if this should be considered a "feature" or something else. Put 🌱 "other" for now.

### DIFF
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -205,7 +205,7 @@ func (m *DataManager) createSecrets(ctx context.Context) error {
 		return nil
 	}
 
-	// Fetch all the Metal3IPPools and set the OwnerReference. Check if the
+	// Fetch all the Metal3IPPools and create Metal3IPClaims as needed. Check if the
 	// IP address has been allocated, if so, fetch the address, gateway and prefix.
 	poolAddresses, err := m.getAddressesFromPool(ctx, *m3dt)
 	if err != nil {
@@ -448,8 +448,9 @@ func (m *DataManager) getAddressFromM3Pool(ctx context.Context, poolRef corev1.T
 		// Create the claim
 		ipClaim = &ipamv1.IPClaim{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      m.Data.Name + "-" + poolRef.Name,
-				Namespace: m.Data.Namespace,
+				Name:       m.Data.Name + "-" + poolRef.Name,
+				Namespace:  m.Data.Namespace,
+				Finalizers: []string{infrav1.DataFinalizer},
 				OwnerReferences: []metav1.OwnerReference{
 					{
 						APIVersion: m.Data.APIVersion,
@@ -478,7 +479,7 @@ func (m *DataManager) getAddressFromM3Pool(ctx context.Context, poolRef corev1.T
 	}
 
 	if !ipClaim.DeletionTimestamp.IsZero() {
-		// This IPClaim is about to be deleted so we cannot use it. Requeue.
+		// This Metal3IPClaim is about to be deleted so we cannot use it. Requeue.
 		m.Log.Info("Found IPClaim with deletion timestamp, requeuing.", "IPClaim", ipClaim)
 		return addressFromPool{}, true, nil
 	}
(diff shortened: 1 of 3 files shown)


lentzi90:
e2e test broken, see https://github.com/metal3-io/cluster-api-provider-metal3/issues/689

baremetal/metal3data_manager.go
```diff
@@ -659,8 +659,8 @@ func (m *DataManager) releaseAddressesFromPool(ctx context.Context,
 	return nil
 }
 
-// getAddressFromPool adds an ownerReference on the referenced Metal3IPPool
-// objects. It then tries to fetch the Metal3IPAddress if it exists, and asks
+// getAddressFromPool creates an IPClaim if it does not exist.
+// It then tries to fetch the Metal3IPAddress if it exists, and asks
```
kashifest, line 663: A bit confused about removing Metal3 prefix from IPClaim in some comments while leaving for others, for example here we are leaving the prefix for Metal3IPAddress

lentzi90: Oh I see. I was mostly changing the comments that were confusing (mentioned IPPool instead of IPClaim) and didn't think about the prefix. Should I add it back? Or better to remove it everywhere?

kashifest: I would let the prefix be everywhere to be specific

kashifest:
I don't see any unit tests added to check this behavior.

baremetal/metal3data_manager.go
```diff
@@ -761,8 +762,8 @@ func (m *DataManager) getAddressFromPool(ctx context.Context, poolName string,
 	return addresses, false, nil
 }
 
-// releaseAddressFromPool removes the owner reference on existing referenced
-// Metal3IPPool objects.
+// releaseAddressFromPool deletes the IPClaim owned by this Metal3Data found in the IPPool.
+// Returns map of poolName => deleted, if it should requeue and error.
```
furkatgofurov7, line 766: I am not sure if this comment is clear, maybe we need to rephrase a bit?

baremetal/metal3data_manager.go
```diff
@@ -784,6 +785,26 @@ func (m *DataManager) releaseAddressFromPool(ctx context.Context, poolName strin
 		addresses[poolName] = true
 		return addresses, false, nil
 	}
+	// Remove finalizer from IPClaim since we no longer need it
+	ipClaim.Finalizers = Filter(ipClaim.Finalizers, infrav1.DataFinalizer)
```
furkatgofurov7, line 789: we have a `UnsetFinalizer()` method which removes only the finalizers from the m3Data, should we reuse that or have something similar ?

baremetal/metal3data_manager.go
```diff
@@ -784,6 +785,26 @@ func (m *DataManager) releaseAddressFromPool(ctx context.Context, poolName strin
...
+	// Remove ownerRef
+	newOwners := []metav1.OwnerReference{}
```
furkatgofurov7, line 791: IMO `newOwnerRefs` better fits in the context

baremetal/metal3data_manager.go
lentzi90: I was thinking it is not necessary to have a function for only this. Maybe for ownerRefs, that is a bit longer, but this is just one line anyway.

furkatgofurov7: you could rework the same method a bit and make it accept the input so that whenever it is called the corresponding objects could be passed to it. But that is just an improvement of course and can be left from this PR

furkatgofurov7:
> test-ubuntu-e2e-parallel-main

let's run at least one flavor normal e2e (instead of parallel) test as well, AFAIR it is a bit different than the one we triggered already in this PR

baremetal/metal3data_manager.go
lentzi90: Updated, PTAL!

lentzi90:
> I don't see any unit tests added to check this behavior.

Good point! I added now so that it at least check that the finalizer is added to the IPClaim when it is created.

For the changes to the owner references it is a bit more complicated. The function that removes the ownerRef from the IPClaim is the same that deletes it from Kubernetes (`releaseAddressFromPool`), which means that it is a bit tricky to test in a unit test AFAIU. The fake client won't respect finalizers, so the IPClaim will be deleted immediately and we cannot check if it removed the ownerRef first. So we would need to use mocks instead and I'm not sure if it is worth it.
Thoughts?

baremetal/metal3data_manager.go
lentzi90: Right, true. I was a bit scared to touch it to be honest since it is a public function that is called from the controller. If it is alright I would like to keep it this way for now.

baremetal/metal3data_manager_test.go
```diff
@@ -1319,14 +1320,21 @@ var _ = Describe("Metal3Data manager", func() {
...
+					Name:      tc.ipClaim.Name,
+					Namespace: tc.ipClaim.Namespace,
 				}
 
-				err = dataMgr.client.Get(context.TODO(), poolNamespacedName, capm3IPClaim)
-				Expect(err).To(HaveOccurred())
-				Expect(apierrors.IsNotFound(err)).To(BeTrue())
+				err = dataMgr.client.Get(context.TODO(), claimNamespacedName, capm3IPClaim)
+				if tc.injectDeleteErr {
+					// There was an error deleting the claim, so we expect it to still be there
```
kashifest, line 1330: is this comment aligned with the code? A bit confused here

kashifest: I mean the next line expects error to be nil

lentzi90: Ah hmm maybe a bit vague. I meant that we expect an error when deleting, meaning that the IPClaim is still expected to exist (since we could not delete it)

lentzi90: We expect error when deleting, which means the IPClaim should still exist. So when we try to `get` it above, we expect no problems, i.e. the error from get should be nil.

kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #688, opened 2022-08-18
🐛 Only use redfish for live-iso test

lentzi90:

The live-iso e2e test is failing because it is trying to use a BMH with ipmi, which does not support live-iso. This PR filters the BMHs so that we only select those that use redfish instead.


lentzi90:
Looks like this does not solve the issue. Closing.

PR #690, opened 2022-08-18
🐛Only use redfish-virtualmedia for live-iso test

adilGhaffarDev:

The live-iso e2e test is failing because it is trying to use a BMH with ipmi or redfish, which do not support live-iso. This PR filters the BMHs so that we only select those that use redfish-virtualmedia instead.


adilGhaffarDev:
e2e tests with the metl3 dev env PR(https://github.com/metal3-io/metal3-dev-env/pull/1055)
centos: https://jenkins.nordix.org/job/metal3_capm3_main_e2e_test_centos/53/
ubuntu: https://jenkins.nordix.org/job/metal3_capm3_main_e2e_test_ubuntu/40/

kashifest (approved):

smoshiur1237 (commented):
/lgtm

PR #683, opened 2022-08-11
Report config error

k82cn:
Signed-off-by: Klaus Ma <klausm@nvidia.com>

**What this PR does / why we need it**:

If Metal3Machine's spec is in-correct, it's better to report this err instead of ignore it.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #682

PR #693, opened 2022-08-23
🐛 fix Image validation requiring urls for checksum

### LINKED ISSUE #692: image checksum can no longer be set to a hash
Set `spec.image.checksum` of `Metal3Machine` to a md5 hash. The m3m is deemed invalid and reconciliation is aborted.

The machine gets reconciled and the hash is used for verification, as per its [description](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/api/v1beta1/common_types.go#L61).

Introduced by https://github.com/metal3-io/cluster-api-provider-metal3/pull/604/files#diff-90dc8b367863881133b9eb436059164daeadba015bed7d4124e6ea3c18828a2aR102

**Environment:**

- Cluster-api-provider-metal3 version: 1.1.2

/kind bug

schrej:
Due to a recent change, the Image.Checksum field only accepted URLs and not checksums directly. This PR fixes that, and applies the same validation for Metal3MachineTemplate and Metal3Machine.

It fixes a few minor linter complaints as well.

Fixes #692


schrej:
unrelated
/retest

lentzi90 (approved):
Nice!

lentzi90:
@schrej did you remember to run `make generate` before commit? It looks like there are some changes that affects the CRDs and this requires `make generate` since we store them in the repo.

schrej:
You're right. There was another test failing due to timeout, so I thought generate might've failed because of that as well. Should be good now.

baremetal/metal3machine_manager.go
```diff
@@ -299,14 +299,6 @@ func (m *MachineManager) Associate(ctx context.Context) error {
...
 	}
 
-	config := m.Metal3Machine.Spec
-	err := config.IsValid()
-	if err != nil {
-		// Should have been picked earlier. Do not requeue
-		m.SetError(err.Error(), capierrors.InvalidConfigurationMachineError)
-		return nil
-	}
-
```
furkatgofurov7, line 309: I remember this being discussed, but this is not directly related to what this PR trying to achieve, is not it?

schrej: It is not. It's essentially an alternative solution to #683. I can split it into a fix for the validate function and moving the validation to the webhooks if you prefer.

furkatgofurov7: hmm, no strong opinion on splitting & moving, but I wanted to make sure is, not to fix different issues with the same commit at least.

furkatgofurov7: also 684 will be redundant after this AFAIU

kashifest: you mean #683 @furkatgofurov7 ?

api/v1beta1/metal3machine_webhook.go
```diff
@@ -53,28 +53,8 @@ func (c *Metal3Machine) ValidateDelete() error {
...
-		allErrs = append(
-			allErrs,
-			field.Invalid(
-				field.NewPath("spec", "Image", "Checksum"),
-				c.Spec.Image.Checksum,
-				"is required",
-			),
-		)
-	}
+	allErrs = append(allErrs, c.Spec.Image.Validate(*field.NewPath("Spec", "Image"))...)
```
dtantsur, line 57: Side note: this assumes an Image is required. BMH can have a CustomDeploy instead.

dtantsur (approved):

api/v1beta1/metal3machine_webhook.go
kashifest: Thanks @dtantsur for pointing to CustomDeploy. This looks pretty interesting. However, for CAPM3 workflow Image field is still a required field. CAPM3 hasn't adopted yet the customdeploy workflow.

Also in BMO your implementation didnt land with any Documentation, and there is no mention of that here https://github.com/metal3-io/baremetal-operator/blob/main/docs/api.md#image.  I think it will be good to have some doc on this in BMO.

kashifest:
I am ok with this one. I will run e2e on this and remove the hold.

/test-ubuntu-e2e-main

schrej:
@kashifest e2e passed, can the hold be removed?

kashifest:
/hold cancel
Thanks a lot @schrej

schrej:
Can we backport this to 1.1 for the next patch release?
This bug prevents us from upgrading to 1.1.2, and might affect others as well.

kashifest:
I am ok with backport. @furkatgofurov7 @lentzi90 wdyt ?

lentzi90:
Makes sense to backport IMO

schrej:
I'll open a PR soonish.

furkatgofurov7:
> Can we backport this to 1.1 for the next patch release? This bug prevents us from upgrading to 1.1.2, and might affect others as well.

@schrej hi, FYI we are planning to have a patch release tomorrow

PR #696, opened 2022-08-25
🌱 optimize ip allocation

schrej:
It splits the allocation of IP addresses from IP pools into two steps: creating the claim, and attempting to fetch the address.
This way we can give the IPAM providers a little bit of time to fulfil the claim. To avoid unnecessary re-queues we wait for 200ms if a claim should be fetched again before attempting to fetch the address.

In an ideal case, one or more claims would be created because they do not exist. We then wait 200ms, fetch the claims again, hoping that they were updated, and then fetch the IP address the claim was fulfilled with.

The delay of 200ms was chosen arbitrarily. Not sure what a good value would be here.


kashifest:
> **What this PR does / why we need it**: It splits the allocation of IP addresses from IP pools into two steps: creating the claim, and attempting to fetch the address. This way we can give the IPAM providers a little bit of time to fulfil the claim. To avoid unnecessary re-queues we wait for 200ms if a claim should be fetched again before attempting to fetch the address.
> 
> In an ideal case, one or more claims would be created because they do not exist. We then wait 200ms, fetch the claims again, hoping that they were updated, and then fetch the IP address the claim was fulfilled with.
> 
> The delay of 200ms was chosen arbitrarily. Not sure what a good value would be here.

A bit unsure about the motivation here. Whats the harm if the reconciliation loop is requeing ? If you are concerned about too many reques, perhaps introducing something like [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/0d830904d16b3b9e4e8884ed91875ccf8afba041/baremetal/metal3machine_manager.go#L65) would make it much cleaner ?

schrej:
> A bit unsure about the motivation here. Whats the harm if the reconciliation loop is requeing ? If you are concerned about too many reques, perhaps introducing something like [this](https://github.com/metal3-io/cluster-api-provider-metal3/blob/0d830904d16b3b9e4e8884ed91875ccf8afba041/baremetal/metal3machine_manager.go#L65) would make it much cleaner ?

That's how it's working right now. If any of the IPAddresses is missing, it will requeue after 30s. The goal is to handle an entire machine with a single reconciliation after it was first created, at least if the IP gets allocated quickly enough. I've initially suggested it in [Slack](https://kubernetes.slack.com/archives/CHD49TLE7/p1659691199093949?thread_ts=1659528555.152719&cid=CHD49TLE7), and just went for an implementation as it's not that complicated.

The ideal solution would be registering watches on IPClaims so we can react as soon as something changes, but with the CAPI contract that would be quite difficult to do.

lentzi90 (commented):
This looks like a good cleanup to me. But I think the 200ms wait that is mentioned in the description is not gone?

schrej:
Yeah. We discussed this during the community call and it was decided that we don't wait after creation. Iirc the main reason was that we don't want to block the reconciliation queue unnecessarily. I'll update the description.

This is also missing a test for claim creation. I had one before, but the complexity increased a lot since preallocations were introduced in the meantime, which requires to mock m3dt, m3m, ma and bmh in addition to m3d.

baremetal/metal3data_manager.go
```diff
@@ -464,78 +490,87 @@ func (m *DataManager) m3IPClaimObjectMeta(name, poolRefName string, preallocatio
...
+
+	if ok := errors.As(err, &hasRequeueAfterError); !ok {
+		return reconciledClaim{m3Claim: ipClaim}, err
+	}
+
+	m3dt, err := fetchM3DataTemplate(ctx, &m.Data.Spec.Template, m.client,
+		m.Log, m.Data.Labels[clusterv1.ClusterLabelName],
+	)
 	if err != nil {
-		if ok := errors.As(err, &hasRequeueAfterError); !ok {
```
furkatgofurov7, line 473: the idea here was: we go fetch m3dt => m3m => bmh only, if IPClaim with normal "Data,Name+pool.Name" is not found, otherwise there is no need to have those objects to be fetched on each call. But now, we have to fetch them every time

schrej: In line 497 I check whether err is nil, and return the claim if it is found. So the behaviour is still the same, just with less nesting.

furkatgofurov7: My bad, misread that

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):
Thanks for the PR @schrej

PR #698, opened 2022-08-30
🌱 Run linter also on api folder

lentzi90:

Run the linter for the api folder. This was already the case for fmt and vet for example but was missing for the linter.

### DIFF
--- a/api/v1alpha5/metal3datatemplate_types.go
+++ b/api/v1alpha5/metal3datatemplate_types.go
@@ -183,7 +183,7 @@ type NetworkDataLinkEthernet struct {
 	Type string `json:"type"`
 
 	// Id is the ID of the interface (used for naming)
-	Id string `json:"id"`
+	ID string `json:"id"`
 
 	// +kubebuilder:default=1500
 	// +kubebuilder:validation:Maximum=9000
@@ -203,7 +203,7 @@ type NetworkDataLinkBond struct {
 	BondMode string `json:"bondMode"`
 
 	// Id is the ID of the interface (used for naming)
-	Id string `json:"id"`
+	ID string `json:"id"`
 
 	// +kubebuilder:default=1500
 	// +kubebuilder:validation:Maximum=9000
@@ -225,7 +225,7 @@ type NetworkDataLinkVlan struct {
 	VlanID int `json:"vlanID"`
 
 	// Id is the ID of the interface (used for naming)
-	Id string `json:"id"`
+	ID string `json:"id"`
 
 	// +kubebuilder:default=1500
 	// +kubebuilder:validation:Maximum=9000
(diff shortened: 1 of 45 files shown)


lentzi90:
To make this easier to review I made this as 2 commits. The first has all the changes to comments and the makefile. The second has some code changes that the linter required. I can squash them before merging

api/v1beta1/metal3datatemplate_types.go
```diff
@@ -196,15 +196,15 @@ type NetworkLinkEthernetMac struct {
...
-// NetworkDataLinkEthernet represents an ethernet link object
+// NetworkDataLinkEthernet represents an ethernet link object.
 type NetworkDataLinkEthernet struct {
 	// +kubebuilder:validation:Enum=bridge;dvs;hw_veb;hyperv;ovs;tap;vhostuser;vif;phy
 	// Type is the type of the ethernet link. It can be one of:
 	// bridge, dvs, hw_veb, hyperv, ovs, tap, vhostuser, vif, phy
 	Type string `json:"type"`
 
 	// Id is the ID of the interface (used for naming)
-	Id string `json:"id"`
```
lentzi90, line 207: Note this! In a few places the linter wanted to change `Id -> ID`. Is this ok? I'm not sure if this should be considered a breaking change or something. The field in the CRDs doesn't change but if someone imported the CAPM3 API somewhere I guess it would impact them.

kashifest (approved):

api/v1alpha5/metal3datatemplate_types.go
```diff
@@ -175,15 +175,15 @@ type NetworkLinkEthernetMac struct {
...
+// NetworkDataLinkEthernet represents an ethernet link object.
 type NetworkDataLinkEthernet struct {
 	// +kubebuilder:validation:Enum=bridge;dvs;hw_veb;hyperv;ovs;tap;vhostuser;vif;phy
 	// Type is the type of the ethernet link. It can be one of:
 	// bridge, dvs, hw_veb, hyperv, ovs, tap, vhostuser, vif, phy
 	Type string `json:"type"`
 
 	// Id is the ID of the interface (used for naming)
-	Id string `json:"id"`
+	ID string `json:"id"`
```
furkatgofurov7, line 186: Bit late on this, but This is an API [change](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) AFAIK for the old API and we should not have touch it..

kashifest: Hmmm true, probably we need to revert it

kashifest: Thats v1a5 btw

lentzi90: Ow a bit quick to merge this. I also commented on this a bit up. Didn't squash yet either. But that means at least we can revert only the second commit.

PR #703, opened 2022-09-02
🐛 Use keyfile format for CentOS network config

lentzi90:

The network config format was already changed in metal3-dev-env. We need to use the same here or it will be ignored and break the tests.

Also fixes the systemctl commands for enabling the monitor.keepalived unit. It looks like there has been some changes in systemd which make `systemctl enable monitor.keepalived.service` fail because it is operating on a symlink.

```
Failed to enable unit: Refusing to operate on alias name or linked unit file: monitor.keepalived.service
```

My understanding is, that instead of first linking and then enabling the link, we should just enable the unit using the absolute path. This also creates a symlink.
This issue is not *yet* seen on Ubuntu because it has an older systemd version.

See:
- https://github.com/metal3-io/metal3-dev-env/pull/1009
- https://github.com/metal3-io/metal3-dev-env/pull/1062

### DIFF
--- a/test/e2e/data/infrastructure-metal3/bases/centos-kubeadm-config/centos-kubeadm-config.yaml
+++ b/test/e2e/data/infrastructure-metal3/bases/centos-kubeadm-config/centos-kubeadm-config.yaml
@@ -163,9 +163,7 @@ spec:
     - nmcli connection load /etc/NetworkManager/system-connections/ironicendpoint.nmconnection
     - nmcli connection up ironicendpoint
     - systemctl enable --now crio keepalived kubelet
-    - systemctl link /lib/systemd/system/monitor.keepalived.service
-    - systemctl enable monitor.keepalived.service
-    - systemctl start monitor.keepalived.service
+    - systemctl enable --now /lib/systemd/system/monitor.keepalived.service
 ---
 apiVersion: bootstrap.cluster.x-k8s.io/${CAPI_VERSION}
 kind: KubeadmConfigTemplate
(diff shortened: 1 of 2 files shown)


lentzi90:
/hold
We don't have the jinja templating in place for this to work. I need to do it in a different way.

scripts/ci-e2e.sh
```diff
@@ -58,6 +58,9 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
 # shellcheck disable=SC1091,SC1090
 source "${M3_DEV_ENV_PATH}/lib/ironic_tls_setup.sh"
 
+export GINKGO_FOCUS="[remediation]"
```
mboukhalfa, line 61:
```suggestion
export GINKGO_FOCUS="remediation"
```

mboukhalfa:
@lentzi90 Are you planning to parametrize the hardcoded config on this PR or merge that to fix CI that follow up with paramterization PR ? 
+1 for unlocking the CI now

furkatgofurov7 (approved):
Thank you. Let's make sure to backport this to the `release-1.1` branch once merged to get the CI back there too. 

/approve

furkatgofurov7:
> Are you planning to parametrize the hardcoded config on this PR or merge that to fix CI than follow up with paramterization PR ?

@mboukhalfa which config are you referring to?

lentzi90:
Let's merge it now and discuss the parameterization separately
/hold cancel

PR #704, opened 2022-09-02
🌱 Don't remove ownerRef from IPClaim when deleting

lentzi90:

The initial implementation for checking if an IPClaim was in use, checked the ownerRef to determine if CAPM3 was using it or not. Since this was an API change in IPAM and a bit now really what ownerRefs are for, we are reverting it and rely on finalizers instead.

Relevant PR in IPAM: https://github.com/metal3-io/ip-address-manager/pull/133


furkatgofurov7 (approved):

kashifest (commented):
/lgtm

PR #709, opened 2022-09-07
🌱 Improve handling of IPClaims with deletion timestamp

lentzi90:

- If the IPClaim is old (=not matching current Metal3Data), attempt to clean it up
- If the IPClaim is in use, accept it even if it has a deletion timestamp. This could happen e.g. if the user tried to delete it but the Metal3Data/Metal3Machine still exists.

### DIFF
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -537,9 +537,26 @@ func (m *DataManager) getAddressFromM3Pool(ctx context.Context, poolRef corev1.T
 	}
 
 	if !ipClaim.DeletionTimestamp.IsZero() {
-		// This Metal3IPClaim is about to be deleted so we cannot use it. Requeue.
-		m.Log.Info("Found IPClaim with deletion timestamp, requeuing.", "IPClaim", ipClaim)
-		return addressFromPool{}, true, nil
+		// Is it "our" ipClaim, or does it belong to an old and deleted Metal3Data with the same name?
+		matchingOwnerRef := false
+		for _, ownerRef := range ipClaim.OwnerReferences {
+			if ownerRef.UID == m.Data.GetUID() {
+				matchingOwnerRef = true
+			}
+		}
+		if !matchingOwnerRef {
+			// It is not our IPClaim so we should not use it. Attempt to remove finalizer if it is still there.
+			m.Log.Info("Found old IPClaim with deletion timestamp. Attempting to clean up and requeue.", "IPClaim", ipClaim)
+			if Contains(ipClaim.Finalizers, infrav1.DataFinalizer) {
+				ipClaim.Finalizers = Filter(ipClaim.Finalizers, infrav1.DataFinalizer)
+				err = updateObject(ctx, m.client, ipClaim)
+				if err != nil {
+					m.Log.Info("Failed to remove finalizer from old IPClaim", "IPClaim", ipClaim, "error", err)
+				}
+			}
+			return addressFromPool{}, true, nil
+		}
+		m.Log.Info("IPClaim has deletion timestamp but is still in use!", "IPClaim", ipClaim)
 	}
 
 	if ipClaim.Status.ErrorMessage != nil {
@@ -563,8 +580,10 @@ func (m *DataManager) getAddressFromM3Pool(ctx context.Context, poolRef corev1.T
 
 	if err := m.client.Get(ctx, addressNamespacedName, ipAddress); err != nil {
 		if apierrors.IsNotFound(err) {
+			m.Log.Info("IPAddress not found, requeuing", "IPAddress", ipClaim.Status.Address.Name)
 			return addressFromPool{}, true, nil
 		}
+		m.Log.Error(err, "Unable to get IPAddress.", "IPAddress", ipClaim.Status.Address.Name)
 		return addressFromPool{}, false, err
 	}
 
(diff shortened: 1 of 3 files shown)


baremetal/metal3data_manager.go
```diff
@@ -505,8 +522,10 @@ func (m *DataManager) getAddressFromM3Pool(ctx context.Context, poolRef corev1.T
 
 	if err := m.client.Get(ctx, addressNamespacedName, ipAddress); err != nil {
 		if apierrors.IsNotFound(err) {
+			m.Log.Info("IPAddress not found, requeuing", "IPAddress", ipClaim.Status.Address.Name)
 			return addressFromPool{}, true, nil
 		}
+		m.Log.Error(err, "Unable to get IPAddress.", "IPAddress", ipClaim.Status.Address.Name)
```
lentzi90, line 528: Not strictly related, but I added these while debugging the tests and found a test that didn't take the expected path as a consequence.

baremetal/metal3data_manager_test.go
```diff
@@ -1184,11 +1184,12 @@ var _ = Describe("Metal3Data manager", func() {
 			},
 			expectRequeue: true,
 		}),
-		Entry("IPClaim with deletion timestamp", testCaseGetAddressFromPool{
+		Entry("Old IPClaim with deletion timestamp", testCaseGetAddressFromPool{
 			m3d: &infrav1.Metal3Data{
 				ObjectMeta: testObjectMeta(metal3DataName, namespaceName, ""),
 			},
 			poolName:        testPoolName,
+			poolRef:         corev1.TypedLocalObjectReference{Name: testPoolName},
```
lentzi90, line 1192: This test was not working properly when the `poolRef` was missing. It just returned early because it could not find the IPClaim and asked for requeue. Now it is working as expected.

baremetal/metal3data_manager_test.go
```diff
@@ -1241,7 +1290,7 @@ var _ = Describe("Metal3Data manager", func() {
 			poolRef:         corev1.TypedLocalObjectReference{Name: testPoolName},
 			expectedAddress: addressFromPool{},
 			ipClaim: &ipamv1.IPClaim{
-				ObjectMeta: testObjectMeta("abc-abc", namespaceName, ""),
+				ObjectMeta: testObjectMeta(metal3DataName+"-"+testPoolName, namespaceName, ""),
```
lentzi90, line 1293: The name was wrong here so the test went to "IPClaim not found" instead of finding the IPClaim and then not fining the IPAddress.

baremetal/metal3data_manager.go
```diff
@@ -479,9 +479,26 @@ func (m *DataManager) getAddressFromM3Pool(ctx context.Context, poolRef corev1.T
 	}
 
 	if !ipClaim.DeletionTimestamp.IsZero() {
-		// This Metal3IPClaim is about to be deleted so we cannot use it. Requeue.
-		m.Log.Info("Found IPClaim with deletion timestamp, requeuing.", "IPClaim", ipClaim)
-		return addressFromPool{}, true, nil
+		// Is it "our" ipClaim, or does it belong to an old and deleted Metal3Data with the same name?
+		our := false
```
furkatgofurov7, line 483: not a blocker but maybe we could rename it to something like: 
"owned" or "withOwnerRef"?

furkatgofurov7 (commented):
Looks good, nit:

baremetal/metal3data_manager.go
kashifest: yes this doesn't seem the right convention for the naming.

kashifest (approved):

baremetal/metal3data_manager.go
lentzi90: Changed to `matchingOwnerRef`

furkatgofurov7 (commented):
/lgtm

baremetal/metal3data_manager_test.go
```diff
@@ -1258,16 +1307,16 @@ var _ = Describe("Metal3Data manager", func() {
...
 					Address: &corev1.ObjectReference{
 						Name:      "abc-192.168.0.11",
 						Namespace: namespaceName,
 					},
 				},
 			},
-			expectRequeue: false,
-			expectError:   true,
+			expectRequeue: true,
+			expectError:   false,
```
lentzi90, line 1319: Please double check! This was changed in https://github.com/metal3-io/cluster-api-provider-metal3/commit/42713219f82f33ca625eb7489a210709a000c162# and I think it was a mistake.
The test is supposed to check "IP address not found". My understanding is that this should result in requeue, not error, as seen [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3data_manager.go#L566).
The reason we got an error is because the name was wrong above so it didn't find the *M3DT*, which would return an error.
/cc @furkatgofurov7

furkatgofurov7: Yeah, should be fine now

lentzi90:
/test-centos-e2e-parallel-main
Different flakes...

lentzi90:
It has now failed in 3 different places :frowning_face: 
1. When scaling *down* KCP in e2e pivoting based tests
2. The flake with empty providerID in e2e remediation
3. When scaling *up* the KCP in e2e pivoting based tests

/test-centos-e2e-parallel-main

PR #722, opened 2022-09-16
🐛 Fix BMOBRANCH fetching in e2e by sourcing lib/releases from dev env

kashifest:
After this PR https://github.com/metal3-io/metal3-dev-env/pull/1072 BMOBRANCH var is moved from lib/common.sh to lib/releases.sh. So we need to source both the files here.


furkatgofurov7 (approved):
/retitle 🐛 Fix BMOBRANCH fetching in e2e by sourcing lib/releases from dev env

/approve

PR #728, opened 2022-09-16
🌱  Refactor e2e: Remove test suffix from function files and use function inputs instead of global vars

mboukhalfa:
Refactor e2e:
use function inputs instead of global vars
remove test suffix from function files

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 
@@ -27,6 +28,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type vmState string
+
+const (
+	running        vmState = "running"
+	paused         vmState = "paused"
+	shutoff        vmState = "shutoff"
+	other          vmState = "other"
+	artifactoryURL         = "https://artifactory.nordix.org/artifactory/metal3/images/k8s"
+	imagesURL              = "http://172.22.0.1/images"
+	ironicImageDir         = "/opt/metal3-dev-env/ironic/html/images"
+)
+
 func Byf(format string, a ...interface{}) {
 	By(fmt.Sprintf(format, a...))
 }
@@ -39,6 +52,13 @@ func LogFromFile(logFile string) {
 	data, err := os.ReadFile(filepath.Clean(logFile))
 	Expect(err).To(BeNil(), "No log file found")
 	Logf(string(data))
+}
+
+// return only the boolean value from ParseBool.
+func getBool(s string) bool {
+	b, err := strconv.ParseBool(s)
+	Expect(err).To(BeNil())
+	return b
 }
 
 // logTable print a formatted table into the e2e logs.
@@ -83,12 +103,6 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
 		}, intervalsGetter(specName, "wait-delete-cluster")...)
 	}
 }
-
-const (
-	artifactoryURL = "https://artifactory.nordix.org/artifactory/metal3/images/k8s"
-	imagesURL      = "http://172.22.0.1/images"
-	ironicImageDir = "/opt/metal3-dev-env/ironic/html/images"
-)
 
 func ensureImage(k8sVersion string) (imageURL string, imageChecksum string) {
 	osType := strings.ToLower(os.Getenv("OS"))
--- a/test/e2e/metal3remediation.go
+++ b/test/e2e/metal3remediation.go
@@ -13,20 +13,31 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func metal3remediation() {
+type Metal3RemediationInput struct {
+	E2EConfig             *clusterctl.E2EConfig
+	BootstrapClusterProxy framework.ClusterProxy
+	TargetCluster         framework.ClusterProxy
+	SpecName              string
+	ClusterName           string
+	Namespace             string
+}
+
+func metal3remediation(ctx context.Context, inputGetter func() Metal3RemediationInput) {
 	Logf("Starting metal3 remediation tests")
+	input := inputGetter()
+	bootstrapClient := input.BootstrapClusterProxy.GetClient()
+	targetClient := input.TargetCluster.GetClient()
 
-	bootstrapClient := bootstrapClusterProxy.GetClient()
-	targetClient := targetCluster.GetClient()
-
-	_, workerM3Machines := getMetal3Machines(ctx, bootstrapClient, clusterName, namespace)
+	_, workerM3Machines := getMetal3Machines(ctx, bootstrapClient, input.ClusterName, input.Namespace)
 	Expect(len(workerM3Machines)).To(BeNumerically(">", 0))
 
 	getBmhFromM3Machine := func(m3Machine infrav1.Metal3Machine) (result bmov1alpha1.BareMetalHost) {
-		Expect(bootstrapClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: metal3MachineToBmhName(m3Machine)}, &result)).To(Succeed())
+		Expect(bootstrapClient.Get(ctx, client.ObjectKey{Namespace: input.Namespace, Name: metal3MachineToBmhName(m3Machine)}, &result)).To(Succeed())
 		return result
 	}
 
@@ -35,7 +46,7 @@ func metal3remediation() {
 
 	workerMachineName, err := metal3MachineToMachineName(workerM3Machine)
 	Expect(err).ToNot(HaveOccurred())
-	workerMachine := getMachine(ctx, bootstrapClient, client.ObjectKey{Namespace: namespace, Name: workerMachineName})
+	workerMachine := getMachine(ctx, bootstrapClient, client.ObjectKey{Namespace: input.Namespace, Name: workerMachineName})
 	workerNodeName := workerMachineName
 	vmName := bmhToVMName(workerBmh)
 
@@ -44,7 +55,7 @@ func metal3remediation() {
 	m3Remediation := &infrav1.Metal3Remediation{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      workerNodeName,
-			Namespace: namespace,
+			Namespace: input.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					APIVersion:         clusterv1.GroupVersion.String(),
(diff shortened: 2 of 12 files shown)


mboukhalfa:
test-ubuntu-integration-main

test/e2e/node_reuse.go
```diff
@@ -21,18 +22,34 @@ import (
...
+	SpecName          string
+	ClusterName       string
+	Namespace         string
+}
+
+func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
 	Logf("Starting node reuse tests")
+	input := inputGetter()
+	managementClusterClient := input.TargetCluster.GetClient()
+	targetClusterClient := input.ManagementCluster.GetClient()
```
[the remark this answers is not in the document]
mboukhalfa: nice catch I think that would break ephemeral test I will fix that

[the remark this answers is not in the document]
mboukhalfa: done

test/e2e/pivoting.go
```diff
@@ -0,0 +1,447 @@
...
+	removeIronic(ctx, func() RemoveIronicInput {
+		return RemoveIronicInput{
+			ManagementCluster: input.BootstrapClusterProxy,
+			IsDeployment:      isDeployment,
+			Namespace:         input.E2EConfig.GetVariable(ironicNamespace),
+			NamePrefix:        input.E2EConfig.GetVariable(NAMEPREFIX),
+		}
+	})
+
+	By("Create Ironic input.Namespace")
```
lentzi90, line 77: A bit too much find/replace maybe :joy: 
```suggestion
	By("Create Ironic namespace")
```

test/e2e/pivoting.go
```diff
@@ -0,0 +1,447 @@
...
+	targetClusterClientSet := input.TargetCluster.GetClientSet()
+	ironicNamespace := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: input.E2EConfig.GetVariable(ironicNamespace),
+		},
+	}
+	_, err := targetClusterClientSet.CoreV1().Namespaces().Create(ctx, ironicNamespace, metav1.CreateOptions{})
+	Expect(err).To(BeNil(), "Unable to create the Ironic input.Namespace")
```
lentzi90, line 85:
```suggestion
	Expect(err).To(BeNil(), "Unable to create the Ironic namespace")
```

lentzi90 (commented):
Nice refactor! This makes the code much easier to reuse and understand I think.

These functions take way too many arguments I think, but let's not make this PR any bigger that it already is. We can fix it later.

- `updateNodeReuse`
- `pointMDtoM3mt`
- `updateBootImage`
- `untaintNodes`

test/e2e/pivoting.go
mboukhalfa: thanks for catching that

test/e2e/pivoting.go
mboukhalfa: fixed

test/e2e/pivoting_test.go
```diff
@@ -24,146 +25,212 @@ import (
...
-func installIronicBMO(targetCluster framework.ClusterProxy, isIronic, isBMO string) {
-	ironicTLSSetup := "IRONIC_TLS_SETUP"
-	ironicBasicAuth := "IRONIC_BASIC_AUTH"
+type installIronicBMOInput struct {
+	ManagementCluster          framework.ClusterProxy
+	BMPPath                    string
+	IsIronic                   string
+	IsBMO                      string
+	IronicTLSSetup             string
+	IronicBasicAuth            string
```
lentzi90, line 208: It would make more sense to use bools instead of strings here IMO, and some of the names seem strange. Using bools also means you don't need the `isTrue` string to make the linter happy.
```suggestion
	BMOPath                    string
	InstallIronic              bool
	InstallBMO                 bool
	IronicTLSSetup             bool
	IronicBasicAuth            bool
```

test/e2e/pivoting_test.go
```diff
@@ -187,38 +254,39 @@ func installIronicBMO(targetCluster framework.ClusterProxy, isIronic, isBMO stri
...
-	for _, container := range ironicContainerList {
-		d := 1 * time.Minute
-		err = dockerClient.ContainerStop(ctx, container, &d)
-		Expect(err).To(BeNil(), "Unable to stop the container %s: %v", container, err)
-		err = dockerClient.ContainerRemove(ctx, container, removeOptions)
-		Expect(err).To(BeNil(), "Unable to delete the container %s: %v", container, err)
-	}
+type RemoveIronicInput struct {
+	ManagementCluster framework.ClusterProxy
+	IsDeployment      string
```
lentzi90, line 259:
```suggestion
	IsDeployment      bool
```

test/e2e/pivoting_test.go
```diff
@@ -24,146 +25,212 @@ import (
...
 
-const bmoPath = "BMOPATH"
+const (
+	bmoPath                    = "BMOPATH"
+	ironicTLSSetup             = "IRONIC_TLS_SETUP"
+	ironicBasicAuth            = "IRONIC_BASIC_AUTH"
+	KIND                       = "kind"
+	NAMEPREFIX                 = "NAMEPREFIX"
+	restartContainerCertUpdate = "RESTART_CONTAINER_CERTIFICATE_UPDATED"
+	ironicNamespace            = "IRONIC_NAMESPACE"
```
lentzi90, line 35: The names are a bit confusing (e.g. `bmoPath` is not the actual path to BMO but the variable name). Why are kind and nameprefix all caps?
```suggestion
	bmoPathVar                    = "BMOPATH"
	ironicTLSSetupVar             = "IRONIC_TLS_SETUP"
	ironicBasicAuthVar            = "IRONIC_BASIC_AUTH"
	kind                       = "kind"
	nameprefixVar                 = "NAMEPREFIX"
	restartContainerCertUpdateVar = "RESTART_CONTAINER_CERTIFICATE_UPDATED"
	ironicNamespaceVar            = "IRONIC_NAMESPACE"
```

lentzi90 (commented):
Added a few more suggestions

test/e2e/pivoting_test.go
mboukhalfa: these [`IsBMO`, `IsIronic` ..] are set later as argument to `deploy.sh` not sure if used as `bool` will still keep the same value ? and since they were already typed as strings this PR is just moving vars to input struct not trying to introduce any other changes. for the new vars I like  `IsDeployment` I kept the same type style .. Can we do that in a follow up after merging these it would make more sense to me even BMO has depracated positional args for deploy.sh https://github.com/metal3-io/baremetal-operator/blob/cc171d0d8ee21107255db5a5239bb29f2afb0477/tools/deploy.sh#L58

test/e2e/pivoting_test.go
mboukhalfa: This was the style used by CAPI https://github.com/kubernetes-sigs/cluster-api/blob/6c6543448bee351430e48f39883265f409c44662/test/e2e/common.go#L37-L47

test/e2e/pivoting_test.go
lentzi90: I'm not suggesting you should pass them to the script directly. But if you make a struct with input args, it makes more sense to use proper go types for it. Then when calling the script you can just [convert to strings](https://pkg.go.dev/strconv#FormatBool) inside the function instead of passing strings around in the go code.
To me it makes little sense adding the struct with all string values and then change it in a follow up. In that case just leave the struct out of this refactor and add it later. What do you think?

test/e2e/pivoting_test.go
lentzi90: Ok I see, let's keep the names as is then. But what about KIND and NAMEPREFIX? Why not change to `kind` and `nameprefix`?

test/e2e/pivoting_test.go
mboukhalfa: I think since we are reading the sometime the value from environment vars : `IronicTLSSetup: input.E2EConfig.GetVariable(ironicTLSSetup)` we need to do more changes for now it is easier to keep them string we can add those changes as a follow up

test/e2e/pivoting_test.go
mboukhalfa: fixed

test/e2e/pivoting_test.go
mboukhalfa: fixed

test/e2e/common.go
```diff
@@ -41,6 +51,12 @@ func LogFromFile(logFile string) {
 	Logf(string(data))
 }
 
+// return only the boolean value from ParseBool.
+func getBool(s string) bool {
+	b, _ := strconv.ParseBool(s)
+	return b
```
lentzi90, line 57: Maybe add an error check just to be sure? Otherwise we could miss empty strings or other unexpected situations.
```suggestion
	b, err := strconv.ParseBool(s)
	Expect(err).To(BeNil())
	return b
```

lentzi90 (commented):
Ok last nit, I promise. After this I'm happy to LGTM

test/e2e/common.go
mboukhalfa: Thanks indeed that was missing

lentzi90 (commented):
/lgtm

test/e2e/pivoting_test.go
```diff
@@ -24,146 +26,210 @@ import (
...
+	listMachines(ctx, input.BootstrapClusterProxy.GetClient(), client.InNamespace(input.Namespace))
+	listNodes(ctx, input.TargetCluster.GetClient())
 
 	By("Remove Ironic containers from the source cluster")
 	ephemeralCluster := os.Getenv("EPHEMERAL_CLUSTER")
-	if ephemeralCluster == KIND {
-		removeIronicContainers()
-	} else {
-		removeIronicDeployment()
+	isDeployment := true
```
furkatgofurov7, line 64: can we rename this to be more specific, i.e: `isIronicDeployment` or similar?

test/e2e/pivoting_test.go
```diff
@@ -24,146 +26,210 @@ import (
...
 
 	By("PIVOTING TESTS PASSED!")
 }
 
-func installIronicBMO(targetCluster framework.ClusterProxy, isIronic, isBMO string) {
-	ironicTLSSetup := "IRONIC_TLS_SETUP"
-	ironicBasicAuth := "IRONIC_BASIC_AUTH"
+type installIronicBMOInput struct {
+	ManagementCluster          framework.ClusterProxy
+	BMPPath                    string
```
furkatgofurov7, line 203: nit: BMOPath

test/e2e/common.go
```diff
@@ -27,6 +28,15 @@ import (
...
 )
 
+type vmState string
+
+const (
+	running vmState = "running"
+	paused  vmState = "paused"
+	shutoff vmState = "shutoff"
+	other   vmState = "other"
+)
```
furkatgofurov7, line 38: I would keep them in their own e2e test file if possible unless there is something I am missing here or at least merge others consts (lines 104-107) in this file into one

test/e2e/pivoting_test.go
```diff
@@ -24,146 +26,210 @@ import (
...
+	IsIronic                   bool
+	IsBMO                      bool
+	IronicTLSSetup             bool
+	IronicBasicAuth            bool
```
furkatgofurov7, line 207: since we are here, I would suggest renaming these to something more descriptive, i.e for someone new to the project:
```suggestion
	deployIronic             bool
	deployBMO                bool
	deployIronicTLSSetup     bool
	DeployIronicBasicAuth    bool
```

test/e2e/pivoting_test.go
```diff
@@ -258,32 +325,62 @@ func labelHDCRDs(targetCluster framework.ClusterProxy) {
...
+	removeIronic(ctx, func() RemoveIronicInput {
+		return RemoveIronicInput{
+			ManagementCluster: input.TargetCluster,
+			IsDeployment:      true,
+			Namespace:         input.E2EConfig.GetVariable(ironicNamespace),
+			NamePrefix:        input.E2EConfig.GetVariable(NamePrefix),
+		}
+	})
 	By("Reinstate Ironic containers and BMH")
+	// TODO: add this case to installIronicBMO function
```
furkatgofurov7, line 361: I am not sure if this is needed at all, if so, can you move it to the spefic func and make it clear what do we mean by "this case"

furkatgofurov7 (commented):
Some comments are inline for further improvement, but overall changes look good to me

test/e2e/pivoting_test.go
mboukhalfa: here in this case we deploy ironic locally using `run_local_ironic.sh` in the function `installIronicBMO` we deploy ironic or bmo or both as deployment using `deploy.sh` I think the same function can cover all the case I do not want to have multiple function deploying ironic that becomes confusing I will do this changes in a separate PR since it is not tight to the targeted changes here (using function inputs and removing file suffix)

test/e2e/metal3remediation.go
```diff
@@ -13,20 +13,31 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func metal3remediation() {
-	Logf("Starting metal3 remediation tests")
+type Metal3remediationInput struct {
```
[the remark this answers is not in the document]
mboukhalfa: fixed thanks

test/e2e/pivoting_test.go
mboukhalfa: changed to `isIronicDeployment`

test/e2e/pivoting_test.go
mboukhalfa: fixed

test/e2e/common.go
mboukhalfa: merged

test/e2e/pivoting_test.go
mboukhalfa: thanks

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):
Great cleanup @mboukhalfa !

/approve

PR #733, opened 2022-09-21
🐛 Fix dev-env templates path

mboukhalfa:
Fix dev-env template path since it was changed on dev-env here https://github.com/metal3-io/metal3-dev-env/pull/1061


kashifest (approved):

furkatgofurov7 (approved):
/retitle 🐛 Fix dev-env templates path

PR #734, opened 2022-09-21
🌱  Bump dependencies in hack/tools to match CAPI 1.2 release branch

kashifest:


furkatgofurov7 (approved):
The title should state maybe: *downgrade*? 

/cc @mboukhalfa @lentzi90

lentzi90 (commented):
/lgtm
But I agree with @furkatgofurov7 that it would be good to mention in the title that it is a downgrade. Normally "bump" would be for a newer version.

PR #735, opened 2022-09-21
🌱  Bump dependencies in hack/tools to match CAPI 1.1 release branch

kashifest:


kashifest:
This is breaking the CI. As the change is not significant, we will keep the current version and not go forward with this PR.
/close

PR #751, opened 2022-10-05
🌱 Add kubernetes upgrade test to e2e upgrade

Sunnatillo:
This PR:
  - adds kubernetes upgrade test to e2e upgrade tests
  - changing workflow from patching new image variables to m3machitemplate to creating new m3mtemplate with new variables in node reuse and k8s upgrade tests    
  - hardcodes INIT_WITH_KUBERNETES_VERSION variable to  v1.23.8 (latest v1a4 supports)

### DIFF
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -111,7 +111,8 @@ variables:
   KUBERNETES_VERSION: "v1.25.2"
   # INIT_WITH_KUBERNETES_VERSION will be used here
   # https://github.com/kubernetes-sigs/cluster-api/blob/bb377163f141d69b7a61479756ee96891f6670bd/test/e2e/clusterctl_upgrade.go#L170
-  INIT_WITH_KUBERNETES_VERSION: ${KUBERNETES_VERSION}
+  # INIT_WITH_KUBERNETES_VERSION  used by management cluster upgrade and it is fixed to 1.23.8 because v1a4 capi does not support k8s version higher
+  INIT_WITH_KUBERNETES_VERSION: "v1.23.8"
   CONTROL_PLANE_MACHINE_COUNT: 3
   WORKER_MACHINE_COUNT: 1
   APIVersion: "infrastructure.cluster.x-k8s.io/${CAPM3_VERSION}"
--- a/test/e2e/node_reuse.go
+++ b/test/e2e/node_reuse.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
-	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/cluster-api/util/patch"
@@ -74,11 +73,11 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
 	By("Scale down MachineDeployment to 0")
 	ScaleMachineDeployment(ctx, managementClusterClient, input.ClusterName, input.Namespace, 0)
 
-	Byf("Wait until the worker is scaled down and %d BMH(s) Available", numberOfWorkers)
+	Byf("Wait until the worker is scaled down and %d BMH(s) Available", 1)
 	WaitForNumBmhInState(ctx, bmov1alpha1.StateAvailable, WaitForNumInput{
 		Client:    managementClusterClient,
 		Options:   []client.ListOption{client.InNamespace(input.Namespace)},
-		Replicas:  numberOfWorkers,
+		Replicas:  1,
 		Intervals: input.E2EConfig.GetIntervals(input.SpecName, "wait-bmh-available"),
 	})
 
@@ -87,10 +86,11 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
 
 	By("Download image")
 	imageURL, imageChecksum := EnsureImage(upgradedK8sVersion)
-	By("Update KCP Metal3MachineTemplate with upgraded image to boot and set nodeReuse field to 'True'")
+	By("Set nodeReuse field to 'True' and create new KCP Metal3MachineTemplate with upgraded image to boot")
 	m3machineTemplateName := fmt.Sprintf("%s-controlplane", input.ClusterName)
 	updateNodeReuse(ctx, input.Namespace, true, m3machineTemplateName, managementClusterClient)
-	updateBootImage(ctx, input.Namespace, m3machineTemplateName, managementClusterClient, imageURL, imageChecksum, "raw", "md5")
+	newM3machineTemplateName := fmt.Sprintf("%s-new-controlplane", input.ClusterName)
+	createNewM3machineTemplate(ctx, input.Namespace, newM3machineTemplateName, m3machineTemplateName, managementClusterClient, imageURL, imageChecksum, "raw", "md5")
 
 	Byf("Update KCP to upgrade k8s version and binaries from %s to %s", kubernetesVersion, upgradedK8sVersion)
 	kcpObj := framework.GetKubeadmControlPlaneByCluster(ctx, framework.GetKubeadmControlPlaneByClusterInput{
@@ -98,18 +98,12 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
 		ClusterName: input.ClusterName,
 		Namespace:   input.Namespace,
 	})
-	patch := []byte(fmt.Sprintf(`{
-		"spec": {
-			"rolloutStrategy": {
-				"rollingUpdate": {
-					"maxSurge": 0
-				}
-			},
-			"version": "%s"
-		}
-	}`, upgradedK8sVersion))
-	err := managementClusterClient.Patch(ctx, kcpObj, client.RawPatch(types.MergePatchType, patch))
-	Expect(err).To(BeNil(), "Failed to patch KubeadmControlPlane")
+	helper, err := patch.NewHelper(kcpObj, managementClusterClient)
+	Expect(err).NotTo(HaveOccurred())
+	kcpObj.Spec.MachineTemplate.InfrastructureRef.Name = newM3machineTemplateName
+	kcpObj.Spec.Version = upgradedK8sVersion
+	kcpObj.Spec.RolloutStrategy.RollingUpdate.MaxSurge.IntVal = 0
+	Expect(helper.Patch(ctx, kcpObj)).To(Succeed())
 
 	By("Check if only a single machine is in Deleting state and no other new machines are in Provisioning state")
 	WaitForNumMachinesInState(ctx, clusterv1.MachinePhaseDeleting, WaitForNumInput{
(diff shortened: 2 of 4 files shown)


test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,263 @@
...
+				"rolloutStrategy": {
+					"rollingUpdate": {
+						"maxSurge": 0
+					}
+				},
+				"version": "%s"
+			}
+		}`, upgradedK8sVersion))
+		err := clusterClient.Patch(ctx, kcpObj, client.RawPatch(types.MergePatchType, patch))
+		Expect(err).To(BeNil(), "Failed to patch KubeadmControlPlane")
```
lentzi90, line 125: I'm a bit afraid of this actually. It is the same as in node reuse but I was not aware of it. In the ansible tests we create the new templates and then in [one single patch](https://github.com/metal3-io/metal3-dev-env/blob/c2ca4f453f31bd2cd7272ba62ba9ea731e86f484/tests/roles/run_tests/tasks/upgrade.yml#L378-L390) we update the infrastructureRef and version. But here we do two patches! And it even looks like we are patching the template itself. Maybe I'm missing something? I thought it was impossible to patch the template and this is why we create new templates.

My issue with 2 separate patches is that we can get inconsistencies. For a little while we will have wrong k8s version for the specified boot image.

test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,263 @@
...
+		imageName := fmt.Sprintf("%s_%s.qcow2", imageNamePrefix, upgradedK8sVersion)
+		Logf("IMAGE_NAME: %v", imageName)
+		rawImageName := fmt.Sprintf("%s_%s-raw.img", imageNamePrefix, upgradedK8sVersion)
+		Logf("RAW_IMAGE_NAME: %v", rawImageName)
+		imageLocation := fmt.Sprintf("%s_%s/", artifactoryURL, upgradedK8sVersion)
+		Logf("IMAGE_LOCATION: %v", imageLocation)
+		imageURL := fmt.Sprintf("%s/%s", imagesURL, rawImageName)
+		Logf("IMAGE_URL: %v", imageURL)
+		imageChecksum := fmt.Sprintf("%s/%s.md5sum", imagesURL, rawImageName)
+		Logf("IMAGE_CHECKSUM: %v", imageChecksum)
```
mboukhalfa, line 81: can we use this function https://github.com/metal3-io/cluster-api-provider-metal3/blob/496bfb481f8754772e8fe77fbda4464f0c339255/test/e2e/common.go#L93

Sunnatillo: done

test/e2e/upgrade_kubernetes_test.go
Sunnatillo: solved. Now it creates new m3machinetemplate

kashifest (approved):

test/e2e/node_reuse.go
```diff
@@ -87,29 +86,24 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
-		}
-	}`, upgradedK8sVersion))
-	err := managementClusterClient.Patch(ctx, kcpObj, client.RawPatch(types.MergePatchType, patch))
-	Expect(err).To(BeNil(), "Failed to patch KubeadmControlPlane")
+	helper, err := patch.NewHelper(kcpObj, managementClusterClient)
+	Expect(err).NotTo(HaveOccurred())
+	kcpObj.Spec.MachineTemplate.InfrastructureRef.Name = newM3machineTemplateName
+	kcpObj.Spec.Version = upgradedK8sVersion
+	kcpObj.Spec.RolloutStrategy.RollingUpdate.MaxSurge.IntVal = 0
+	Expect(helper.Patch(ctx, kcpObj)).To(Succeed())
```
furkatgofurov7, line 106: This is not correct based on what I previously dealt with usage of new / old m3mts'. The problem is: here you already have KCP and 3 control plane machines under it created and 3 metal3 machines / 3 BMH one more layer below. Existing m3machines are created using existing m3mt. Now, you are creating a new m3mt here and pointing to it from the KCP, however controller logic to check the node reuse is set or not from the m3m point of view and from which m3mt it was created from. So, the logic would be: getting the exisitng m3m, find it is m3mt which it is created from and check for node reuse field. In your case, if you want your new m3mt to be used properly, just pointing from KCP to the new one is not enough. What you have to do to use it is, scale down the machines (as n consequence m3m which was using old m3mt being deleted) and scale up (as a consequence m3m would use new template you are pointing).

furkatgofurov7 (changes requested):
/hold
To discuss the introduction of new m3mt in the node reuse tests.

kashifest:
> /hold To discuss the introduction of new m3mt in the node reuse tests.

/approve cancel

test/e2e/node_reuse.go
furkatgofurov7: Based on the offline call we had with @Sunnatillo, this should be fine to leave as is. Patching the same template should not hurt, but creating a new template is also fine and may be a better approach.

test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/onsi/gomega/format"
+	corev1 "k8s.io/api/core/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Kubernetes (version) k8s_upgrade in target nodes [k8s_upgrade]", func() {
+	upgradeKubernetes()
```
mboukhalfa, line 22: Would you please add UpgradeKubernetesInputs like https://github.com/metal3-io/cluster-api-provider-metal3/blob/fa6a05594977cdf7511a1e0d2cd8e57b86f378f1/test/e2e/node_reuse.go#L27 so we can use local vars for e2e namespace .. instead of global

test/e2e/config/e2e_conf.yaml
```diff
@@ -111,7 +111,7 @@ variables:
   KUBERNETES_VERSION: "v1.25.2"
   # INIT_WITH_KUBERNETES_VERSION will be used here
   # https://github.com/kubernetes-sigs/cluster-api/blob/bb377163f141d69b7a61479756ee96891f6670bd/test/e2e/clusterctl_upgrade.go#L170
-  INIT_WITH_KUBERNETES_VERSION: ${KUBERNETES_VERSION}
+  INIT_WITH_KUBERNETES_VERSION: "v1.23.8"
```
mboukhalfa, line 114: Please add a comment explaining that this var used by management cluster upgrade and it is fixed to 1.23.8 because v1a4 capi does not support k8v version higher

test/e2e/upgrade_kubernetes_test.go
Sunnatillo: done

test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,236 @@
...
+	WaitForNumMachinesInState(ctx, clusterv1.MachinePhaseRunning, WaitForNumInput{
+		Client:    clusterClient,
+		Options:   []client.ListOption{client.InNamespace(namespace)},
+		Replicas:  numberOfAllBmh,
+		Intervals: e2eConfig.GetIntervals(specName, "wait-machine-running"),
+	})
+
+	format.MaxLength = 0
+
+	// Verify that all nodes is using the new boot-image
```
mboukhalfa, line 226:
```suggestion
	// Verify that all nodes is using the new boot-image k8s version
```

Sunnatillo: done

mboukhalfa (commented):
It looks good to me can you change the ginkgo focus label down here

test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/onsi/gomega/format"
+	corev1 "k8s.io/api/core/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Kubernetes (version) k8s_upgrade in target nodes [k8s_upgrade]", func() {
```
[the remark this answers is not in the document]
Sunnatillo: done

furkatgofurov7 (commented):
One general question: what is the difference between the node reuse tests we already have and the new k8s upgrade tests? I get a feeling that we are duplicating the same workflow in both..

test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,236 @@
...
+		Intervals: e2eConfig.GetIntervals(specName, "wait-bmh-deprovisioning"),
+	})
+
+	Byf("Wait until %d BMH(s) in provisioning state", numberOfWorkers)
+	WaitForNumBmhInState(ctx, bmov1alpha1.StateProvisioning, WaitForNumInput{
+		Client:    clusterClient,
+		Options:   []client.ListOption{client.InNamespace(namespace)},
+		Replicas:  numberOfWorkers,
+		Intervals: e2eConfig.GetIntervals(specName, "wait-bmh-deprovisioning"),
+	})
```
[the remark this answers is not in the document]
Sunnatillo: Agree. Changing it to 1

test/e2e/upgrade_kubernetes_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	Byf("Wait until the worker machine becomes running")
+	WaitForNumMachinesInState(ctx, clusterv1.MachinePhaseRunning, WaitForNumInput{
+		Client:    clusterClient,
+		Options:   []client.ListOption{client.InNamespace(namespace)},
+		Replicas:  numberOfAllBmh,
+		Intervals: e2eConfig.GetIntervals(specName, "wait-machine-running"),
+	})
+
+	format.MaxLength = 0
```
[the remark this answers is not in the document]
Sunnatillo: I guess When I deleted some functions it was leftover. Thank you for catching it.
Deleted it now.

Sunnatillo:
> One general question: what is the difference between the node reuse tests we already have and the new k8s upgrade tests? I get a feeling that we are duplicating the same workflow in both..

You have got a point. Actually we wanted to have separate test for k8s upgrade.  That was something agreed on the discussion with group. Somehow we ended up in an idea where we add k8s upgrade to management upgrade workflow.

test/e2e/upgrade_kubernetes_test.go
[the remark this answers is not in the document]
lentzi90: Seems this was not changed yet?

furkatgofurov7 (approved):
/approve

furkatgofurov7:
> * fixes the issue of patching new image variables to m3machitemplate in node reuse test, instead creates new m3mtemplate with new variables

can you please reword this comment in PR description? it is not that node reuse tests are fixed but they are using a different approach (instead of patching same template, new one is created and used later, but that is not **fixing** something but changing the workflow slightly)

test/e2e/upgrade_kubernetes_test.go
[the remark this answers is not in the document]
Sunnatillo: It should be ok now

Sunnatillo:
> > * fixes the issue of patching new image variables to m3machitemplate in node reuse test, instead creates new m3mtemplate with new variables
> 
> can you please reword this comment in PR description? it is not that node reuse tests are fixed but now they are using a different approach (instead of patching same template, new one is created and used later, but that is not **fixing** something but changing the workflow slightly) so it is better to clarify that.

done

lentzi90:
Let's run also node reuse test since we make changes to it.
/test-ubuntu-e2e-main

lentzi90 (commented):
/lgtm

PR #752, opened 2022-10-07
🌱 E2e: fix upgrade, don't print cluster template

lentzi90:

- The cluster template is huge and clutters the logs. Skip printing it.
- The upgrade test for k8s was waiting for 4 BMHs to be provisioned when in fact the cluster only has 2 machines.

### DIFF
--- a/test/e2e/upgrade_management_cluster_test.go
+++ b/test/e2e/upgrade_management_cluster_test.go
@@ -82,7 +82,7 @@ func upgradeManagementCluster() {
 
 		Expect(upgradeClusterTemplate).ToNot(BeNil(), "Failed to get the cluster template")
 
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade_management_cluster_test.go
```diff
@@ -179,7 +179,7 @@ func upgradeManagementCluster() {
 		waitForNumBmhInState(ctx, bmov1alpha1.StateProvisioned, waitForNumInput{
 			Client:    upgradeClusterClient,
 			Options:   []client.ListOption{client.InNamespace(namespace)},
-			Replicas:  numberOfAllBmh,
+			Replicas:  2,
```
furkatgofurov7, line 182: This might be worth parameterizing, but maybe in the follow-up PR.

furkatgofurov7 (approved):
/approve

PR #756, opened 2022-10-12
📖 Add IP reuse documentation

furkatgofurov7:
Documents IP Reuse feature in https://github.com/metal3-io/cluster-api-provider-metal3/pull/656

### DIFF
--- a/docs/ip_reuse.md
+++ b/docs/ip_reuse.md
@@ -1,0 +1,68 @@
+# IP reuse
+
+## What is IP reuse
+
+As a developer or user there could be a need to make IPAddresses that will be attached to the Kubernetes nodes in the chain of: **BareMetalHost <=> Metal3Machine <=> CAPI Machine <=> Node** objects predictable, for example during the upgrade process and the main goal of this feature is to make that possible.
+
+## Logic behind
+
+As is now, IPPool is an object representing a set of IPAddress pools to be used for IPAddress allocations. An IPClaim is an object representing a request for an IPAddress allocation. Consequently, the IPClaim object name is structured as following:
+
+**IPClaimName** = **Metal3DataName** + **(-)** + **IPPoolName**
+
+Example: metal3datatemplate-0-pool0
+
+The `Metal3DataName` is derived from the `Metal3DataTemplateName` with an added index (`Metal3DataTemplateName-index`), and the `IPPoolName` comes from the IPPool object directly. (See the [IP Address manager repo](https://github.com/metal3-io/ip-address-manager) for more details on these objects).
+In the CAPM3 workflow, when a Metal3Machine is created and a Metal3Data object is requested, the
+process of choosing an `index` to be appended to the name of the `Metal3DataTemplateName`, is random.
+For example, let's imagine we have two Metal3Machines: `metal3machine-0` and `metal3machine-1`
+which creates the following `metal3datatemplate-0` and `metal3datatemplate-1` Metal3Data objects respectively. However, if two nodes are being upgraded at a time, there is no guarantee that same indices will be appended to the respective objects and in fact it can be in completely reverse order (i.e. `metal3machine-0` will get `m3datatemplate-1` and `metal3machine-1` will get `m3datatemplate-0`).
+In order to make it predictable, we structure IPClaim object name using the BareMetalHost name, as following:
+
+**IPClaimName** = **BareMetalHostName** + **(-)** + **IPPoolName**
+
+Example: node-0-pool0
+
+Now, the first part consists of `BareMetalHostName` which is the name of the BareMetalHost object, and should always stay the same once created (predictable). The second part of it is kept unchanged.
+
+## What is the use of PreAllocations field
+
+Once we have a predictable `IPClaimName`, we can make use of a
+`PreAllocations map[string]IPAddressStr` field in the IPPool object to achieve our goal.
+
+We simply add the claim name(s) using the new format (BareMetalHost name included) to the `preAllocations` field in the `IPPool`, i.e:
+
+```yaml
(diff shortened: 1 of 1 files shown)


docs/ip_reuse.md
```diff
@@ -0,0 +1,53 @@
...
+node-0-pool0 : 192.168.1.1
+node-1-pool0: 192.168.1.2
+
+```
+
+Since claim names include BMH names on them, we are able to predict an IP address assigned to specific node.
+
+## How to enable IP Reuse
+
+To enable the feature, a boolean flag called `enableBMHNameBasedPreallocation` is made configurable via clusterctl and it can be passed to clusterctl configuration file
```
lentzi90, line 46:
```suggestion
To enable the feature, a boolean flag called `enableBMHNameBasedPreallocation` was added. It is configurable via clusterctl and it can be passed to the clusterctl configuration file
```

docs/ip_reuse.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Logic behind
+
+As is now, M3IPPool is an object representing a set of IP address pools to be used for IP address allocations. An M3IPClaim is an object representing a request for an IP address allocation. Consequently, the M3IPClaim name is structured as following:
+
+**M3IPClaimName** = **M3DataName** + **(-)** + **M3IPPoolName**
+
+Example: m3datatemplate-0-pool0
+
+First part in `M3IPClaimName`, `M3DataName` is always created in a format of `M3DataTemplateName-index` and second part is `M3IPPoolName` which is the name of Metal3IPPool object defined in the [IP Address manager repo](https://github.com/metal3-io/ip-address-manager).
```
lentzi90, line 15: This part was a bit hard to understand I think. Tried to make it a bit easier in this suggestion:
```suggestion
The `M3DataName` is derived from the `M3DataTemplateName` with an added index (`M3DataTemplateName-index`), and the `M3IPPoolName` comes from the Metal3IPPool object directly. (See the [IP Address manager repo](https://github.com/metal3-io/ip-address-manager) for more details on these objects).
```

docs/ip_reuse.md
```diff
@@ -0,0 +1,53 @@
...
+In the CAPM3 workflow, when Metal3Machine is created and M3Data is requested, the
+process of choosing an `index` to be appended in the name of `M3DataTemplateName` is random.
+For example, let's imagine we have two Metal3Machines: `metal3machine-0` and `metal3machine-1`
+which creates following `m3datatemplate-0` and `m3datatemplate-1` Metal3Data objects respectively. However, if two nodes are being upgraded at a time, there is no guarantee that same indexes will be appended to respective objects and in fact it can be in completely reverse order (i.e. `metal3machine-0` gets `m3datatemplate-1`).
+In order to make it predictable, we structure M3IPClaim name using the BareMetalHost name, as following:
```
lentzi90, line 20:
```suggestion
In the CAPM3 workflow, when a Metal3Machine is created and a M3Data is requested, the
process of choosing an `index` to be appended to the name of the `M3DataTemplateName`, is random.
For example, let's imagine we have two Metal3Machines: `metal3machine-0` and `metal3machine-1`
which creates the following `m3datatemplate-0` and `m3datatemplate-1` Metal3Data objects respectively. However, if two nodes are being upgraded at a time, there is no guarantee that same indices will be appended to the respective objects and in fact it can be in completely reverse order (i.e. `metal3machine-0` gets `m3datatemplate-1`).
In order to make it predictable, we structure M3IPClaim name using the BareMetalHost name, as following:
```

docs/ip_reuse.md
```diff
@@ -0,0 +1,53 @@
...
+
+**M3IPClaimName** = **BMHName** + **(-)** + **M3IPPoolName**
+
+Example: node-0-pool0
+
+Now, first part consists of `BMHName` which is the name of the BareMetalHost object, and should always stay the same once created (predictable). Second part of it is kept unchanged.
```
lentzi90, line 26:
```suggestion
Now, the first part consists of `BMHName` which is the name of the BareMetalHost object, and should always stay the same once created (predictable). The second part of it is kept unchanged.
```

docs/ip_reuse.md
```diff
@@ -0,0 +1,53 @@
...
+
+## What is the use of PreAllocations field
+
+Once we have a predictable `M3IPClaimName`, we can make use of a
+`PreAllocations map[string]IPAddressStr` field in the Metal3IPPool object to achieve our goal.
+
+We simply fill in `M3IPPool` object with claim names and IPs, i.e:
```
lentzi90, line 33:
```suggestion
We simply add the claim names and IPs to the `preAllocations` in the `M3IPPool`, i.e:
```

docs/ip_reuse.md
```diff
@@ -0,0 +1,53 @@
...
+
+```yaml
+
+node-0-pool0 : 192.168.1.1
+node-1-pool0: 192.168.1.2
+
+```
```
lentzi90, line 40: Could you add a bit more here so it is possible to see where in the IPPool these should be added?

lentzi90 (commented):
Nice work with this! I added some suggestions below that should hopefully make it even easier to understnad, but I think it is already in good shape!

docs/ip_reuse.md
furkatgofurov7: Sure, updated

docs/ip_reuse.md
furkatgofurov7: Good point, added the IPPool object definition with more fields

furkatgofurov7:
@lentzi90 thanks for the review, this should be good now.

lentzi90 (commented):
/lgtm

kashifest (approved):

PR #760, opened 2022-10-14
🌱 Format e2e table logs

mboukhalfa:
Impliment suggestions discussed here https://github.com/metal3-io/cluster-api-provider-metal3/pull/714#pullrequestreview-1117934522.
So instead of : 
```
INFO: Listing Nodes:
INFO: Name	Status	Version
INFO: ---------------------------------------------------------------------------------
INFO: test1-2qn2z	Ready	v1.23.8
INFO: test1-6b5bd59967-kt87d	Ready	v1.23.8
INFO: test1-mrg4m	Ready	v1.23.8
INFO: test1-w7n7w	NotReady	v1.23.8
INFO: ---------------------------------------------------------------------------------
INFO: 4 Nodes in total
INFO: =================================================================================
```
This will log:
```
Listing Nodes
Name:                   Status:	 Version:
test1-2qn2z	        Ready	  v1.23.8
test1-6b5bd59967-kt87d  Ready	  v1.23.8
test1-mrg4m	        Ready	  v1.23.8
test1-w7n7w	        NotReady  v1.23.8

```

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"text/tabwriter"
 
 	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	infrav1 "github.com/metal3-io/cluster-api-provider-metal3/api/v1beta1"
@@ -37,6 +38,24 @@ func LogFromFile(logFile string) {
 	data, err := os.ReadFile(filepath.Clean(logFile))
 	Expect(err).To(BeNil(), "No log file found")
 	Logf(string(data))
+}
+
+// logTable print a formatted table into the e2e logs.
+func logTable(title string, rows [][]string) {
+	getRowFormatted := func(row []string) string {
+		rowFormatted := ""
+		for i := range row {
+			rowFormatted = fmt.Sprintf("%s\t%s\t", rowFormatted, row[i])
+		}
+		return rowFormatted
+	}
+	w := tabwriter.NewWriter(GinkgoWriter, 0, 0, 2, ' ', tabwriter.TabIndent)
+	fmt.Fprintln(w, title)
+	for _, r := range rows {
+		fmt.Fprintln(w, getRowFormatted(r))
+	}
+	fmt.Fprintln(w, "")
+	w.Flush()
 }
 
 func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterProxy framework.ClusterProxy, artifactFolder string, namespace string, intervalsGetter func(spec, key string) []interface{}, clusterName, clusterctlLogFolder string, skipCleanup bool) {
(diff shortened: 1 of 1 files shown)


mboukhalfa:
test-ubuntu-e2e-parallel-main

test/e2e/common.go
```diff
@@ -39,6 +40,24 @@ func LogFromFile(logFile string) {
...
 
+func logTable(title string, columnNames []string, rows [][]string) {
+	getRowFormatted := func(row []string) string {
+		rowFormatted := row[0]
+		for i := 1; i < len(row); i++ {
+			rowFormatted = fmt.Sprintf("%s\t%s", rowFormatted, row[i])
+		}
+		return rowFormatted
+	}
+	w := tabwriter.NewWriter(GinkgoWriter, 0, 0, 2, ' ', tabwriter.TabIndent)
```
lentzi90, line 51: A comment to explain this would be good. From the [docs](https://pkg.go.dev/text/tabwriter#Writer.Init) I understand that this means 0 minwidth, 0 tabwidth, 2 padding and one space used as padchar. But why should tabwidth be 0? I would expect that we set it to maybe 2 or 4.

test/e2e/common.go
```diff
@@ -39,6 +40,24 @@ func LogFromFile(logFile string) {
 	Logf(string(data))
 }
 
+func logTable(title string, columnNames []string, rows [][]string) {
+	getRowFormatted := func(row []string) string {
+		rowFormatted := row[0]
+		for i := 1; i < len(row); i++ {
+			rowFormatted = fmt.Sprintf("%s\t%s", rowFormatted, row[i])
+		}
```
lentzi90, line 48: From the [docs](https://pkg.go.dev/text/tabwriter#Writer) it looks like it is important that each cell ends with a tab.

```suggestion
		rowFormatted := ""
		for i, _ := range row {
			rowFormatted = fmt.Sprintf("%s\t%s\t", rowFormatted, row[i])
		}
```

test/e2e/common.go
```diff
@@ -39,6 +40,24 @@ func LogFromFile(logFile string) {
 	Logf(string(data))
 }
 
+func logTable(title string, columnNames []string, rows [][]string) {
```
lentzi90, line 43: Why are columnNames separate from rows? We are anyway printing everything as one big table. I would suggest just skipping columnNames for simplicity. Speaking about simplicity, I would also suggest skipping the title. It can be printed separately by the caller and then this function becomes minimal and nice, just doing one thing and doing it well. :slightly_smiling_face:

test/e2e/common.go
mboukhalfa: I think because I am not using tabs for padding it is meaningless to set any value there that what example in docs also did

test/e2e/common.go
mboukhalfa: Well I had the same thoughts but as conceptually the names can be formatted  separately like we add a line between them and the other rows we can print then in bold .. I keep them separate for any future nice formatting needed the title also play a role of a comment when calling the function it say what is it about I think it is still simple and nice

lentzi90: Ok the formatting I understand. The title I think is a bit strange. It is not a table. The caller can just print it before calling this and then this whole function becomes better focused on one thing.

test/e2e/common.go
lentzi90: Ah that makes sense! I missed the second example... Never mind then :slightly_smiling_face:

furkatgofurov7:
Does it really need a ✨ label (new feature) for the PR title?

test/e2e/common.go
```diff
@@ -331,7 +340,6 @@ func waitForNumMachinesInState(ctx context.Context, phase clusterv1.MachinePhase
 		return machine.Status.GetTypedPhase() == phase
 	}
 	waitForNumMachines(ctx, inPhase, input)
-	listMachines(ctx, input.Client, input.Options...)
```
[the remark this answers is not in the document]
mboukhalfa: already listed in `waitForNumMachines`

test/e2e/common.go
lentzi90: @mboukhalfa any comment? :point_up:

mboukhalfa: I removed column names from the parameter to be the first row

test/e2e/common.go
```diff
@@ -39,6 +40,23 @@ func LogFromFile(logFile string) {
 	Logf(string(data))
 }
 
+func logTable(title string, rows [][]string) {
```
lentzi90, line 43: Please add a description above

mboukhalfa: done

lentzi90 (commented):
/lgtm

test/e2e/common.go
furkatgofurov7: @mboukhalfa how does it look like now, is it the same as in PR description?

mboukhalfa: still the same you check on the logs

mboukhalfa:
```
  Listing Nodes
  Name:                     Status:    Version:  
  test1-77fdcff7b8-2bfb4    Ready      v1.25.2   
  test1-txbnn               Ready      v1.25.2
```

furkatgofurov7: Okay, I though it is changed now based on the above discussion, fine by me

PR #762, opened 2022-10-17
🌱 Use latest k8s version instead of previous minor version for e2e tests

mboukhalfa:
Co-authored-by: Kashif Khan <kashif.khan@est.tech>

For now all e2e test use the previous minor version as default k8s version instead of the latest.

This PR refactor e2e to use the latest k8s version for all the test except tests that require an old version like `clusterctl upgrade` from `v1alpha5` that requires k8s `v1.23.8` or `nodeReuse` that does k8s upgrade from the previous minor version to the latest

### DIFF
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -49,8 +49,8 @@ else
 fi
 
 export FROM_K8S_VERSION="v1.24.1"
-export KUBERNETES_VERSION=${FROM_K8S_VERSION}
-export UPGRADED_K8S_VERSION="v1.25.2"
+export KUBERNETES_VERSION="v1.25.2"
+
 # Can be overriden from jjbs
 export CAPI_VERSION=${CAPI_VERSION:-"v1beta1"}
 export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
@@ -62,16 +62,7 @@ export PATH=$PATH:$HOME/.krew/bin
 # Upgrade test environment vars and config
 if [[ ${GINKGO_FOCUS:-} == "upgrade" ]]; then
   export NUM_NODES=${NUM_NODES:-"5"}
-  export FROM_K8S_VERSION="v1.23.8"
-  export KUBERNETES_VERSION=${FROM_K8S_VERSION}
-fi
-
-if [[ "${IMAGE_OS}" == "ubuntu" ]]; then
-  export UPGRADED_IMAGE_NAME="UBUNTU_22.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
-  export UPGRADED_RAW_IMAGE_NAME="UBUNTU_22.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
-else
-  export UPGRADED_IMAGE_NAME="CENTOS_9_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
-  export UPGRADED_RAW_IMAGE_NAME="CENTOS_9_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
+  export KUBERNETES_VERSION="v1.23.8"
 fi
 
 # Exported to the cluster templates
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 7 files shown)


test/e2e/common.go
```diff
@@ -64,6 +65,56 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
...
 
+const (
+	artifactoryURL = "https://artifactory.nordix.org/artifactory/metal3/images/k8s"
+	imagesURL      = "http://172.22.0.1/images"
+	ironicImageDir = "/opt/metal3-dev-env/ironic/html/images"
+)
+
+func downloadImage(k8sVersion string) (imageURL string, imageChecksum string) {
+	osType := strings.ToLower(os.Getenv("OS"))
+	Expect(osType).ToNot(Equal(""))
```
lentzi90, line 76: Maybe better to make this explicit
```suggestion
	Expect(osType).To(BeElementOf([]string{"ubuntu", "centos"}))
```

test/e2e/common.go
```diff
@@ -64,6 +65,56 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
 	}
 }
 
+const (
+	artifactoryURL = "https://artifactory.nordix.org/artifactory/metal3/images/k8s"
+	imagesURL      = "http://172.22.0.1/images"
+	ironicImageDir = "/opt/metal3-dev-env/ironic/html/images"
+)
+
+func downloadImage(k8sVersion string) (imageURL string, imageChecksum string) {
```
lentzi90, line 74: Since we are actually checking and only downloading if necessary, I would suggest renaming this to `getImageUrl` or `ensureImage` or similar. Please also add a description.

test/e2e/common.go
mboukhalfa: done

test/e2e/common.go
mboukhalfa: done

lentzi90 (commented):
/lgtm

scripts/environment.sh
```diff
@@ -66,14 +66,6 @@ if [[ ${GINKGO_FOCUS:-} == "upgrade" ]]; then
   export KUBERNETES_VERSION=${FROM_K8S_VERSION}
```
kashifest, line 66: I think we don't need these two vars, probably these can be removed now.

kashifest (changes requested):
Thanks a lot for this change. I think this is going to right direction. Few comments in line.

test/e2e/common.go
```diff
@@ -64,6 +65,56 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
...
+func ensureImage(k8sVersion string) (imageURL string, imageChecksum string) {
+	osType := strings.ToLower(os.Getenv("OS"))
+	Expect(osType).To(BeElementOf([]string{"ubuntu", "centos"}))
+	imageNamePrefix := "CENTOS_9_NODE_IMAGE_K8S"
+	if osType != "centos" {
+		imageNamePrefix = "UBUNTU_22.04_NODE_IMAGE_K8S"
+	}
+	imageName := fmt.Sprintf("%s_%s.qcow2", imageNamePrefix, k8sVersion)
+	Logf("IMAGE_NAME: %v", imageName)
+	rawImageName := fmt.Sprintf("%s_%s-raw.img", imageNamePrefix, k8sVersion)
```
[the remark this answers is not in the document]
mboukhalfa: probably this adds the `.qcow2` which we do not want in the raw image name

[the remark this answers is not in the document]
kashifest: Ah true, ignore the suggestion.

test/e2e/common.go
```diff
@@ -64,6 +65,56 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
...
+		err = downloadFile(imagePath, fmt.Sprintf("%s/%s", imageLocation, imageName))
+		Expect(err).To(BeNil())
+		cmd := exec.Command("qemu-img", "convert", "-O", "raw", imagePath, rawImagePath) // #nosec G204:gosec
+		err = cmd.Run()
+		Expect(err).To(BeNil())
+		cmd = exec.Command("md5sum", rawImagePath) // #nosec G204:gosec
+		output, err := cmd.CombinedOutput()
+		Expect(err).To(BeNil())
+		md5sum := strings.Fields(string(output))[0]
+		err = os.WriteFile(fmt.Sprintf("%s/%s.md5sum", ironicImageDir, rawImageName), []byte(md5sum), 0544)
```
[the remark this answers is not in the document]
mboukhalfa: it is gosec complaining https://github.com/metal3-io/cluster-api-provider-metal3/pull/762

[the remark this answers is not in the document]
kashifest: Makes sense, probably we should do the same in ansible as well to not get confused in future.

scripts/environment.sh
mboukhalfa: done

lentzi90 (commented):
/lgtm

lentzi90 (commented):
/lgtm

kashifest (approved):

furkatgofurov7 (commented):
To make sure where on track with upgrade and normal e2e tests as well

/test-upgrade-e2e-main
/test-ubuntu-e2e-main

PR #765, opened 2022-10-17
Add cleaning to upgrade test

mboukhalfa:


test/e2e/upgrade_management_cluster_test.go
```diff
@@ -26,18 +27,19 @@ var _ = Describe("When testing cluster upgrade v1alpha5 > current [upgrade]", fu
...
-			PreInit:                   preInitFunc,
-			PreWaitForCluster:         preWaitForCluster,
-			PreUpgrade:                preUpgrade,
-			MgmtFlavor:                osType,
-			WorkloadFlavor:            osType,
+			E2EConfig:                   e2eConfig,
+			ClusterctlConfigPath:        clusterctlConfigPath,
+			BootstrapClusterProxy:       bootstrapClusterProxy,
+			ArtifactFolder:              artifactFolder,
+			SkipCleanup:                 skipCleanup,
```
lentzi90, line 34: I'm trying to understand if this is changing something or if it will be the same. What is the default for `skipCleanup`? I'm guessing it is `false`, but we had `true` hard coded before. :thinking:

mboukhalfa: We had true hard coded before because we where skipping  cluster deletion from bootstrap. And we had to that because capi function did allow us to move back ironic to the bootstrap which is  required for cluster deletion.

After injecting this new hook we are able to move back ironic then we delete the cluster at the end or we can skip based on the var `skipCleanup`

lentzi90: Thanks for the explanation! You should put this in the description of the PR!

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -152,3 +154,35 @@ func preUpgrade(clusterProxy framework.ClusterProxy) {
...
+// it moves back Ironic to the bootstrap cluster.
+func preCleanupManagementCluster(clusterProxy framework.ClusterProxy) {
+	// Remove ironic
+	removeIronic := func() {
+		By("Remove Ironic containers from the target cluster")
+		deploymentName := e2eConfig.GetVariable("NAMEPREFIX") + "-ironic"
+		ironicNamespace := e2eConfig.GetVariable("IRONIC_NAMESPACE")
+		err := clusterProxy.GetClientSet().AppsV1().Deployments(ironicNamespace).Delete(ctx, deploymentName, metav1.DeleteOptions{})
+		Expect(err).To(BeNil(), "Failed to delete Ironic from the target cluster")
+	}
```
lentzi90, line 168: It would be nicer to re-use [removeIronicDeploymentOnTarget](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/pivoting_test.go#L217-L222) instead of creating a new function for this

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):

PR #795, opened 2022-12-01
✨ Support namespace in Metal3Machine's HostSelector

Lirt:
## What this PR does / why we need it

This PR adds support for `namespace` field in a `HostSelector` of `Metal3Machine` and `Metal3MachineTemplate` CRDs.

Short discussion about this feature was in [this Slack thread](https://kubernetes.slack.com/archives/CHD49TLE7/p1669736106751629).

cc. @jsen- and @schrej 

## Motivation

In case each `CAPI`/`Metal3` cluster (or tenant) is located in a dedicated namespace - it is not possible to have a shared pool of BMHs among those clusters.

This feature will allow to pick `BareMetalHosts` (BMHs) from different namespace than namespace of a `Metal3Machine` (or `Cluster` namespace). With namespace selection feature each cluster (or tenant) can be isolated in a dedicated namespace and reference BMHs from different namespace in which they can be pre-provisioned. 

Signed-off-by: Ondrej Vasko <ondrej.vaskoo@gmail.com>

### DIFF
--- a/api/v1alpha5/common_types.go
+++ b/api/v1alpha5/common_types.go
@@ -43,6 +43,11 @@ type HostSelector struct {
 
 	// Label match expressions that must be true on a chosen BareMetalHost
 	MatchExpressions []HostSelectorRequirement `json:"matchExpressions,omitempty"`
+
+	// Namespace where to look for BareMetalHosts. If empty, look for
+	// BareMetalHosts in the namespace of Metal3Machine
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
 }
 
 type HostSelectorRequirement struct {
--- a/api/v1alpha5/metal3machine_types.go
+++ b/api/v1alpha5/metal3machine_types.go
@@ -45,9 +45,9 @@ type Metal3MachineSpec struct {
 	// namespace if not specified.
 	UserData *corev1.SecretReference `json:"userData,omitempty"`
 
-	// HostSelector specifies matching criteria for labels on BareMetalHosts.
-	// This is used to limit the set of BareMetalHost objects considered for
-	// claiming for a metal3machine.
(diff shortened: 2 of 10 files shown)


Lirt:
After more tests I found out that after `Metal3Machine` was created, an `ownerReference` was added to `BareMetalHost` telling that owner of `BMH` is `M3M`. It looks like this:
```yaml
API Version:  metal3.io/v1alpha1
Kind:         BareMetalHost
Name:         s4anb866
Namespace:    c-caas-bm-test1
Labels:       cluster.x-k8s.io/cluster-name=ovasko-test-cluster-01
Metadata:
  Owner References:
    API Version:     infrastructure.cluster.x-k8s.io/v1beta1
    Controller:      true
    Kind:            Metal3Machine
    Name:            ovasko-test-cluster-01-master-wb99h
    UID:             48464ad2-5674-4f2a-b02b-a01783595d7a
```

```bash
  Warning  OwnerRefInvalidNamespace  4m    garbage-collector-controller  ownerRef [infrastructure.cluster.x-k8s.io/v1beta1/Metal3Machine, namespace: c-caas-bm-test1, name: ovasko-test-cluster-01-master-wb99h, uid: 48464ad2-5674-4f2a-b02b-a01783595d7a] does not exist in namespace "c-caas-bm-test1"
```

The issue is that ownerReference can only point to resource in the same namespace or cluster-wide resource:
> OwnerReference contains enough information to let you identify an owning
> object. An owning object must be in the same namespace as the dependent, or
> be cluster-scoped, so there is no namespace field.

And then BMH gets deleted as M3M could not be found (it's in different namespace for this case). I think this is a deal-breaker for this Pull Request and we cannot overcome it unless someone has idea how to combine this BMO feature with CAPM3.

PR #797, opened 2022-12-02
🐛hack: download latest CAPI for tilt

tuminoid:
This logic change matches metal3-dev-env PR#1118.

Rely on Github's release order + prefix filter, instead of jq's max() which sorts v1.3.0-rc1 as newer than v1.3.0, which is wrong for the versioning convention we use.

/assign furkatgofurov7

### DIFF
--- a/hack/gen_tilt_settings.sh
+++ b/hack/gen_tilt_settings.sh
@@ -25,7 +25,8 @@ function get_latest_release() {
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
   # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\"))] | max ")"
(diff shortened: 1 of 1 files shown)


hack/gen_tilt_settings.sh
```diff
@@ -25,7 +25,8 @@ function get_latest_release() {
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
   # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\"))] | max ")"
+  # This relies on the order of the Github release page + filter, max() sorts rc versions newer than releases
```
furkatgofurov7, line 28: We are dropping a max(), but not sure why it is referenced again in the comment? 

/approve

furkatgofurov7 (approved):

hack/gen_tilt_settings.sh
tuminoid: Yeah makes no sense to leave comment on code that was removed. Fixed.

PR #798, opened 2022-12-05
🌱 Add integration test to e2e

mquhuy:
This PR adds ci-provisioning and provisioning (which currently can be run only from dev-env using ansible) to e2e tests.

Also added logs collections for all containers before pivoting, which is the behavior in dev-env pivoting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,6 +168,7 @@ _SKIP_ARGS := $(foreach arg,$(strip $(GINKGO_SKIP)),-ginkgo.skip="$(arg)")
 endif
 .PHONY: e2e-tests
 e2e-tests: CONTAINER_RUNTIME?=docker ## Env variable can override this default
+export CONTAINER_RUNTIME
 e2e-tests: e2e-substitutions cluster-templates ## This target should be called from scripts/ci-e2e.sh
 	for image in $(E2E_CONTAINERS); do \
 		$(CONTAINER_RUNTIME) pull $$image; \
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -101,6 +101,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	// Before all ParallelNodes.
 
 	Expect(configPath).To(BeAnExistingFile(), "Invalid test suite argument. e2e.config should be an existing file.")
+	Expect(os.RemoveAll(artifactFolder)).To(Succeed())
 	Expect(os.MkdirAll(artifactFolder, 0755)).To(Succeed(), "Invalid test suite argument. Can't create e2e.artifacts-folder %q", artifactFolder)
 
 	By("Initializing a runtime.Scheme with all the GVK relevant for this test")
--- a/test/e2e/integration.go
+++ b/test/e2e/integration.go
@@ -1,0 +1,40 @@
+package e2e
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func createDirIfNotExist(dirPath string) {
+	err := os.MkdirAll(dirPath, 0750)
+	if err != nil && os.IsNotExist(err) {
+		log.Fatal(err)
+	}
+}
+
+func fetchContainerLogs(containerNames *[]string) {
+	By("Create directories and storing container logs")
+	for _, name := range *containerNames {
+		containerRuntime := e2eConfig.GetVariable("CONTAINER_RUNTIME")
+		logDir := filepath.Join(artifactFolder, containerRuntime, name)
+		By(fmt.Sprintf("Create log directory for container %s at %s", name, logDir))
+		createDirIfNotExist(logDir)
+		By(fmt.Sprintf("Fetch logs for container %s", name))
+		cmd := exec.Command("sudo", containerRuntime, "logs", name) // #nosec G204:gosec
+		out, err := cmd.Output()
+		if err != nil {
+			writeErr := os.WriteFile(filepath.Join(logDir, "stderr.log"), []byte(err.Error()), 0444)
+			Expect(writeErr).ToNot(HaveOccurred())
+			log.Fatal(err)
+		}
+		writeErr := os.WriteFile(filepath.Join(logDir, "stdout.log"), out, 0444)
+		Expect(writeErr).ToNot(HaveOccurred())
+	}
+}
+
--- a/test/e2e/integration_test.go
+++ b/test/e2e/integration_test.go
@@ -1,0 +1,71 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/cluster-api/test/framework"
+)
+
+var _ = Describe("When testing integration [integration]", func() {
+	It("CI Test Provision", func() {
+		numberOfWorkers       = int(*e2eConfig.GetInt32PtrVariable("WORKER_MACHINE_COUNT"))
+		numberOfControlplane = int(*e2eConfig.GetInt32PtrVariable("CONTROL_PLANE_MACHINE_COUNT"))
+		k8sVersion := e2eConfig.GetVariable("KUBERNETES_VERSION")
+		By("Provision Workload cluster")
+		targetCluster, result := createTargetCluster(k8sVersion)
+		By("Pivot objects to target cluster")
+		pivoting(ctx, func() PivotingInput {
+			return PivotingInput{
+				E2EConfig:             e2eConfig,
+				BootstrapClusterProxy: bootstrapClusterProxy,
+				TargetCluster:         targetCluster,
+				SpecName:              specName,
+				ClusterName:           clusterName,
+				Namespace:             namespace,
+				ArtifactFolder:        artifactFolder,
+				ClusterctlConfigPath:  clusterctlConfigPath,
+			}
+		})
+		By("Repivot objects to the source cluster")
+		rePivoting(ctx, func() RePivotingInput {
+			return RePivotingInput{
+				E2EConfig:             e2eConfig,
+				BootstrapClusterProxy: bootstrapClusterProxy,
+				TargetCluster:         targetCluster,
+				SpecName:              specName,
+				ClusterName:           clusterName,
+				Namespace:             namespace,
+				ArtifactFolder:        artifactFolder,
+				ClusterctlConfigPath:  clusterctlConfigPath,
+			}
+		})
+		By("Deprovision target cluster")
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+		intervals := e2eConfig.GetIntervals(specName, "wait-deprovision-cluster")
+		// In pivoting step we labeled the BMO CRDs (so that the objects are moved 
+		// by CAPI pivoting feature), which made CAPI DeleteClusterAndWait()
+		// fail as it has a check to make sure all resources managed by CAPI
+		// is gone after Cluster deletion. Therefore, we opted not to use
+		// DeleteClusterAndWait(), but only delete the cluster and then wait 
+		// for it to be deleted.
+		framework.DeleteCluster(ctx, framework.DeleteClusterInput{
+			Deleter: bootstrapClient,
+			Cluster: result.Cluster,
+		})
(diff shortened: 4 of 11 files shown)


mboukhalfa:
> /test-ubuntu-integration-main

this will still run the ansible test
use:
/test-ubuntu-e2e-main

Makefile
```diff
@@ -133,6 +133,7 @@ GINKGO_NOCOLOR ?= false
 ARTIFACTS ?= $(ROOT_DIR)/_artifacts
 E2E_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/e2e_conf.yaml
 E2E_OUT_DIR ?= $(ROOT_DIR)/test/e2e/_out
+export E2E_OUT_DIR
```
mboukhalfa, line 136: To store logs, instead of exporting this I would prefer to use `artifactFolder` [already exists in the e2e](https://github.com/metal3-io/cluster-api-provider-metal3/blob/e735768beba749e46b43c143e71c8fad6161a52f/test/e2e/e2e_suite_test.go#L42)

Makefile
```diff
@@ -174,6 +175,7 @@ _SKIP_ARGS := $(foreach arg,$(strip $(GINKGO_SKIP)),-ginkgo.skip="$(arg)")
 endif
 .PHONY: e2e-tests
 e2e-tests: CONTAINER_RUNTIME?=docker ## Env variable can override this default
+export CONTAINER_RUNTIME
```
mboukhalfa, line 178: No need to export this already done here https://github.com/metal3-io/cluster-api-provider-metal3/blob/3c299790d3af391e85de6b64852c2fd5ae0a953c/scripts/environment.sh#L39-L42

test/e2e/integration_test.go
```diff
@@ -0,0 +1,116 @@
...
+	})
+	It("CI Test Provision", func() {
+		// ciTestProvision()
+		healthCheck()
+	})
+})
+
+// Get variable values from env
+func initVars() {
+	Expect(err).NotTo(HaveOccurred())
```
mboukhalfa, line 37: what kind of error you are trying to catch I also do not think we need to init these vars in the before each can be done in the It block

test/e2e/integration_test.go
```diff
@@ -0,0 +1,116 @@
...
+			SpecName:              specName,
+			ClusterName:           clusterName,
+			Namespace:             namespace,
+			ArtifactFolder:        artifactFolder,
+			ClusterctlConfigPath:  clusterctlConfigPath,
+		}
+	})
+}
+
+func healthCheck() {
```
mboukhalfa, line 105: What tasks from ansible integration you are trying to implement here I do not see healthcheck tasks in integration logs at least

test/e2e/integration_test.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+var (
+	k8sVersion          string
+	containerRuntime string
+	err                 error
+)
+
+var _ = Describe("When testing integration [integration]", func() {
+	BeforeEach(func() {
```
mboukhalfa, line 26: Please not that we select the cluster template based on `osType`  https://github.com/metal3-io/cluster-api-provider-metal3/blob/3c299790d3af391e85de6b64852c2fd5ae0a953c/test/e2e/pivoting_based_feature_test.go#L31 
that is used in `createTargetCluster` function 
https://github.com/metal3-io/cluster-api-provider-metal3/blob/3c299790d3af391e85de6b64852c2fd5ae0a953c/test/e2e/pivoting_based_feature_test.go#L134

mboukhalfa (changes requested):
I skimmed this looks already most needed parts here some feedback suggestions

Makefile
mquhuy: Thanks. I'll use that one

mquhuy:
/retitle :sparkles:  Add ci-provision test to e2e

test/e2e/integration_test.go
```diff
@@ -0,0 +1,59 @@
...
+		updateCalico(e2eConfig, cniPath, "enp2s0")
+	}
+	By("Verify Calico pods running")
+	targetCluster = bootstrapClusterProxy.GetWorkloadCluster(ctx, namespace, clusterName)
+	targetClusterClientset := targetCluster.GetClientSet()
+	waitForAllPodsRunning(targetClusterClientset, "kube-system", 9, 20)
+	By("Wait for all pods to be in running state")
+	waitForAllPodsRunning(targetClusterClientset, namespace, 150, 20)
+	By("Wait for all nodes to be in ready state")
+	waitForAllNodesReady(targetClusterClientset, 150, 3)
```
mboukhalfa, line 32: We do not need these because we use clusterResourceSet to install the CNI

test/e2e/integration_test.go
```diff
@@ -0,0 +1,59 @@
...
+			E2EConfig:             e2eConfig,
+			BootstrapClusterProxy: bootstrapClusterProxy,
+			TargetCluster:         targetCluster,
+			SpecName:              specName,
+			ClusterName:           clusterName,
+			Namespace:             namespace,
+			ArtifactFolder:        artifactFolder,
+			ClusterctlConfigPath:  clusterctlConfigPath,
+		}
+	})
```
mboukhalfa, line 58: better to have this call directly in It block

test/e2e/integration_test.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+var _ = Describe("When testing integration [integration]", func() {
+	It("CI Test Provision", func() {
+		ciTestProvision()
+	})
+})
+
+func ciTestProvision() {
+	k8sVersion := e2eConfig.GetVariable("KUBERNETES_VERSION")
```
mboukhalfa, line 15: please not that by default e2e uses 4 nodes and intergration tests needs only 2 nodes 
https://github.com/metal3-io/cluster-api-provider-metal3/blob/87f2c8466fd444f6ac5951fcdeed80be5031cbed/scripts/ci-e2e.sh#L27
and it should try to provision only 1 kcp and 1 md
https://github.com/metal3-io/cluster-api-provider-metal3/blob/87f2c8466fd444f6ac5951fcdeed80be5031cbed/test/e2e/config/e2e_conf.yaml#L115-L116

mboukhalfa: and I do not think we need this function `ciTestProvision` you can just move `createTargetCluster` to the It block

mboukhalfa (commented):
Wawoo that's already passing the test did not expected it good job
here is some more comments and do not forget to add deprovisioning

mboukhalfa:
`Provision` tasks in ansible is equivalent to `createTargetCluster` and since this test is doing more than that `pivoting` and `repivoting` test and I expect `deprovisioning` will be added in the same PR let's rename it 
/retitle 🌱 Add integration test to e2e

mquhuy:
> Wawoo that's already passing the test did not expect it good job here is some more comments and do not forget to add deprovisioning

haha, thanks. It still failed sometimes in my trials though.

Makefile
mboukhalfa: you have also to remove the export here

test/e2e/e2e_suite_test.go
```diff
@@ -101,6 +101,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	// Before all ParallelNodes.
 
 	Expect(configPath).To(BeAnExistingFile(), "Invalid test suite argument. e2e.config should be an existing file.")
+    Expect(os.RemoveAll(artifactFolder)).To(Succeed())
```
lentzi90, line 104: Indentation looks strange here. This line is not at the same level as the others.

test/e2e/pivoting_based_feature_test.go
```diff
@@ -145,3 +148,14 @@ func createTargetCluster(k8sVersion string) (targetCluster framework.ClusterProx
...
+func deprovisionTargetCluster(client *client.Client) {
+	c := clusterv1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: clusterName,
+			Namespace: namespace,
+		},
+	}
+	err := (*client).Delete(ctx, &c)
+	Expect(err).ToNot(HaveOccurred())
+}
```
lentzi90, line 161: I think we should use [DeleteCluster](https://pkg.go.dev/sigs.k8s.io/cluster-api/test/framework#DeleteCluster) from the framework instead.

test/e2e/integration_test.go
```diff
@@ -0,0 +1,72 @@
...
+	})
+	intervals = e2eConfig.GetIntervals(specName, "wait-metal3-machine-deprovisioning")
+	waitForResourceGone(&bootstrapClient, &infrav1.Metal3MachineList{}, namespace, intervals)
+	intervals = e2eConfig.GetIntervals(specName, "wait-machine-deprovisioning")
+	waitForResourceGone(&bootstrapClient, &clusterv1.MachineList{}, namespace, intervals)
+	intervals = e2eConfig.GetIntervals(specName, "wait-metal3-cluster-deprovisioning")
+	waitForResourceGone(&bootstrapClient, &infrav1.Metal3ClusterList{}, namespace, intervals)
+	intervals = e2eConfig.GetIntervals(specName, "wait-cluster-deprovisioning")
+	waitForResourceGone(&bootstrapClient, &clusterv1.ClusterList{}, namespace, intervals)
+}
```
lentzi90, line 72: I think we can safely use DeleteCluster from the framework instead of this. It will check that all resources are gone. The only thing left is to check that the BMHs become available.

lentzi90 (commented):
Nice job!
Added some suggestions below.

lentzi90 (commented):
Really nice job!
I think you should mention also the log collection in the PR description. One small nit below, otherwise this is good to go for me :slightly_smiling_face:

mquhuy:
> Really nice job! I think you should mention also the log collection in the PR description. One small nit below, otherwise this is good to go for me slightly_smiling_face

Thank you. I've updated the PR description. What was the small nit were you referring to? It seems I failed to see it :-?

test/e2e/integration.go
```diff
@@ -0,0 +1,52 @@
...
+
+func waitForResourceGone(c *client.Client, list client.ObjectList, namespace string, intervals []interface{}) {
+	Logf("Waiting for all %s to be gone in namespace %s", list.GetObjectKind().GroupVersionKind().Kind, namespace)
+	listOption := client.InNamespace(namespace)
+	Eventually(func(g Gomega) {
+		(*c).List(ctx, list, listOption)
+		metaValue := reflect.ValueOf(list).Elem()
+		g.Expect(metaValue.FieldByName("Items").Len()).To(BeZero())
+	}, intervals...).Should(Succeed())
+}
```
lentzi90, line 51: Sorry, I just wrote the comment but never clicked to publish it. :smile: 
I think this is not used anymore so we can delete it.

lentzi90 (commented):
/lgtm

test/e2e/integration_test.go
```diff
@@ -0,0 +1,69 @@
...
+				Namespace:             namespace,
+				ArtifactFolder:        artifactFolder,
+				ClusterctlConfigPath:  clusterctlConfigPath,
+			}
+		})
+		By("Deplovisioning target cluster")
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+		intervals := e2eConfig.GetIntervals(specName, "wait-deprovision-cluster")
+		// Capi DeleteClusterAndWait() has a check to make sure all resources managed by CAPI
+		// is gone after Cluster deletion. This check will fails in CAPM3 as all BMH-related
```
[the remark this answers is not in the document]
lentzi90: Maybe I can help clarify a bit. We label the BMO CRDs for pivot, so that CAPI will take care of moving the objects. Because of this, the e2e framework will also expect them to disappear when deleting the cluster since they are part of the owner chain. So the point of the comment is that we cannot rely on this "delete cluster and wait" method. Instead we just delete the cluster and then separately wait for the relevant resources.

test/e2e/integration_test.go
```diff
@@ -0,0 +1,71 @@
...
+		framework.DeleteCluster(ctx, framework.DeleteClusterInput{
+			Deleter: bootstrapClient,
+			Cluster: result.Cluster,
+		})
+		Logf("Waiting for the Cluster object to be deleted")
+		framework.WaitForClusterDeleted(ctx, framework.WaitForClusterDeletedInput{
+			Getter: bootstrapClient,
+			Cluster: result.Cluster,
+		}, intervals...)
+		numberOfAvailableBMHs := 2 * numberOfWorkers
```
kashifest, line 61: why are we hardcoding number 2 ? Shouldn't we use NUM_NODES var here?

kashifest: or just use numberOfWorkers + numberofControlplanes ?

lentzi90 (commented):
/lgtm

kashifest (approved):

PR #799, opened 2022-12-06
🌱 Ginkgo v2: Use --show-node-events parameter instead of --progress

furkatgofurov7:
`--progress ` was deprecated in Ginkgo v2 (https://github.com/onsi/ginkgo/blob/master/CHANGELOG.md#ginkgo-output-now-includes-a-timeline-view-of-the-spec) and we need to move to `--show-node-events` parameter, see the deprecation warning below:
 ```
Test Suite Failed
  --progress is deprecated .  The functionality provided by --progress was confusing and is no longer needed.  Use --show-node-events instead to see node entry and exit events included in the timeline of failed and verbose specs.  Or you can run with -vv to always see all node events.  Lastly, --poll-progress-after and the PollProgressAfter decorator now provide a better mechanism for debugging specs that tend to get stuck. 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ e2e-tests: e2e-substitutions cluster-templates ## This target should be called f
 		$(CONTAINER_RUNTIME) pull $$image; \
 	done
 	time go test -v -timeout 24h --tags=e2e ./test/e2e/... -args \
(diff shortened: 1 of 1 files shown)


kashifest (approved):

mboukhalfa:
/lgtm
I think as TODO `-timeout=24h` and `--ginkgo.timeout=6h` are confusing I think we can remove the `time` command and use `ginkgo` binary to run e2e

PR #802, opened 2022-12-09
🌱 Uplift BMO and IPAM in go modules and config/

kashifest:


go.mod
```diff
@@ -87,7 +87,7 @@ require (
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattn/go-isatty v0.0.16 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
-	github.com/metal3-io/baremetal-operator/pkg/hardwareutils v0.1.1 // indirect
+	github.com/metal3-io/baremetal-operator/pkg/hardwareutils v0.1.2 // indirect
```
furkatgofurov7, line 90: this is not latest, I am seeing the same locally

kashifest: Good Catch! fixed

furkatgofurov7: and it should be coming from https://github.com/metal3-io/baremetal-operator/blob/v0.2.0/apis/go.mod#L7?

kashifest: Yes you are right :(

lentzi90 (commented):
/lgtm

go.mod
kashifest: but as I have uplifted here, it should still be fine, BMO issue has to be solved separately.

furkatgofurov7 (approved):

lentzi90 (commented):
/lgtm

PR #194, opened 2021-04-16
🌱Add test with CAPI E2E testing framework

Insullone:
Add test in HA cluster with CAPI E2E testing framework

Fixes #191

PR #805, opened 2022-12-09
🐛 Use gingko cli and timeout

mboukhalfa:
Remove time command when running e2e and ginkgo cli instead of go test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ KUSTOMIZE := $(TOOLS_BIN_DIR)/kustomize
 ENVSUBST_BIN := envsubst
 ENVSUBST := $(TOOLS_BIN_DIR)/$(ENVSUBST_BIN)-drone
 SETUP_ENVTEST = $(TOOLS_BIN_DIR)/setup-envtest
+GINKGO := "$(ROOT_DIR)/$(TOOLS_BIN_DIR)/ginkgo"
 
 ENVTEST_K8S_VERSION := 1.25.x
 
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

Makefile
```diff
@@ -235,6 +239,10 @@ $(SETUP_ENVTEST): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go get sigs.k8s.io/controller-runtime/tools/setup-envtest; \
 	go build -tags=tools -o $(BIN_DIR)/setup-envtest sigs.k8s.io/controller-runtime/tools/setup-envtest
 
+$(GINKGO): $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR); go get github.com/onsi/ginkgo/v2/ginkgo@v2.6.0
```
tuminoid, line 243: `cd; cmd` pattern should always be written as `cd && cmd`, so if the cd fails it doesn't continue execution in wrong path.

Makefile
```diff
@@ -162,27 +163,30 @@ cluster-templates: $(KUSTOMIZE) ## Generate cluster templates
...
-	cd test; \
-	time go test -v -timeout 24h --tags=e2e ./e2e/... -args \
-		--ginkgo.timeout=6h --ginkgo.v --ginkgo.trace --ginkgo.show-node-events --ginkgo.no-color=$(GINKGO_NOCOLOR) \
-		--ginkgo.junit-report="junit.e2e_suite.1.xml" \
-		--ginkgo.focus="$(GINKGO_FOCUS)" $(_SKIP_ARGS) \
+
+	$(GINKGO) --timeout=$(GINKGO_TIMEOUT) -v --trace --tags=e2e  \
+		--show-node-events --no-color=$(GINKGO_NOCOLOR) \
+		--junit-report="junit.e2e_suite.1.xml" \
+		--focus="$(GINKGO_FOCUS)" $(_SKIP_ARGS) $(ROOT_DIR)/$(TEST_DIR)/e2e/ -- \
```
tuminoid, line 183: ROOT_DIR may contain spaces so this breaks - needs quotes

tuminoid (changes requested):

Makefile
tuminoid: Same with $GINKGO as its using $ROOT_DIR earlier

mboukhalfa: Thanks fixed

Makefile
mboukhalfa: fixed

mquhuy:
I noticed that ginkgo cli was used in the past, then got changed to `go test`. Could you clarify what caused that, and why do we change it back to ginkgo cli now?

mboukhalfa:
> I noticed that ginkgo cli was used in the past, then got changed to `go test`. Could you clarify what caused that, and why do we change it back to ginkgo cli now?

Hmm I was not aware that `ginkgo` was used before and it seems changed to `go test` just to avoid installing `ginkgo` though my motivation to use g`inkgo cli` here is to avoid changing directory to run e2e tests and clean the arguments i.e avoid using `--ginkgo` prefix before each arg also I noted on ginkgo docs that ginkgo cli must be needed for certain feature like parallel which still not needed in our case

mquhuy:
> > I noticed that ginkgo cli was used in the past, then got changed to `go test`. Could you clarify what caused that, and why do we change it back to ginkgo cli now?
> 
> Hmm I was not aware that `ginkgo` was used before and it seems changed to `go test` just to avoid installing `ginkgo` though my motivation to use `ginkgo cli` here is to avoid changing directory to run e2e tests and clean the arguments i.e avoid using `--ginkgo` prefix before each arg also I noted on ginkgo docs that ginkgo cli must be needed for certain feature like parallel which still not needed in our case

Thanks. I think that's reasonable, but cannot make sure since it was purposefully changed to `go test` before. Here is the PR link: https://github.com/metal3-io/cluster-api-provider-metal3/pull/194

tuminoid (approved):

mboukhalfa:
/assign @kashifest 
cc @furkatgofurov7

PR #806, opened 2022-12-13
Change bmo deployment in e2e to using getopts

mquhuy:

[BMO is no longer using positional arguments](https://github.com/metal3-io/baremetal-operator/blob/main/tools/deploy.sh#L58), hence we need to change this in e2e


mboukhalfa:
changes already merged on main 
/close

PR #807, opened 2022-12-13
🌱 shellcheck: pin to 0.9.0

tuminoid:
Pin shellcheck to 0.9.0 with digest.

Shellcheck 0.9.0 was released this morning. This way it is in line with other repositories, and pinned by digest.

### DIFF
--- a/hack/shellcheck.sh
+++ b/hack/shellcheck.sh
@@ -14,6 +14,6 @@ else
     --volume "${PWD}:/capm3:ro,z" \
     --entrypoint sh \
     --workdir /capm3 \
-    docker.io/koalaman/shellcheck-alpine:stable \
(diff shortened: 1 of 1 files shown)


kashifest (approved):

furkatgofurov7 (commented):
Infra is still behind it, are you planning to bump that too? https://github.com/metal3-io/project-infra/blob/main/prow/config/config.yaml#L538-L551

/hold to clarify
/lgtm

kashifest:
> Infra is still behind it, are you planning to bump that too? https://github.com/metal3-io/project-infra/blob/main/prow/config/config.yaml#L538-L551
> 
> /hold to clarify /lgtm

Do you mean this @furkatgofurov7 https://github.com/metal3-io/project-infra/pull/456/

furkatgofurov7:
Yes, and I think ideally merge order should be the other way around (merge infra PR, come back here and merge with extra needed changes in case needed)

kashifest:
> Yes, and I think ideally merge order should be the other way around (merge infra PR, come back here and merge with extra needed changes in case needed)

True good point

PR #811, opened 2022-12-19
🌱 Bump upgraded k8s version and add basic logging after e2e tests

lentzi90:

Due to changes in kubeadm and cluster-api for the container registry, we need to bump the upgraded k8s version to pass the test.

It is currently hard to tell what state the cluster was in when it failed. This should hopefully make it a bit clearer.
Similar to https://github.com/metal3-io/cluster-api-provider-metal3/pull/714 (but this was hard to backport it seems).

Fixes #

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -9,8 +9,10 @@ import (
 	"path/filepath"
 
 	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	infrav1 "github.com/metal3-io/cluster-api-provider-metal3/api/v1beta1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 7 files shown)


test/e2e/common.go
```diff
@@ -175,3 +177,99 @@ func deploymentRolledOut(ctx context.Context, clientSet *kubernetes.Clientset, n
...
+		Logf("%s	%s	%s	%s", machine.GetName(), machine.Status.GetTypedPhase(), providerID, *machine.Spec.Version)
+	}
+	Logf("---------------------------------------------------------------------------------")
+	Logf("%d Machines in total", len(machines.Items))
+	Logf("=================================================================================")
+}
+
+// listNodes logs the names, status and Kubernetes version of all Nodes.
+// Similar to kubectl get nodes.
+func listNodes(ctx context.Context, c client.Client) {
```
mquhuy, line 255: Any reasons this doesn't include opts?

lentzi90: Just no need for now. All the other resources we list are namespaced, so we need to at least say what namespace to look in. Nodes are cluster wide so no need. We could at some point add filters for control plane and workers but for now we don't use it for anything like that

kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #827, opened 2023-01-10
🌱 E2e: Lint and build the e2e package

lentzi90:

This makes the linters, fmt, vet, gosec, etc. run for the e2e code and adds a make target for building the package also.
When running `make test` it will now also check that the e2e code can be built. This avoids issues with non-test files depending on `_test.go` files.


test/e2e/live_iso_test.go
```diff
@@ -91,6 +91,7 @@ func liveIsoTest() {
 		Expect(err).NotTo(HaveOccurred(), "Error getting all BMHs")
 
 		for _, bmh := range bmhs {
+			bmh := bmh
```
mboukhalfa, line 94: Is this calming the lint

test/e2e/integration_test.go
```diff
@@ -43,19 +43,19 @@ var _ = Describe("When testing integration [integration]", func() {
...
+		// DeleteClusterAndWait(), but only delete the cluster and then wait
 		// for it to be deleted.
 		framework.DeleteCluster(ctx, framework.DeleteClusterInput{
 			Deleter: bootstrapClient,
 			Cluster: result.Cluster,
 		})
 		Logf("Waiting for the Cluster object to be deleted")
 		framework.WaitForClusterDeleted(ctx, framework.WaitForClusterDeletedInput{
-			Getter: bootstrapClient,
+			Getter:  bootstrapClient,
```
mboukhalfa, line 58: extra space

mboukhalfa (commented):
LGTM

test/e2e/live_iso_test.go
lentzi90: Yes, it is for gosec G601: "Implicit memory aliasing of items from a range statement".
Ref: https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop

test/e2e/integration_test.go
lentzi90: This is from `go fmt`, it aligns with the line below so it is easier to read. Go fmt would not accept this with one less space because then it would not follow the go fmt style.

mboukhalfa: Weird! Next line seems only one space !

mboukhalfa: Oh I see format the values on the same line

test/e2e/live_iso_test.go
mboukhalfa:
```suggestion
			bmh := bmh # prevent gosec G601
```

lentzi90: Done

hack/gofmt.sh
```diff
@@ -16,7 +16,8 @@ if [ "${IS_CONTAINER}" != "false" ]; then
   # File length to be checked should be 2 since the following headers should be present
```
smoshiur1237, line 16: nit: I think it should be 3
```suggestion
  # File length to be checked should be 3 since the following headers should be present
```

smoshiur1237 (changes requested):
I leave two comments, one of them has nit.

hack/gofmt.sh
lentzi90: Thanks, fixed

test/e2e/e2e_suite_test.go
```diff
@@ -290,6 +290,6 @@ func updateCalico(config *clusterctl.E2EConfig, calicoYaml, calicoInterface stri
 
 	yamlOut, err := printYaml(yamlDocuments)
 	Expect(err).To(BeNil())
-	err = os.WriteFile(calicoYaml, yamlOut, 0664)
+	err = os.WriteFile(calicoYaml, yamlOut, 0600)
```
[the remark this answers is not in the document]
lentzi90: Gosec was complaining about more permissions than "default". I tried changing it to see if something breaks. So far it looks like it is ok like this :slightly_smiling_face:

lentzi90:
/hold
Let's discuss the `build` target and how/if we want to use it.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -86,6 +86,7 @@ func preWaitForCluster(clusterProxy framework.ClusterProxy, clusterNamespace str
 	// Split the <fileName> file into <resourceName>.yaml files based on the name of each resource and return list of the filename created
 	// so we can apply limited number on BMH in each namespace.
 	splitYAMLFile := func(fileName string, filePath string) []string {
+		//#nosec G204 -- This is just a test. We are not concerned with command injection
```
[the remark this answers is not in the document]
lentzi90: I didn't see any reasonable way to get rid of this easily right now unfortunately. However, I think we could definitely rework it later so that we do all of it in go. But I would prefer to first get this merged so we avoid adding more issues in the mean time.

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -235,6 +236,7 @@ func preCleanupManagementCluster(clusterProxy framework.ClusterProxy) {
 			By("Install Ironic in the source cluster as containers")
 			bmoPath := e2eConfig.GetVariable("BMOPATH")
 			ironicCommand := bmoPath + "/tools/run_local_ironic.sh"
+			//#nosec G204 -- This is just a test. We are not concerned with command injection
```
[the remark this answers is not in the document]
lentzi90: Same here, even though this one is a bit "simpler". We could just hard code the BMOPATH for now but since we have a variable for it I would prefer not to. I think we can rework it in a better way but it will require larger changes that I would prefer to do as a separate PR.

test/e2e/upgrade_management_cluster_test.go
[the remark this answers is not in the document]
lentzi90: I did update the comments though to reflect that we do care about this even though it is a test.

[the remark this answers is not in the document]
tuminoid: Thanks for checking. Please create a ticket to follow-up so its not forgotten.

test/e2e/upgrade_management_cluster_test.go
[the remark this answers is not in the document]
tuminoid: Its not a new bug, so just create a follow-up ticket to get it resolved later.

tuminoid (approved):

furkatgofurov7 (commented):
Let's run related e2e jobs in this PR to make sure nothing is left out and working as expected, since this is touching most of the files inside the e2e tests folder.

furkatgofurov7 (approved):
/approve

PR #836, opened 2023-01-23
🌱 Add m3d waiting condition to m3m object

mboukhalfa:
Co-authored-by: Adil Ghaffar <muhammad.adil.ghaffar@est.tech>

When there is an issue on ip allocation on ipam `m3m` will be waiting without explicitly showing what the user should look for. This PR adds a waiting condition to `m3m` when `m3d` is not ready and the latter would show in the `ErrorMessage` any error in the ip allocation.

### DIFF
--- a/api/v1beta1/condition_consts.go
+++ b/api/v1beta1/condition_consts.go
@@ -51,4 +51,9 @@ const (
 	MissingBMHReason = "MissingBMH"
 	// Could not set the ProviderID on the target cluster's Node object.
 	SettingProviderIDOnNodeFailedReason = "SettingProviderIDOnNodeFailed"
+	// Metal3DataReadyCondition reports a summary of Metal3Data status.
+	Metal3DataReadyCondition clusterv1.ConditionType = "Metal3DataReady"
+	// WaitingForMetal3DataReason used when waiting for Metal3Data
+	// to be ready before proceeding.
+	WaitingForMetal3DataReason = "WaitingForMetal3Data"
 )
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -595,7 +595,7 @@ func (m *DataManager) addressFromM3Claim(ctx context.Context, poolRef corev1.Typ
 	}
 
 	if ipClaim.Status.ErrorMessage != nil {
-		m.Data.Status.ErrorMessage = pointer.StringPtr(fmt.Sprintf(
+		m.setError(ctx, fmt.Sprintf(
 			"IP Allocation for %v failed : %v", poolRef.Name, *ipClaim.Status.ErrorMessage,
 		))
 		return addressFromPool{}, false, errors.New(*m.Data.Status.ErrorMessage)
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -3463,15 +3463,17 @@ var _ = Describe("Metal3Machine manager", func() {
 	)
 
 	type testCaseM3MetaData struct {
-		M3Machine          *infrav1.Metal3Machine
-		Machine            *clusterv1.Machine
-		DataClaim          *infrav1.Metal3DataClaim
-		Data               *infrav1.Metal3Data
-		ExpectError        bool
-		ExpectRequeue      bool
-		ExpectDataStatus   bool
-		ExpectSecretStatus bool
-		expectClaim        bool
+		M3Machine                            *infrav1.Metal3Machine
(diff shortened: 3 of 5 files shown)


baremetal/metal3data_manager.go
```diff
@@ -595,7 +595,7 @@ func (m *DataManager) addressFromM3Claim(ctx context.Context, poolRef corev1.Typ
 	}
 
 	if ipClaim.Status.ErrorMessage != nil {
-		m.Data.Status.ErrorMessage = pointer.StringPtr(fmt.Sprintf(
+		m.setError(ctx, fmt.Sprintf(
```
lentzi90, line 598: Will it recover from this? Just checking since we know that if we bubble up an error all the way to the Machine, it will get stuck.

mboukhalfa: m3m is  not failing when an error set here in m3d

lentzi90 (commented):
/lgtm

api/v1beta1/condition_consts.go
```diff
@@ -51,4 +51,9 @@ const (
 	MissingBMHReason = "MissingBMH"
 	// Could not set the ProviderID on the target cluster's Node object.
 	SettingProviderIDOnNodeFailedReason = "SettingProviderIDOnNodeFailed"
+	// Metal3DataReadyCondition reports a summary of m3d status.
+	Metal3DataReadyCondition clusterv1.ConditionType = "Metal3DataReady"
+	// WaitingForMetal3DataReason used when waiting for metal3data
```
furkatgofurov7, line 56: can we please use consistent naming for the object in question: m3d vs metal3data

baremetal/metal3data_manager.go
furkatgofurov7: is there a reason we have to do it this way? in practice it is the same thing at the end: https://github.com/metal3-io/cluster-api-provider-metal3/blob/47af8777f5e218c0785b6d3e738522fd21188cca/baremetal/metal3data_manager.go#L105-L106

api/v1beta1/condition_consts.go
mboukhalfa: fixed

baremetal/metal3data_manager.go
mboukhalfa: The function was already declared so I just decided to use it at the end it is the same thing

Sunnatillo (commented):
LGTM

lentzi90 (commented):
/lgtm

adilGhaffarDev:
It would be nice if we also change the unit test to check these new conditions:
https://github.com/metal3-io/cluster-api-provider-metal3/blob/57fa30e8aea2ef9da06f373e11e57916fd77b9d3/baremetal/metal3machine_manager_test.go#L3595

Sunnatillo (commented):
LGTM

baremetal/metal3machine_manager_test.go
```diff
@@ -3626,6 +3627,12 @@ var _ = Describe("Metal3Machine manager", func() {
 			} else {
 				Expect(tc.M3Machine.Status.RenderedData).To(BeNil())
 			}
+			metal3DataReadyCondition := filterCondition(tc.M3Machine.Status.Conditions, infrav1.Metal3DataReadyCondition)
+			if tc.ExpectMetal3DataReadyCondition {
+				Expect(metal3DataReadyCondition).To(HaveLen(1))
+			} else {
+				Expect(metal3DataReadyCondition).To(HaveLen(0))
+			}
```
mboukhalfa, line 3635: instead of checking if it exists or no let's check if it is true or false

adilGhaffarDev: fixed it kindly check.

lentzi90 (commented):
/lgtm

controllers/metal3machine_controller.go
```diff
@@ -185,6 +185,7 @@ func patchMetal3Machine(ctx context.Context, patchHelper *patch.Helper, metal3Ma
 	conditions.SetSummary(metal3Machine,
 		conditions.WithConditions(
 			infrav1.AssociateBMHCondition,
+			infrav1.Metal3DataReadyCondition,
```
furkatgofurov7, line 188: setting the condition here, but actual patching in conditions seems to be happening later in line 194-202

furkatgofurov7 (approved):
/approve

tuminoid (approved):
/lgtm

lentzi90 (commented):
/lgtm

PR #837, opened 2023-01-23
:bug: Clean varenv after mngmt upgrade

mboukhalfa:
Fix upgrade test issue:
Unset custom var env after management upgrade
Apply bmh to metal3 namespace after removing management ns

### DIFF
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -129,6 +129,7 @@ variables:
   UPGRADED_BMO_IMAGE_TAG: "${UPGRADED_BMO_IMAGE_TAG:-main}"
 
   INIT_WITH_BINARY: "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CAPI_FROM_RELEASE}/clusterctl-{OS}-{ARCH}"
+  PROVIDER_ID_FORMAT: "metal3://{{ ds.meta_data.providerid }}"
   # Pin Calico version
   CALICO_PATCH_RELEASE: "v3.24.1"
   # Pin CertManager for upgrade tests
--- a/test/e2e/upgrade_management_cluster_test.go
+++ b/test/e2e/upgrade_management_cluster_test.go
@@ -266,4 +266,24 @@ func preCleanupManagementCluster(clusterProxy framework.ClusterProxy) {
 		}
 	})
 	reInstallIronic()
+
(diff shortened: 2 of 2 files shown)


Sunnatillo (commented):
LGTM

smoshiur1237 (commented):
/lgtm

test/e2e/config/e2e_conf.yaml
```diff
@@ -129,6 +129,7 @@ variables:
   UPGRADED_BMO_IMAGE_TAG: "${UPGRADED_BMO_IMAGE_TAG:-main}"
 
   INIT_WITH_BINARY: "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CAPI_FROM_RELEASE}/clusterctl-{OS}-{ARCH}"
+  PROVIDER_ID_FORMAT: "metal3://{{ ds.meta_data.providerid }}"
```
kashifest, line 132: Can you explain a bit more about this addition?

test/e2e/upgrade_management_cluster_test.go
```diff
@@ -266,4 +266,24 @@ func preCleanupManagementCluster(clusterProxy framework.ClusterProxy) {
 		}
 	})
 	reInstallIronic()
+
+	//#nosec G204 -- We need to pass in the file name here.
+	cmd := exec.Command("bash", "-c", "kubectl apply -f bmhosts_crs.yaml  -n metal3")
+	cmd.Dir = workDir
+	output, err := cmd.CombinedOutput()
+	Logf("Applying bmh to meta3 namespace : \n %v", string(output))
```
kashifest, line 274: where are we removing the namespace as mentioned in the description?

test/e2e/config/e2e_conf.yaml
mboukhalfa: management upgrade test init the cluster with v1a4 which use the uuid format for the provider I am setting the v1b1 as default format here so after cleaning the management upgrade I just unset the env var and the other tests would use this default format

test/e2e/upgrade_management_cluster_test.go
mboukhalfa: The upgrade management test will remove the namespace generated randomly it doesn't use meta3 ns this namespace should be created by dev-env and stays always there

kashifest (approved):

PR #840, opened 2023-01-26
🌱 hack: update ensure-kind.sh

tuminoid:
Take ensure-kind.sh from CAPI, that has added support for Mac (Darwin), raised minimum kind version to 0.17.0 and added support for `set -x` via `TRACE` env.

Also, fix the script not to fail if GOPATH_BIN in not in user's path before running the script, but print a warning and then use full path for the kind so we can do version check properly, and also warn user that he needs to fix their PATH.

### DIFF
--- a/hack/ensure-kind.sh
+++ b/hack/ensure-kind.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2021 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

hack/ensure-kind.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2021 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
```
furkatgofurov7, line 3: why going back in years here?

tuminoid: CAPI script has 2019 as the copyright, despite having newer changes than us. I took their script as is and then added the path check + warning on top. Which year you prefer to have there, I never understood the copyright years that are not actively maintained along with changes, so to me both are equally bad.

furkatgofurov7: No strong opinion, from my experience these scripts are merged once (with proper year) and not touched that frequently. But when touched, years can be updated to reflect the current date. 

/approve

PR #849, opened 2023-02-13
🌱 Add new condition reasons to m3machine

Sunnatillo:
This PR adds granular condition reasons to metal3machine object.

### DIFF
--- a/api/v1beta1/condition_consts.go
+++ b/api/v1beta1/condition_consts.go
@@ -44,6 +44,17 @@ const (
 	WaitingForBootstrapReadyReason = "WaitingForBootstrapReady"
 	// AssociateBMHFailedReason documents any errors while associating Metal3Machine with a BaremetalHost.
 	AssociateBMHFailedReason = "AssociateBMHFailed"
+	// WaitingForMetal3MachineOwnerRefReason is used when Metal3Machine is waiting for OwnerReference to be
+	// set before proceeding.
+	WaitingForMetal3MachineOwnerRefReason = "WaitingForM3MachineOwnerRef"
+	// Metal3MachinePausedReason is used when Metal3Machine or Cluster is paused.
+	Metal3MachinePausedReason = "Metal3MachinePaused"
+	// WaitingforMetal3ClusterReason is used when Metal3Machine is waiting for Metal3Cluster.
+	WaitingforMetal3ClusterReason = "WaitingforMetal3Cluster"
+	// PauseAnnotationRemoveFailedReason is used when failed to remove/check pause annotation on associated bmh.
+	PauseAnnotationRemoveFailedReason = "PauseAnnotationRemoveFailed"
+	// PauseAnnotationSetFailedReason is used when failed to set pause annotation on associated bmh.
+	PauseAnnotationSetFailedReason = "PauseAnnotationSetFailedReason"
 
 	// KubernetesNodeReadyCondition documents the transition of a Metal3Machine into a Kubernetes Node.
 	KubernetesNodeReadyCondition clusterv1.ConditionType = "KubernetesNodeReady"
@@ -56,4 +67,13 @@ const (
 	// WaitingForMetal3DataReason used when waiting for Metal3Data
 	// to be ready before proceeding.
 	WaitingForMetal3DataReason = "WaitingForMetal3Data"
+	// AssociateM3MetaDataFailedReason is used when failed to associate Metadata to Metal3Machine.
+	AssociateM3MetaDataFailedReason = "AssociateM3MetaDataFailed"
+	// DisassociateM3MetaDataFailedReason is used when failed to remove OwnerReference of Meta3DataTemplate.
+	DisassociateM3MetaDataFailedReason = "DisassociateM3MetaDataFailed"
+	// DeletingReason (Severity=Info) documents a condition not in Status=True because the underlying object it is currently being deleted.
+	DeletingReason = "Deleting"
+	// DeletionFailedReason (Severity=Warning) documents a condition not in Status=True because the underlying object
+	// encountered problems during deletion. This is a warning because the reconciler will retry deletion.
+	DeletionFailedReason = "DeletionFailed"
 )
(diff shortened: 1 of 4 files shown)


furkatgofurov7:
cc @kashifest

lentzi90 (commented):
Nice job with the tests!
One nit below.

api/v1beta1/condition_consts.go
```diff
@@ -44,11 +44,33 @@ const (
...
+	// PauseAnnotationSetFailedReason is used when failed to set pause annotation on associated bmh.
+	PauseAnnotationSetFailedReason = "PauseAnnotationSetFailedReason"
 
 	// KubernetesNodeReadyCondition documents the transition of a Metal3Machine into a Kubernetes Node.
 	KubernetesNodeReadyCondition clusterv1.ConditionType = "KubernetesNodeReady"
 	// Could not find the BMH associated with the Metal3Machine.
 	MissingBMHReason = "MissingBMH"
 	// Could not set the ProviderID on the target cluster's Node object.
 	SettingProviderIDOnNodeFailedReason = "SettingProviderIDOnNodeFailed"
+	// WaitingforM3DataReason used when Metal3Machine is waiting to Metal3Data to be available.
```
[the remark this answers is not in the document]
Sunnatillo: Done

lentzi90 (commented):
/lgtm

Sunnatillo:
/hold 
Until we discuss details with @mboukhalfa

baremetal/metal3machine_manager.go
```diff
@@ -1633,6 +1633,8 @@ func (m *MachineManager) WaitForM3Metadata(ctx context.Context) error {
 
 	// If it is not ready yet, wait.
 	if !metal3Data.Status.Ready {
+		m.Log.Info("Waiting for Metal3Data to be ready")
+		m.SetConditionMetal3MachineToFalse(infrav1.KubernetesNodeReadyCondition, infrav1.WaitingforMetal3DataReason, clusterv1.ConditionSeverityInfo, "")
```
mboukhalfa, line 1637: I it would make sense here if we add new condition like  https://github.com/metal3-io/cluster-api-provider-metal3/pull/836/files#diff-cd44a3d05ab777f57cbb6377e9ad6aa4e783aace2741f72c89b76aa4d540daccR55

Sunnatillo: This part is handled in your PR

Sunnatillo:
> /hold Until we discuss m3data waiting reason with @mboukhalfa

we decided to add one more Metal3DataReady condition. This part handled in this [PR](https://github.com/metal3-io/cluster-api-provider-metal3/pull/836)

Sunnatillo:
@kashifest, @furkatgofurov7 PTAL

furkatgofurov7 (commented):
/retitle 🌱 Add new condition reasons to m3machine

controllers/metal3machine_controller.go
```diff
@@ -155,20 +157,23 @@ func (r *Metal3MachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 			machineLog.Info("failed to set pause annotation on associated bmh")
+			conditions.MarkFalse(capm3Machine, infrav1.AssociateBMHCondition, infrav1.PauseAnnotationSetFailedReason, clusterv1.ConditionSeverityInfo, "")
 			return ctrl.Result{}, nil
 		}
 	}
 
 	// Return early if the M3Machine or Cluster is paused.
 	if annotations.IsPaused(cluster, capm3Machine) {
 		machineLog.Info("reconciliation is paused for this object")
+		conditions.MarkFalse(capm3Machine, infrav1.AssociateBMHCondition, infrav1.Metal3MachinePausedReason, clusterv1.ConditionSeverityInfo, "")
```
furkatgofurov7, line 176: should not the condition marked as True after this?

controllers/metal3machine_controller.go
```diff
@@ -244,6 +249,7 @@ func (r *Metal3MachineReconciler) reconcileNormal(ctx context.Context,
 	// Make sure that the metadata is ready if any
 	err := machineMgr.AssociateM3Metadata(ctx)
 	if err != nil {
+		machineMgr.SetConditionMetal3MachineToFalse(infrav1.KubernetesNodeReadyCondition, infrav1.AssociateM3MetaDataFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
```
furkatgofurov7, line 252: same here as above. I always thought for some reason the conditions has to be marked as true

furkatgofurov7 (commented):
Mainly questions around marking the conditions as "True":

controllers/metal3machine_controller.go
Sunnatillo: I did not quite get what you are asking. 
If annotations Paused, then AssociateBMHCondition set to false, and the reason of the Condition set to Metal3MachinePausedReason.

controllers/metal3machine_controller.go
Sunnatillo: KubernetesNodeReadyCondition as name says it should be ready, when it is false it will show a reason why it is false.

controllers/metal3machine_controller.go
Sunnatillo: We are setting condition true [here](https://github.com/Nordix/cluster-api-provider-metal3/blob/704b5f3aef3f380f625a9a4d496249a8a2db562b/baremetal/metal3machine_manager.go#L1430) inside SetProviderID function

furkatgofurov7: it is set for `KubernetesNodeReadyCondition` but above is `AssociateBMHCondition`

furkatgofurov7 (approved):
This looks good, discussed the concerns offline

/approve

controllers/metal3machine_controller.go
furkatgofurov7: Currently there are two conditions exist in the provider:

1. `AssociateBMHCondition`
2. `KubernetesNodeReadyCondition`

First one (AssociateBMHCondition) deals with all the things concerned with BMH<->M3M association, and the latter (KubernetesNodeReadyCondition) deals with M3M <-> Machine (k8s node) part. 

This PR just sets the appropriate reasons based on different situations and according to the above logic, which object  we dealing  with at specific point. 

Concerning the conditions marked as True for the `AssociateBMHCondition`: 

1. reconciliation starts with `Reconcile` loop in the controller with condition marked as False but at the end it calls `ReconcileNormal`, where  we mark `AssociateBMHCondition` as True. 

2. `KubernetesNodeReady `: marked as True inside the SetProviderID call, which is the last one to be called after `KubernetesNodeReady` marked as False.

PR #857, opened 2023-02-15
:bug: hack: gomod needs to verify all go.mod files

tuminoid:
gomod.sh has only checked top-level go.mod/go.sum so far, which means api/ and hack/tools go.mods have not been verified.

The script also did not work locally at all, since /workdir was mounted read-only and "go mod tidy" writes to the directory.

Utilize existing "make modules" target.

PR #862, opened 2023-02-16
hack: gomod needs to verify all go.mod files

tuminoid:
gomod.sh has only checked top-level go.mod/go.sum so far, which means api/ and hack/tools go.mods have not been verified.

The script also did not work locally at all, since /workdir was mounted read-only and "go mod tidy" writes to the directory.

Utilize existing "make modules" target.

Uplift the Go used from insecure Go 1.17 to Go 1.18.

Manual backport of #857

### DIFF
--- a/hack/gomod.sh
+++ b/hack/gomod.sh
@@ -5,44 +5,40 @@
 #
 # NOTE: This won't work unless the build environment has internet access
 
-set -ux
+set -eux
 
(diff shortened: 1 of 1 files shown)


kashifest (approved):

lentzi90 (commented):
/lgtm

PR #868, opened 2023-02-20
🐛 Remove bmhID arg from SetNodeProviderID call

### LINKED ISSUE #590: get BMH UID from metal3Machine resource in SetNodeProviderID
The bmhID passed by the caller to SetNodeProviderID is not reliable and should be used for two reasons.
It is not new information, It was extracted from the providerID (also sent by the caller).

The providerID is retrieved from a metal3Machine resource
https://github.com/metal3-io/cluster-api-provider-metal3/blob/1464a1693312b669e16aa78c13ab05190aa9c984/baremetal/metal3machine_manager.go#L1320

The source of truth for the bmhID should be the current metal3Machine (m). 

To Do:
create a function that gets a list of bmh resources and filters based on the owner reference. 
It should return a single bmh resource. Then assign the ID to bmhID

Note:
Do not change the signature of SetNodeProviderID as doing so would required changing a lot of tests.
That can be done in test improvements in other PRs.

adilGhaffarDev:
Fixes #590

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -93,7 +93,7 @@ type MachineManagerInterface interface {
 	Update(context.Context) error
 	HasAnnotation() bool
 	GetProviderIDAndBMHID() (string, *string)
-	SetNodeProviderID(context.Context, *string, *string, ClientGetter) error
+	SetNodeProviderID(context.Context, *string, ClientGetter) error
 	SetProviderID(string)
 	SetPauseAnnotation(context.Context) error
 	RemovePauseAnnotation(context.Context) error
@@ -1332,8 +1332,7 @@ func (m *MachineManager) GetProviderIDAndBMHID() (string, *string) {
 type ClientGetter func(ctx context.Context, c client.Client, cluster *clusterv1.Cluster) (clientcorev1.CoreV1Interface, error)
 
 // SetNodeProviderID sets the metal3 provider ID on the kubernetes node.
-func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID *string, providerIDOnM3M *string, clientFactory ClientGetter) error {
-	// todo: bmhID should not be trusted and needs to removed from the signature.
+func (m *MachineManager) SetNodeProviderID(ctx context.Context, providerIDOnM3M *string, clientFactory ClientGetter) error {
 	corev1Remote, err := clientFactory(ctx, m.client, m.Cluster)
 	if err != nil {
 		return errors.Wrap(err, "Error creating a remote client")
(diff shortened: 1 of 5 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -1371,7 +1375,7 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID *string, p
 	if matchingNodesCount == 1 {
 		return nil
 	}
-	if bmhID == nil {
+	if bmhID == "" {
```
lentzi90, line 1378: I'm not sure if we should keep this. The only reason it would happen is if the bmhID is explicitly set to `""` if I understand it correctly, and in that case we are stuck anyway. So I'm thinking it would be better to add a check in `getBmhIDFromM3Machine` to see if `host.UID == ""`, if it is, we return an error there directly. What do you think?

lentzi90 (commented):
Thanks for taking care of this!

baremetal/metal3machine_manager.go
adilGhaffarDev: Yes, I agree, I think I should also remove the `if bmhID != "" {` check on the line 1354.

lentzi90: Good point, it makes sense to simplify that also! :slightly_smiling_face:

lentzi90 (commented):
/lgtm

baremetal/metal3machine_manager.go
```diff
@@ -1345,12 +1344,15 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, bmhID *string, p
 		m.Log.Info("unable to retrieve BMH name from Metal3Machine")
 		return &RequeueAfterError{RequeueAfter: requeueAfter}
 	}
-	providerIDLegacy := "metal3://unknown"
-	nodeLabel := fmt.Sprintf("%s=unknown", ProviderLabelPrefix)
-	if bmhID != nil {
-		providerIDLegacy = fmt.Sprintf("metal3://%s", *bmhID)
-		nodeLabel = fmt.Sprintf("%s=%s", ProviderLabelPrefix, *bmhID)
+	bmhID, err := m.getBmhIDFromM3Machine(ctx)
```
furkatgofurov7, line 1347: even though the var had this naming before, let's fix it here and all other places (including logs) since it is not "ID" but "UID"

furkatgofurov7 (approved):
Thanks for working on it! Looks good, except one small nit. 
I would suggest to backport it to at least release-1.3 as well.

/retitle 🐛 Remove bmhUID arg from SetNodeProviderID call
/approve
/hold

to avoid the merge.

baremetal/metal3machine_manager.go
adilGhaffarDev: I changed it in the places related to this PR, but there are some other places too where we are using bmhID instead of bmhUID. Should I change them too or create a separate PR for that?

lentzi90 (commented):
/lgtm

baremetal/metal3machine_manager.go
furkatgofurov7: To keep the scope of this PR relevant, this should be good enough. Feel free to send a follow-up PR later.

Sunnatillo (commented):
LGTM

PR #872, opened 2023-03-01
Revert "🐛 Remove bmhID arg from SetNodeProviderID call"

furkatgofurov7:
Running the unit test on latest main branch fails with:

```

• [FAILED] [0.000 seconds]
Metal3Machine manager Test SetNodeProviderID Test SetNodeProviderID [It] Set target ProviderID, matching node
/Users/fgofurov/workspace/cluster-api-provider-metal3/baremetal/metal3machine_manager_test.go:2618

  [FAILED] Unexpected error:
      <*baremetal.RequeueAfterError | 0x14000be7a48>: {
          RequeueAfter: 30000000000,
      }
      requeue in: 30s
  occurred
  In [It] at: /Users/fgofurov/workspace/cluster-api-provider-metal3/baremetal/metal3machine_manager_test.go:2595 @ 03/01/23 09:31:45.073
------------------------------
• [FAILED] [0.000 seconds]
Metal3Machine manager Test SetNodeProviderID Test SetNodeProviderID [It] Set target ProviderID, providerID set
/Users/fgofurov/workspace/cluster-api-provider-metal3/baremetal/metal3machine_manager_test.go:2633

  [FAILED] Unexpected error:
      <*baremetal.RequeueAfterError | 0x14000402fe0>: {
          RequeueAfter: 30000000000,
      }
      requeue in: 30s
  occurred
  In [It] at: /Users/fgofurov/workspace/cluster-api-provider-metal3/baremetal/metal3machine_manager_test.go:2595 @ 03/01/23 09:31:45.074
------------------------------
• [FAILED] [0.000 seconds]
Metal3Machine manager Test SetNodeProviderID Test SetNodeProviderID with noCloudProvider set to false [It] Accept providerID when set on a node
/Users/fgofurov/workspace/cluster-api-provider-metal3/baremetal/metal3machine_manager_test.go:2702

  [FAILED] Unexpected error:
      <*baremetal.RequeueAfterError | 0x14000403d78>: {
          RequeueAfter: 30000000000,
      }
      requeue in: 30s
  occurred
  In [It] at: /Users/fgofurov/workspace/cluster-api-provider-metal3/baremetal/metal3machine_manager_test.go:2688 @ 03/01/23 09:31:45.074
------------------------------
•••••••••requeue in: 30s
```

For now this reverts metal3-io/cluster-api-provider-metal3#868 and unit tests should pass locally but we still have to investigate how it passed initially in #868

PR #873, opened 2023-03-01
:bug: Fix Makefile command chaining

kashifest:


Makefile
```diff
@@ -104,15 +104,11 @@ help:  ## Display this help
 .PHONY: unit
 unit: $(SETUP_ENVTEST) ## Run unit test
 	$(shell $(SETUP_ENVTEST) use -p env --os $(ENVTEST_OS) --arch $(ARCH) $(ENVTEST_K8S_VERSION)); \
```
tuminoid, line 106: You need to change all of the `; \` to `&& \` to chain them only when previous one succeeds. This allows make to catch error in earlier commands, otherwise it will only fail if the last command fails.

Similar '; \` chains are also present in other targets as well, please fix them as well.

tuminoid (changes requested):

furkatgofurov7:
https://github.com/metal3-io/cluster-api-provider-metal3/pull/872 is merged, can we pick it up to fix the unit tests here?

kashifest:
> #872 is merged, can we pick it up to fix the unit tests here?

Now the test should pass.

/test unit

kashifest:
> > #872 is merged, can we pick it up to fix the unit tests here?
> 
> Now the test should pass.
> 
> /test unit

or probably I have to rebase first.

tuminoid (approved):
/lgtm

PR #877, opened 2023-03-02
🐛Remove bmhID arg from SetNodeProviderID call

### LINKED ISSUE #590: get BMH UID from metal3Machine resource in SetNodeProviderID

adilGhaffarDev:
Fixes #590

### DIFF
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -488,7 +488,7 @@ var _ = Describe("Metal3Machine manager", func() {
 			},
 		}
 
-		availableHost := newBareMetalHost("availableHost", &bmov1alpha1.BareMetalHostSpec{}, bmov1alpha1.StateReady, &bmov1alpha1.BareMetalHostStatus{}, true, "metadata", false)
+		availableHost := newBareMetalHost("availableHost", &bmov1alpha1.BareMetalHostSpec{}, bmov1alpha1.StateReady, &bmov1alpha1.BareMetalHostStatus{}, true, "metadata", false, "")
 
 		hostWithConRef := bmov1alpha1.BareMetalHost{
 			ObjectMeta: metav1.ObjectMeta{
@@ -1098,7 +1098,7 @@ var _ = Describe("Metal3Machine manager", func() {
 			UserDataNamespace:         "otherns",
 			ExpectedUserDataNamespace: "otherns",
 			Host: newBareMetalHost("host2", nil, bmov1alpha1.StateNone,
-				nil, false, "metadata", false,
+				nil, false, "metadata", false, "",
 			),
 			ExpectedImage:  expectedImg(),
 			ExpectUserData: true,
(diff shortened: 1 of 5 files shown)


adilGhaffarDev:
/hold
testing if unit test issue is fixed on PR

baremetal/metal3machine_manager_test.go
```diff
@@ -4761,22 +4763,29 @@ func newBareMetalHost(name string,
...
 	if name == "" {
 		return &bmov1alpha1.BareMetalHost{
 			ObjectMeta: testObjectMeta("", namespaceName, ""),
 		}
 	}
 
+	BMHUid := types.UID(bmhuid)
+	if len(bmhUID) > 0 {
+		BMHUid = types.UID(bmhUID[0])
+	}
```
lentzi90, line 4777: I'm getting confused with all these BMH UIDs. :flushed: 
Why do we need `bmhuid` that is coming from somewhere outside of this file? Is it like a default if nothing is specified?
Could you add some comment to describe the expected behavior?

adilGhaffarDev: yes, it is confusing, let me fix it. In `metal3machine_manager_test.go` we were already using it from two places one `bmhuid` was coming from suite_test.go and one `Bmhuid` was in the same file, which needs to be refactored. I will change it to the same uid everywhere in `metal3machine_manager_test.go` and also fix the above function.
And yes aim of the above implementation was to set the default value if nothing is passed.

furkatgofurov7:
cc

baremetal/metal3machine_manager_test.go
adilGhaffarDev: I have done some fixes can you check if they look ok.

baremetal/metal3machine_manager_test.go
```diff
@@ -4761,22 +4778,30 @@ func newBareMetalHost(name string,
...
-	clusterlabel bool) *bmov1alpha1.BareMetalHost {
+	clusterlabel bool,
+	bmhUID string) *bmov1alpha1.BareMetalHost {
 	if name == "" {
 		return &bmov1alpha1.BareMetalHost{
 			ObjectMeta: testObjectMeta("", namespaceName, ""),
 		}
 	}
 
+	BMHUid := Bmhuid
```
lentzi90, line 4789: I think it would be better to keep this lower case to avoid confusion with exported variables. Perhaps we could change it to `uid` to keep things simple?
```suggestion
	uid := Bmhuid
```

lentzi90 (commented):
Nice cleanup!

lentzi90 (commented):
Great job!
It would be nice to get rid of the global `Bmhuid` at some point but that can be handled separately, this is already a good cleanup!

/lgtm

furkatgofurov7:
> /test-centos-e2e-main
> /test-ubuntu-e2e-main

These triggers are not relevant anymore

/test-ubuntu-e2e-feature-main
/test-centos-e2e-feature-main

furkatgofurov7 (approved):

PR #882, opened 2023-03-06
remove unsupported OSes

tuminoid:
Remove unsupported OSes. We only support and test with Centos9 and Ubuntu 20.04 and 22.04.

Main PR: https://github.com/metal3-io/metal3-dev-env/pull/1164

/hold

### DIFF
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -24,10 +24,10 @@ function os_check() {
     export DISTRO="${ID}${VERSION_ID%.*}"
     export OS="${ID}"
     export OS_VERSION_ID=$VERSION_ID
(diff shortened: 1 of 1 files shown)


tuminoid:
As discussed in community meeting March 8th 2023, Centos8 stream is EOL in May2024, so it is still maintained, and will not be removed until that time.

Ubuntu 18 is also not EOL for a month at least, so not removing that one either.

And as agreed with RedHat developers, RHEL is expected to be identical and working with the CentOS testing, and they are using it, so it will not be removed either. I will merge the other PR that will fix the support for the older OSes instead, and will improve documentation on the supported OS status.

/close

PR #884, opened 2023-03-07
✨Enable concurrency in metal3 controllers

kashifest:

Fixes #

### DIFF
--- a/main.go
+++ b/main.go
@@ -150,14 +150,14 @@ func initFlags(fs *pflag.FlagSet) {
 	logs.AddFlags(fs, logs.SkipLoggingConfigurationFlags())
 	logsv1.AddFlags(logOptions, fs)
 
-	flag.StringVar(
+	fs.StringVar(
 		&metricsBindAddr,
 		"metrics-bind-addr",
 		"localhost:8080",
 		"The address the metric endpoint binds to.",
 	)
 
-	flag.BoolVar(
+	fs.BoolVar(
 		&enableLeaderElection,
 		"leader-elect",
 		false,
@@ -192,7 +192,7 @@ func initFlags(fs *pflag.FlagSet) {
 		"Duration the LeaderElector clients should wait between tries of actions (duration string)",
 	)
 
-	flag.StringVar(
+	fs.StringVar(
 		&watchNamespace,
 		"namespace",
 		"",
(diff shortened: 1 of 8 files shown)


mboukhalfa:
testing the new trigger format

main.go
```diff
@@ -198,33 +206,54 @@ func initFlags(fs *pflag.FlagSet) {
...
 
-	flag.StringVar(
+	fs.StringVar(
 		&healthAddr,
 		"health-addr",
 		":9440",
 		"The address the health endpoint binds to.",
 	)
+
+	fs.IntVar(&metal3MachineConcurrency, "metal3Machine-concurrency", 10,
```
lentzi90, line 237: In CAPI they seem to [stick to all lower case](https://github.com/kubernetes-sigs/cluster-api/blob/838dcbbb20c45fc5966727151f9ca1320dfcd783/main.go#L163) for the flags, maybe we should do the same? (Same for the rest of the concurrency flags)
```suggestion
	fs.IntVar(&metal3MachineConcurrency, "metal3machine-concurrency", 10,
```

lentzi90 (commented):
Thanks! This looks good to me and in line with how CAPI is doing it.
Just one comment below about the flags.

main.go
kashifest: yes I thought about it and later choose to just go this way. I can change it.

lentzi90: I have no strong preference on this but I think it is quite common to keep flags lower case so maybe it is better that way

tuminoid: Please do not put upper case characters into flags, I can guarantee nobody can write correctly if you do. :)

kashifest: Done

lentzi90 (commented):
/lgtm

kashifest:
@lentzi90 @furkatgofurov7 what should be the icon in the title , ⚠️ or 🌱 ?

tuminoid:
> @lentzi90 @furkatgofurov7 what should be the icon in the title , warning or seedling ?

:sparkles: I think is correct for new features

lentzi90:
:sparkles: makes sense for added feature, but we also change the default from 1 to 10 with this. I don't think such a config change would require :warning: so I would say :sparkles:

furkatgofurov7:
I would say ✨ is an appropriate one IMO as well.

furkatgofurov7 (approved):
/retitle ✨Enable concurrency in metal3 controllers
/approve
/hold unhold if you think also new feature label is appropriate

furkatgofurov7:
As discussed offline, we will enable the concurrency in BMO, IPAM controllers as well to align all of them.

kashifest:
/hold cancel
I agree with all of you regarding the title

PR #885, opened 2023-03-08
:seedling: Add support for new dev-env network variable names in e2e tests

Rozzii:
A number of different networking related variables will be renamed in metal3-dev-env and the changes will be tested with capm3 e2e integration tests thus capm3 have to support both the old and the new variable names during the transitioning period.

Related dev-env PR https://github.com/metal3-io/metal3-dev-env/pull/1169


test/e2e/config/e2e_conf.yaml
```diff
@@ -141,15 +141,15 @@ variables:
   CONFIG_FILE_PATH: "${HOME}/.cluster-api/clusterctl.yaml"
   SERVICE_CIDR: "10.96.0.0/12"
   POD_CIDR: "192.168.0.0/18"
-  PROVISIONING_CIDR: "24"
+  BM_PROVISIONER_CIDR: "24"
   CLUSTER_APIENDPOINT_HOST: "192.168.111.249"
   CLUSTER_APIENDPOINT_PORT: "6443"
   EXTERNAL_SUBNET_V4_PREFIX: "24"
   CLUSTER_PROVISIONING_INTERFACE: "ironicendpoint"
-  BAREMETALV4_POOL_RANGE_START: "192.168.111.100"
```
kashifest, line 149: why external and not baremetal?

test/e2e/config/e2e_conf.yaml
```diff
@@ -141,15 +141,15 @@ variables:
   CONFIG_FILE_PATH: "${HOME}/.cluster-api/clusterctl.yaml"
   SERVICE_CIDR: "10.96.0.0/12"
   POD_CIDR: "192.168.0.0/18"
-  PROVISIONING_CIDR: "24"
+  BM_PROVISIONER_CIDR: "24"
```
kashifest, line 144: BM stands for?

Rozzii: bare metal

test/e2e/config/e2e_conf.yaml
Rozzii: In our existing code base, the network we refer to as BAREMETAL is in reality the external network , same interface is used for K8s API for example.

kashifest (approved):
One small nit, change BM to Baremetal

Rozzii:
/hold
I am debugging something else , let's hold until I have fixed Kashif's nit.

tuminoid (approved):

scripts/environment.sh
```diff
@@ -85,6 +85,10 @@ export PROVISIONING_URL_HOST="172.22.0.1"
 export CLUSTER_PROVISIONING_IP="172.22.0.2"
 export CLUSTER_URL_HOST="$CLUSTER_PROVISIONING_IP"
 
+export BM_RPOVISIONER_URL_HOST="172.22.0.1"
+export CLUSTER_BM_PROVISIONER_IP="172.22.0.2"
```
mboukhalfa, line 89: this two names are confusing I think both means ironic ip ? or there is a difference between `BM_PROVISIONER` and `CLUSTER_BM_PROVISIONER`

Rozzii: The situation is the same as with the original variables:
export PROVISIONING_URL_HOST="172.22.0.1"
export CLUSTER_PROVISIONING_IP="172.22.0.2"

Later in dev-env I can take look at these vars because indeed they seem to be redundant but the scope of the current PR is just to rename the variables.

PR #906, opened 2023-03-27
:bug: Fix typo in e2e environment config

Rozzii:


kashifest (approved):

scripts/environment.sh
```diff
@@ -85,9 +85,9 @@ export PROVISIONING_URL_HOST="172.22.0.1"
 export CLUSTER_PROVISIONING_IP="172.22.0.2"
 export CLUSTER_URL_HOST="$CLUSTER_PROVISIONING_IP"
 
-export BARE_METAL_RPOVISIONER_URL_HOST="172.22.0.1"
+export BARE_METAL_PROVISIONER_URL_HOST="172.22.0.1"
```
kashifest, line 88: where are we using these? Shall we not change it there as well?

lentzi90 (commented):
/lgtm

scripts/environment.sh
Rozzii: These are used in dev-env and I have a PR that implements these names.

PR #913, opened 2023-03-31
:seedling: Backport support for dev-env new network variable names in e2e tests (release-1.2)

Rozzii:
A number of different networking related variables have been renamed in metal3-dev-env and the changes will be tested with capm3 e2e integration tests thus capm3 have to support both the old and the new variable names during the transitioning period.


lentzi90 (commented):
/lgtm

I would suggest relabeling as seedling since it only has to do with test code/variables

tuminoid (commented):
/lgtm

PR #918, opened 2023-04-03
🌱 Disable scheduled workflows from running in forks

kashifest:
Workflows use resources that are only available when run in the [metal3-io/cluster-api-provider-metal3](https://github.com/metal3-io/cluster-api-provider-metal3) repository, so workflow run would always fail when they run in a forked repository. They are of no value to fork repositories. They are triggered by the schedule event, so they cause regular annoying and confusing workflow failure notifications for every fork owner. Rather checking these workflows only in upstream repo is enough. This PR is inspired by https://github.com/arduino/arduino-cli/pull/888

### DIFF
--- a/.github/workflows/kubesec.yml
+++ b/.github/workflows/kubesec.yml
@@ -10,6 +10,9 @@ permissions:
 
 jobs:
   setup:
+    # This workflow is only of value to the metal3-io/cluster-api-provider-metal3 repository and
+    # would always fail in forks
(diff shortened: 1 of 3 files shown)


tuminoid (approved):
/lgtm

This should do the trick.

smoshiur1237 (approved):
/lgtm

dtantsur (approved):

PR #922, opened 2023-04-05
 🌱 Improve target log collection for e2e

smoshiur1237:
This PR will help the log collection of target cluster during e2e tests. It has the following parts:

-  It will fetch target cluster kubeconfig and save it to tmp folder with kubeconfig-test1.yaml, which will help the project-infra log collection of target during failure in target cluster and also helps to collect logs of target cluster before re-pivoting in case of successful run.
- Fetch target clulster resources during integration test.
- Add script to collect target logs before re-pivot.

### DIFF
--- a/scripts/fetch_target_logs.sh
+++ b/scripts/fetch_target_logs.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -x
+
+DIR_NAME="/tmp/target_cluster_logs"
+NAMESPACES="$(kubectl --kubeconfig="${KUBE_CONFIG}" get namespace -o json | jq -r '.items[].metadata.name')"
+mkdir -p "${DIR_NAME}"
+for NAMESPACE in ${NAMESPACES}
+do
+  mkdir -p "${DIR_NAME}/${NAMESPACE}"
+# Fetch logs from target cluster according to the pods name
+  PODS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" -o json | jq -r '.items[].metadata.name')"
+  for POD in ${PODS}
+  do
+    mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}"
+    CONTAINERS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" "${POD}" -o json | jq -r '.spec.containers[].name')"
+    for CONTAINER in ${CONTAINERS}
+    do
+      LOG_DIR="${DIR_NAME}/${NAMESPACE}/${POD}/${CONTAINER}"
+      mkdir -p "${LOG_DIR}"
+      kubectl --kubeconfig="${KUBE_CONFIG}" logs -n "${NAMESPACE}" "${POD}" "${CONTAINER}" \
+        > "${LOG_DIR}/stdout.log" 2> "${LOG_DIR}/stderr.log"
+    done
+  done
+done
--- a/test/e2e/pivoting.go
+++ b/test/e2e/pivoting.go
@@ -83,6 +83,19 @@ func pivoting(ctx context.Context, inputGetter func() PivotingInput) {
 		fetchContainerLogs(&ironicContainers, input.ArtifactFolder, input.E2EConfig.GetVariable("CONTAINER_RUNTIME"))
 	}
 
+	By("Fetch target cluster kubeconfig for target cluster log collection")
+	kubeconfigPath := input.TargetCluster.GetKubeconfigPath()
+	os.Setenv("KUBE_CONFIG", kubeconfigPath)
+	Logf("Save kubeconfig in temp folder for project-infra target log collection")
+	// TODO(smoshiur1237): This is a workaround to copy the target kubeconfig and enable project-infra
+	// target log collection. There is possibility to handle the kubeconfig in better way.
+	// KubeconfigPathTemp will be used by project-infra target log collection only incase of failed e2e test
+	kubeconfigPathTemp := "/tmp/kubeconfig-test1.yaml"
+	cmd := exec.Command("cp", kubeconfigPath, kubeconfigPathTemp) // #nosec G204:gosec
+	stdoutStderr, er := cmd.CombinedOutput()
+	Logf("%s\n", stdoutStderr)
+	Expect(er).To(BeNil(), "Cannot fetch target cluster kubeconfig")
+
 	By("Remove Ironic containers from the source cluster")
 	isIronicDeployment := true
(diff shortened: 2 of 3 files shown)


scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -x
+
+KUBE_CONFIG="$(sudo find /tmp/ -type f -name "kubeconfig*")"
```
mboukhalfa, line 5: we should export this from golang directly instead of using `cp`

smoshiur1237: We need to keep it locally as , project-infra log collection will call log collection in-case of failure.

test/e2e/pivoting.go
```diff
@@ -83,6 +83,15 @@ func pivoting(ctx context.Context, inputGetter func() PivotingInput) {
 		fetchContainerLogs(&ironicContainers, input.ArtifactFolder, input.E2EConfig.GetVariable("CONTAINER_RUNTIME"))
 	}
 
+	By("Fetch target cluster kubeconfig and save for target log collection")
+	kubeconfigFile := "/tmp/kubeconfig-test1.yaml"
+	kubeconfig := input.TargetCluster.GetKubeconfigPath()
```
mboukhalfa, line 88: Here we can use:
`os.Setenv("KUBE_CONFIG", "kubeconfig")`
and remove  from the script
`KUBE_CONFIG="$(sudo find /tmp/ -type f -name "kubeconfig*")"`

scripts/fetch_target_logs.sh
smoshiur1237: We need to save it locally, so that both from project-infra and log collection and log collection from e2e can work with same file.

test/e2e/pivoting.go
smoshiur1237: I agree that this is an option. But the log collection from project-infra will not work as it couldn't find the kubeconfig locally.

scripts/fetch_target_logs.sh
smoshiur1237: But same kubeconfig that is saved in temp are used by project-infra script.

smoshiur1237:
Link for e2e test with target cluster logs: https://jenkins.nordix.org/job/metal3_capm3_e2e_integration_test_centos/93/

test/e2e/pivoting.go
```diff
@@ -83,6 +83,19 @@ func pivoting(ctx context.Context, inputGetter func() PivotingInput) {
 		fetchContainerLogs(&ironicContainers, input.ArtifactFolder, input.E2EConfig.GetVariable("CONTAINER_RUNTIME"))
 	}
 
+	By("Fetch target cluster kubeconfig for target cluster log collection")
+	kubeconfigPath := input.TargetCluster.GetKubeconfigPath()
+	os.Setenv("KUBE_CONFIG", kubeconfigPath)
```
lentzi90, line 88: I would suggest using the standard `KUBECONFIG` and make sure to unset it after we are done. If we cannot use the standard variable, then maybe a name that explains better which cluster it is for. For example `KUBECONFIG_WORKLOAD` or similar.

scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -x
+
+DIR_NAME="/tmp/target_cluster_logs"
+NAMESPACES="$(kubectl --kubeconfig="${KUBE_CONFIG}" get namespace -o json | jq -r '.items[].metadata.name')"
```
lentzi90, line 6: Nit. It is not necessary to use `jq` here.
```suggestion
NAMESPACES="$(kubectl --kubeconfig="${KUBE_CONFIG}" get namespace -o jsonpath='{.items[*].metadata.name}')"
```

scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,25 @@
...
+mkdir -p "${DIR_NAME}"
+for NAMESPACE in ${NAMESPACES}
+do
+  mkdir -p "${DIR_NAME}/${NAMESPACE}"
+# Fetch logs from target cluster according to the pods name
+  PODS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" -o json | jq -r '.items[].metadata.name')"
```
lentzi90, line 12:
```suggestion
  PODS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" -o jsonpath='{.items[*].metadata.name}')"
```

scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,25 @@
...
+  for POD in ${PODS}
+  do
+    mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}"
+    CONTAINERS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" "${POD}" -o json | jq -r '.spec.containers[].name')"
```
lentzi90, line 16:
```suggestion
    CONTAINERS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" "${POD}" -o jsonpath='{.spec.containers[*].name}')"
```

kashifest:
since we clone dev-env here in e2e tests, do we need to keep a copy of fetch target cluster log file here or can we leverage https://github.com/metal3-io/metal3-dev-env/blob/main/tests/scripts/fetch_target_logs.sh ? Thinking about code maintenance and entropy here.

lentzi90:
(I am fine with taking this in directly and improving in follow up PRs also. Quite important to get these logs!)

kashifest:
> (I am fine with taking this in directly and improving in follow up PRs also. Quite important to get these logs!)

Agree.
/approve

test/e2e/pivoting.go
smoshiur1237: It will be handled in separate followup PR

lentzi90:
Alright let's get the logs
/lgtm

smoshiur1237:
> since we clone dev-env here in e2e tests, do we need to keep a copy of fetch target cluster log file here or can we leverage https://github.com/metal3-io/metal3-dev-env/blob/main/tests/scripts/fetch_target_logs.sh ? Thinking about code maintenance and entropy here.

We can check the scope in future as the main goal is to get these logs in e2e. I will open a ticket and mention this scope and improvements.

smoshiur1237:
/cherry-pick release-1.2
/cherry-pick release-1.3
Adding the log collection script to release-1.2 and release 1.3

PR #934, opened 2023-04-24
🐛 Fix cleanup for keep tests

Sunnatillo:
Currently, during keep tests environment is not cleaned up after successful e2e test. The env should be kept only tests fails. This PR fixes this issue by introducing KEEP_TEST_ENV variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ E2E_CONF_FILE_ENVSUBST ?= $(E2E_OUT_DIR)/$(notdir $(E2E_CONF_FILE))
 E2E_CONTAINERS ?= quay.io/metal3-io/cluster-api-provider-metal3 quay.io/metal3-io/baremetal-operator quay.io/metal3-io/ip-address-manager
 
 SKIP_CLEANUP ?= false
+KEEP_TEST_ENV ?= false
 EPHEMERAL_TEST ?= false
 SKIP_CREATE_MGMT_CLUSTER ?= true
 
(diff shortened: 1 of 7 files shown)


Sunnatillo:
failed with downloading a package flake. 

/keep-test-centos-e2e-feature-main
/test-centos-e2e-feature-main

adilGhaffarDev:
I don't think we need to do cleanup in case of keep tests, we run keep tests for debugging and even if the test is passed, developer might want to check something in the environment.

lentzi90 (commented):
Skip cleanup is useful also when running tests manually. If we need to clean up in the CI, then why not just set `skipCleanup = false` instead of removing the functionality?

mboukhalfa (approved):
Make sense to failfast when we are skipping cleanup whether with keep tests or other tests

mboukhalfa:
/approve cancel 
we need to remove `AbortSuite` we do not need them anymore

Makefile
```diff
@@ -182,11 +183,13 @@ e2e-tests: $(GINKGO) e2e-substitutions cluster-templates ## This target should b
 
 	$(GINKGO) --timeout=$(GINKGO_TIMEOUT) -v --trace --tags=e2e  \
 		--show-node-events --no-color=$(GINKGO_NOCOLOR) \
+		--fail-fast="$(KEEP_TEST_ENV)" \
 		--junit-report="junit.e2e_suite.1.xml" \
 		--focus="$(GINKGO_FOCUS)" $(_SKIP_ARGS) "$(ROOT_DIR)/$(TEST_DIR)/e2e/" -- \
 		-e2e.artifacts-folder="$(ARTIFACTS)" \
 		-e2e.config="$(E2E_CONF_FILE_ENVSUBST)" \
 		-e2e.skip-resource-cleanup=$(SKIP_CLEANUP) \
+		-e2e.keep-test-environment=$(KEEP_TEST_ENV) \
```
lentzi90, line 192: What is the difference between `SKIP_CLEANUP` and `KEEP_TEST_ENV`? I see that this PR changes many of the `skipCleanup` to `keepTestEnv` so I'm wondering if we now should remove one or in what way they differ.

PR #936, opened 2023-04-25
🌱 Add wait for ironic/bmo deployment becomes available

mboukhalfa:
This PR adds a wait function for deployment to become available after installing ironic or bmo

### DIFF
--- a/test/e2e/pivoting.go
+++ b/test/e2e/pivoting.go
@@ -120,12 +120,12 @@ func pivoting(ctx context.Context, inputGetter func() PivotingInput) {
 
 	By("Create Ironic namespace")
 	targetClusterClientSet := input.TargetCluster.GetClientSet()
-	ironicNamespace := &corev1.Namespace{
+	ironicNamespaceObj := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: input.E2EConfig.GetVariable(ironicNamespace),
 		},
 	}
-	_, err := targetClusterClientSet.CoreV1().Namespaces().Create(ctx, ironicNamespace, metav1.CreateOptions{})
+	_, err := targetClusterClientSet.CoreV1().Namespaces().Create(ctx, ironicNamespaceObj, metav1.CreateOptions{})
 	Expect(err).To(BeNil(), "Unable to create the Ironic namespace")
 
 	By("Initialize Provider component in target cluster")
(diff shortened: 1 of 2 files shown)


dtantsur (approved):

test/e2e/config/e2e_conf.yaml
```diff
@@ -167,7 +167,7 @@ intervals:
   default/wait-vm-state: ["20m", "100ms"]
   default/monitor-vm-state: ["1m", "500ms"]
   default/monitor-provisioning: ["5m", "500ms"]
-  default/wait-deployment: ["10m", "10s"]
+  default/wait-deployment: ["20m", "10s"]
```
lentzi90, line 170: 10 minutes is really not enough? :flushed:

test/e2e/pivoting.go
```diff
@@ -296,24 +305,22 @@ func installIronicBMO(inputGetter func() installIronicBMOInput) {
...
-	Expect(er).To(BeNil(), "Cannot get the stderr from the command")
-	err = cmd.Wait()
-	if len(errorData) > 0 {
-		Logf("Error of the shell: %v\n", string(errorData))
+	stdoutStderr, er := cmd.CombinedOutput()
+	Logf("%s\n", stdoutStderr)
+	Expect(er).To(BeNil(), "Failed to deploy Ironic")
+
+	name := "-ironic"
+	if !input.deployIronic {
```
lentzi90, line 313: Makes more sense to check for BMO I think. Note that both could also be set to true (at least that is how the deploy.sh script works).
```suggestion
	if input.deployBMO {
```

test/e2e/config/e2e_conf.yaml
mboukhalfa: fixed

test/e2e/pivoting.go
mboukhalfa: done thanks now it should wait for both if the function deployed both at the same time

lentzi90 (commented):
/lgtm

test/e2e/pivoting.go
```diff
@@ -296,24 +305,27 @@ func installIronicBMO(inputGetter func() installIronicBMOInput) {
...
+	}
+	// Wait for the deployments to become available
+	clientSet := input.ManagementCluster.GetClientSet()
+	for _, name := range deploymentNameList {
+		deployment, err := clientSet.AppsV1().Deployments(input.Namespace).Get(ctx, input.NamePrefix+name, metav1.GetOptions{})
+		Expect(err).To(BeNil(), "Unable to get the deployment %s in namespace %s \n error message: %s", input.NamePrefix+name, input.Namespace, err)
+		framework.WaitForDeploymentsAvailable(ctx, framework.WaitForDeploymentsAvailableInput{
+			Getter:     input.ManagementCluster.GetClient(),
+			Deployment: deployment,
+		}, input.E2EConfig.GetIntervals(input.SpecName, "wait-deployment")...)
```
kashifest, line 327: I am thinking if this waiting part should be added in a function?  What if when we scale down and up ironic/bmo deployment and we want to check /wait until deployment is available or not ? That workflow wont necessarily install ironic and bmo.

kashifest: And this could be then generic function for any deployment/deployments you would want to wait to be available .

mboukhalfa: We are already using `framework.WaitForDeploymentsAvailable` which is defined in the framework I do think we have to wrap this function call in another function

kashifest: ow, then why are we putting deployment names in a list, then running a loop to get an individual deployment name again from that list and then calling this function. Why dont we just call the function with the deployment name directly.

mboukhalfa: I do not know why CAPI call this function `WaitForDeploymentsAvailable` although it accept only one deployment to wait for! since I am not sure if the install function is deployment only ironic or bmo or both at the same call I am doing this check and looping over one improvement I can do is to get the list of all deployment in the namespace and loop over it but I cannot be should that this name space only contains bmo and ironic so it is safer to check with the name

kashifest: Ok I get the issue is the uncertainity if the install function is called for installing bmo, ironic or both. In that case , instead of putting the name of the deployment in the list in the 
`if input.deployBMO {`  we might actually call the `framework.WaitForDeploymentsAvailable` function already for the deployment. This would be a good improvement for the future. For now I am happy to take this in so that we have proper workflow atleast.
/approve

PR #950, opened 2023-04-28
:seedling: Added pod security standards manifest file for CAPM3

as20203:
Added manifest file to enforce restricted pod security standards for capm3 repository.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -356,6 +356,11 @@ def strip_sec_ctx(yaml):
             spec["securityContext"] = {}
             for container in spec.get("containers", []):
                 container["securityContext"] = {}
+        if data.get("kind") == "Namespace":
+            spec = data["metadata"]["labels"]
+            pod_security_standard_label = "pod-security.kubernetes.io/enforce" 
+            spec.pop(pod_security_standard_label, None)
+                
         output.append(str(encode_yaml(data)))
 
     return "---\n".join(output)
--- a/config/default/capm3/namespace.yaml
+++ b/config/default/capm3/namespace.yaml
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: capm3-system
+  labels:
+    pod-security.kubernetes.io/enforce: restricted
+


tuminoid:
@kashifest what do you think where this Pod Security Standard definition should go? Should it be default or in kustomization etc

lentzi90:
I would suggest adding this to the current namespace definition here: https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/config/default/capm3/namespace.yaml
Adding the manifest in default without adding a reference to it in the `kustomization.yaml` will not result in anything since kustomize will not use it. We could do it as a separate overlay or component and make it optional, but I don't see the benefit of that. Secure by default is better, so let's add it to the already defined namespace. :slightly_smiling_face:

as20203:
Updated pod security standards as mentioned by @lentzi90

lentzi90 (commented):
/lgtm

config/default/capm3/namespace.yaml
```diff
@@ -2,3 +2,11 @@ apiVersion: v1
...
 metadata:
   name: capm3-system
+  labels:
+    kubernetes.io/metadata.name: capm3-system
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/audit-version: v1.27
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/warn: restricted
+    pod-security.kubernetes.io/warn-version: v1.27
```
zhouhao3, line 12: The last line lacks a newline character.

as20203: @zhouhao3  fixed new line comment.

lentzi90:
Please squash the commits also @as20203 , then I'll put lgtm again :slightly_smiling_face:

config/default/capm3/namespace.yaml
as20203: @zhouhao3 added a new line

tuminoid (changes requested):
Please keep holding this until we have finished discussion with CAPI community, and also have Tilt workaround included.

tuminoid:
> Please keep holding this until we have finished discussion with CAPI community, and also have Tilt workaround included.

Discussion had, no objections. Namespace is good place to have it. We should just clean up the extra warnings, versions etc, and fix the Tilt support.

config/default/capm3/namespace.yaml
```diff
@@ -2,3 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: capm3-system
+  labels:
+    kubernetes.io/metadata.name: capm3-system
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: latest
```
tuminoid, line 8: Same question as for the BMO PR, can we use the latest when we need to support much older k8s versions?

lentzi90: It should work well in most cases AFAIK. The only mention of needing a specific version that I have seen is this: https://kubernetes.io/docs/concepts/security/pod-security-standards/#pod-os-field
> Note: Kubelets prior to v1.24 don't enforce the pod OS field, and if a cluster has nodes on versions earlier than v1.24 the restricted policies should be pinned to a version prior to v1.25.

So in case the management cluster is v1.24 or older this would need to be changed. Kubernetes v1.24 is EOL in ~2 months so I think it would be fine to keep it as is.

config/default/capm3/namespace.yaml
```diff
@@ -2,3 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: capm3-system
+  labels:
+    kubernetes.io/metadata.name: capm3-system
```
tuminoid, line 6: Do we need this? Seems like duplication if the L4.

config/default/capm3/namespace.yaml
tuminoid: OK, so if the `latest` is default, we can probably remove this line too.

as20203: I didn't find anything in the documentation https://kubernetes.io/docs/concepts/security/pod-security-standards/#pod-os-field that this value is applied by default as the latest, so I'm not sure if I should remove this line. Even in the sample files given below they are providing the enforce-version.
https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/security/podsecurity-restricted.yaml

config/default/capm3/namespace.yaml
as20203: I'll remove this, I don't think it is required.

config/default/capm3/namespace.yaml
lentzi90: For some reason they don't mention it in the [page for namespace labels](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-namespace-labels/). But [here](https://kubernetes.io/docs/concepts/security/pod-security-admission/) they mention that the version labels can be used to pin the version (so implies the default would be not pinned -> latest). In the page for configuring [cluster wide pod security standards](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-admission-controller/) they explicitly mention that latest is the default.

as20203: I see, I'll update this as well since this makes much sense after going through these links.

as20203:
PR is updated with tilt workaround  and I've also updated manifest file according to our above discussion. Kindly review @lentzi90  @tuminoid @kashifest

Tiltfile
```diff
@@ -356,6 +356,12 @@ def strip_sec_ctx(yaml):
             spec["securityContext"] = {}
             for container in spec.get("containers", []):
                 container["securityContext"] = {}
+        if data.get("kind") == "Namespace":
+            spec = data["metadata"]["labels"]
+            pod_security_standard_label =  "pod-security.kubernetes.io/enforce" 
+            spec.pop(pod_security_standard_label)
```
tuminoid, line 362: This raises `KeyError` if label is not found. Add default value of `None` to avoid it.

tuminoid (changes requested):

Tiltfile
```diff
@@ -356,6 +356,11 @@ def strip_sec_ctx(yaml):
             spec["securityContext"] = {}
             for container in spec.get("containers", []):
                 container["securityContext"] = {}
+        if data.get("kind") == "Namespace":
+            spec = data["metadata"]["labels"]
+            pod_security_standard_label =  "pod-security.kubernetes.io/enforce"
```
tuminoid, line 361: Nit: extra whitespace.

as20203: fixed spacing issue.

tuminoid:
/retitle :seedling: Added pod security standards manifest file for CAPM3
/test-ubuntu-integration-main

tuminoid (commented):
I think we're good to go with this one.

/lgtm

tuminoid (approved):

kashifest:
/hold cancel
cc @kashifest @mboukhalfa

config/default/capm3/namespace.yaml
```diff
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: capm3-system
+  labels:
+    pod-security.kubernetes.io/enforce: restricted
```
furkatgofurov7, line 6: while this seems a best practice & recommended security hardening, I have not seen something similar in the CAPI/*providers repos. Is there any specific use case for having this?

furkatgofurov7 (commented):
PR title does not seem to match the made changes in this PR, would you mind retitling it accordingly?

config/default/capm3/namespace.yaml
mboukhalfa: Where can we check the security constraints definition for the restricted value?

tuminoid: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

tuminoid: @furkatgofurov7 The use case is exactly that: it doesn't change anything per se, but it also reinforces security posture, by making it impossible to create pods that violate the security standard in the namespace.

PR #960, opened 2023-05-03
🐛 Fix kubernetes version for e2e tests

kashifest:
This PR fixes e2e tests with correct k8s versions. Currently we are wrongly initializing the e2e integration tests with one older minor version but we should always initialize it to the latest k8s. This PR also adds a matrix to better understand which k8s version we are using for different e2e tests.


docs/e2e-test.md
```diff
@@ -102,3 +102,14 @@ For instance, to skip the upgrade E2E tests set `GINKGO_SKIP="[upgrade]"`
...
 The idea is to move them to the e2e upgrade test and then upgrade Ironic and BMO together with CAPM3 from the previous minor release to the latest.
+
+### Test matrix for k8s version
+
+Current e2e tests use the following Kubernetes versions for source and target clusters:
+
+| tests | source cluster | target init  | target final  |
+| ----------- | ------------ | ----------- | ------------- |
+| integration | v1.26.4 | v1.26.4 | X |
+| remediation | v1.26.4   | v1.25.2 | v1.26.4 |
```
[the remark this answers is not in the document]
kashifest: good catch, fixed

docs/e2e-test.md
```diff
@@ -102,3 +102,14 @@ For instance, to skip the upgrade E2E tests set `GINKGO_SKIP="[upgrade]"`
 Both Ironic and BMO upgrade tests currently only check that a new version can be rolled out (by going from `latest` to `main`).
 However, they do not actually upgrade from some older release to a newer since they are not yet integrated with the e2e upgrade test.
 The idea is to move them to the e2e upgrade test and then upgrade Ironic and BMO together with CAPM3 from the previous minor release to the latest.
+
+### Test matrix for k8s version
+
+Current e2e tests use the following Kubernetes versions for source and target clusters:
```
lentzi90, line 108: I know we have this terminology everywhere, but I think we should try to use "bootstrap", "management", "workload" and "self-hosted" terminology instead of "source" and "target". In CAPI "source" and "target" is only used in the context of pivoting. In other contexts it is quite confusing to use these terms.

We start with the bootstrap cluster, we pivot to the target management cluster that is self-hosted.

```suggestion
Current e2e tests use the following Kubernetes versions for bootstrap and self-hosted management clusters:
```

docs/e2e-test.md
```diff
@@ -102,3 +102,14 @@ For instance, to skip the upgrade E2E tests set `GINKGO_SKIP="[upgrade]"`
...
+
+| tests | source cluster | target init  | target final  |
```
lentzi90, line 110:
```suggestion
| tests | bootstrap cluster | self-hosted init  | self-hosted final  |
```

docs/e2e-test.md
lentzi90: Ah even more reason to not say target cluster! What about "CAPI cluster"? Or we can say "workload/management" cluster

docs/e2e-test.md
kashifest: Thanks for the review. I will push a change to this table soon with the uplift to v1.27.1. Can I address the comments in that?

docs/e2e-test.md
mboukhalfa: as management can also refer too bootstrap I suggest use metal3 cluster or metal3 provisioned cluster

docs/e2e-test.md
lentzi90: Sure

docs/e2e-test.md
kashifest: this is overcomplicating @mboukhalfa , anyways lets take this PR in so that I can test e2e with correct k8s versions and lets debate on the name of the cluster in a followup I would do soon.

PR #967, opened 2023-05-08
🐛 Deploy ironic before bmo in upgrade test

mboukhalfa:
BMO require some secret that will be created when ironic get deployed in order to be available so we are switching the order of deploying them


smoshiur1237:
Changing the order of installing BMO and ironic helps to resolve the volume issue of the BMO.

PR #977, opened 2023-05-10
:bug: Remove kubeconfig from .gitignore

honza:
In `baremetal/remote/remote.go`, we import `"sigs.k8s.io/cluster-api/util/kubeconfig"`, and the "kubeconfig" line in the `.gitignore` file causes this directory to be omitted from the vendor directory.  In vendorized scenarios, this causes builds to fail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,7 @@ envfile
 # kubeconfigs
 kind.kubeconfig
 minikube.kubeconfig
-kubeconfig
+/kubeconfig
 
 # ssh keys
 .ssh*
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-coverage.*
 
 # Ansible
 *.retry


.gitignore
```diff
@@ -26,7 +26,6 @@ envfile
 # kubeconfigs
 kind.kubeconfig
 minikube.kubeconfig
-kubeconfig
```
tuminoid, line 29: We're still generating `kubeconfig` in the Makefile, and need to have it ignored. Change it to `/kubeconfig` which ignores the file on top-level, but not in subdirectories.

tuminoid (changes requested):

.gitignore
honza: Done, thanks

.gitignore
```diff
@@ -10,7 +10,6 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-coverage.*
```
kashifest, line 13: why are we removing this line?

kashifest (approved):

PR #993, opened 2023-05-17
:bug: Modify kubeadm scripts to re-enable IPAM integration in e2e tests

Rozzii:
From the centos node image I have removed the code that blocekd cloud-init
network configuration features that also blocked the IPAM integration.

As the result of the node image change, some processes have to be modified in the centos kubeadm
scripts. Kubeadm script changes include:
  - Moving keepalived's K8s api monitoring from custom daemon to vrrp script in keeaplived
  - Adding a delay after the kubelet and crio are started to avoid flakes with kubeadm
    (this will be improved in a follow-up from a delay to an actual status check)
  - Fixing the startup order of eth0 and ironicendpoint interfaces

A polling check was added to delay the certmanager deployment during target cluster bootstrapping
until all pods are running, this helps to avoid accidentally assigning IPs from crio default cni
IP range that can happen if not all of the calico nodes are up.

Moved up the test aborting logic in e2e after-each sequence to actually make it work in a keep vm
as intended.

### DIFF
--- a/test/e2e/pivoting_based_feature_test.go
+++ b/test/e2e/pivoting_based_feature_test.go
@@ -8,6 +8,8 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


Rozzii:
/cc @kashifest 

Let's see if this is okay or not.

Rozzii:
keep-test-centos-e2e-integration-main

test/e2e/pivoting_based_feature_test.go
```diff
@@ -148,5 +150,10 @@ func createTargetCluster(k8sVersion string) (framework.ClusterProxy, *clusterctl
 		WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 	}, &result)
 	targetCluster := bootstrapClusterProxy.GetWorkloadCluster(ctx, namespace, clusterName)
+	framework.WaitForPodListCondition(ctx, framework.WaitForPodListConditionInput{
+		Lister:      targetCluster.GetClient(),
+		ListOptions: &client.ListOptions{LabelSelector: labels.Everything(), Namespace: "kube-system"},
+		Condition:   framework.PhasePodCondition(v1.PodRunning),
```
kashifest, line 156:
```suggestion
		Condition:   framework.PhasePodCondition(corev1.PodRunning),
```

kashifest:
Is the feature failure now related to the issue we are facing?

Rozzii:
> Is the feature failure now related to the issue we are facing?

I have not figured it out yet, what I can say at this point in time is that, during the remediation (so way after the target cluster has been set up) something happens with the target cluster's K8s API (it becomes unreachable) even though the .249 network is there and I can ping the .249 IP and keepalived works as expected we have 1 master and 2 backup keepalived.

 I am not sure at this point what happens with the API and whether it is related to my changes or not.

adilGhaffarDev:
/lgtm 
@mboukhalfa kindly check.

PR #1005, opened 2023-05-30
🌱 Make container registry for e2e calico manifest configurable

Rozzii:
The main motivations for this change are that:
- Using docker.io in CI triggers the rate limiting thus breaking the CI jobs.
- Allowing the user to have more flexibility over selecting the registry that provides calico containers.


lentzi90 (commented):
/lgtm

My only worry is if someone outside of Nordix is trying to run this. I guess they should not use the Nordix proxy. Not sure if there are limitations for who is allowed to use it.

Rozzii:
> /lgtm
> 
> My only worry is if someone outside of Nordix is trying to run this. I guess they should not use the Nordix proxy. Not sure if there are limitations for who is allowed to use it.

I agree with you, let's make the default to docker.io, I actually wanted it to be docker.io just copied the wrong string.

mboukhalfa (approved):

PR #1008, opened 2023-05-31
🐛 Remove m3m ownerRef from bmh

shibaPuppy:
don't control bmh power state if it goes to provisioned state
provisioned was completed, we should be able to power off the node on a case-by-case. 

Fixes #

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -474,9 +474,6 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 		if !consumerRefMatches(host.Spec.ConsumerRef, m.Metal3Machine) {
 			m.Log.Info("host already associated with another metal3 machine",
 				"host", host.Name)
-			// Remove the ownerreference to this machine, even if the consumer ref
-			// references another machine.
-			host.OwnerReferences, err = m.DeleteOwnerRef(host.OwnerReferences)
 			if err != nil {
 				return err
 			}
@@ -660,8 +657,6 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 
 		host.Spec.ConsumerRef = nil
 
-		// Remove the ownerreference to this machine.
-		host.OwnerReferences, err = m.DeleteOwnerRef(host.OwnerReferences)
 		if err != nil {
 			return err
 		}
@@ -708,12 +703,6 @@ func (m *MachineManager) Update(ctx context.Context) error {
 
 	// ensure that the BMH specs are correctly set.
 	err = m.setHostConsumerRef(ctx, host)
-	if err != nil {
-		return err
-	}
-
-	// ensure that the BMH specs are correctly set.
-	err = m.setHostSpec(ctx, host)
 	if err != nil {
 		return err
 	}
@@ -1122,13 +1111,6 @@ func (m *MachineManager) setHostConsumerRef(_ context.Context, host *bmov1alpha1
 		APIVersion: m.Metal3Machine.APIVersion,
 	}
 
-	// Set OwnerReferences.
-	hostOwnerReferences, err := m.SetOwnerRef(host.OwnerReferences, true)
-	if err != nil {
-		return err
-	}
-	host.OwnerReferences = hostOwnerReferences
-
 	// Delete nodeReuseLabelName from host.
 	m.Log.Info("Deleting nodeReuseLabelName from host, if any")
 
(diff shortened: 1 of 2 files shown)


furkatgofurov7:
> provisioned was completed, we should be able to power off the node on a case-by-case.

Thanks for the PR @shibaPuppy. Maybe I missing the use case here, but regardless of the host state, we should not power down the node. Can you elaborate more on your use-case?

shibaPuppy:
> > provisioned was completed, we should be able to power off the node on a case-by-case.
> 
> Thanks for the PR @shibaPuppy. Maybe I missing the use case here, but regardless of the host state, we should not power down the node. Can you elaborate more on your use-case?

@furkatgofurov7 
We are using libvirt node with CAPM3.

I think it should be possible to force power off the vm to proceed with the evacuation due to memory issues.

Similarly, I thought that the PM HW issue should be able to be forcibly turned off.

furkatgofurov7:
> I think it should be possible to force power off the vm to proceed with the evacuation due to memory issues.

that could be taken care by features, like [remediation](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/healthchecking.html) IMO and not forcing the controller to power down the node

shibaPuppy:
> that could be taken care by features, like [remediation](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/healthchecking.html) IMO rather than forcing the controller to power down the node

In some cases, after turning off the power, the node where the OS is installed is moved to an isolated rack for inspection.

I haven't used machine-health-check yet.
the kubelet may have issues, but the libvirt vm may not.
may not be able to proceed with reinstalling the OS.

Rozzii:
As far as I can see there were no decision on this PR and @shibaPuppy is still active so I have reopened this.
/cc @dtantsur @kashifest @zaneb @elfosardo @hroyrh @hardys    let's expand the scope of reviewers a a bit.

zaneb:
I agree that we should not be continuously forcing the power control to be on, which it appears we are currently doing because `setHostSpec()` is called from `Update()`. We just need to make sure it is on at the time the host is provisioned.

I think I'd prefer to set it inside the same block that we are setting the image, rather than relying on the state, since that could be racy. But in practice I expect this is probably ok.

/cc @honza

shibaPuppy:
@honza @zaneb @Rozzii 
really want this feature :sob:

Rozzii (approved):
/lgtm

kashifest:
> I think I'd prefer to set it inside the same block that we are setting the image, rather than relying on the state, since that could be racy. But in practice I expect this is probably ok.
> 
> /cc @honza

This also makes sense to me. @honza opinions?

zaneb:
Looking into this a bit more deeply...

I looked at [our fork of the original CAPBM](https://github.com/openshift/cluster-api-provider-baremetal/blob/master/pkg/cloud/baremetal/actuators/machine/actuator.go) (from the pre-CAPM3 times), and the `provisionHost()` function (equivalent of `setHostSpec()` here) also [sets `Online` to `true` unconditionally](https://github.com/openshift/cluster-api-provider-baremetal/blob/master/pkg/cloud/baremetal/actuators/machine/actuator.go#L590). That's OK though, because it is only called from `Create()`, so it only happens when the Machine is in the `Provisioning` phase.

What really changed in CAPM3 is that 31d394f1ab4323c74d11ccbe559dc46d6287533e started calling `setHostSpec()` not only from `Associate()` but also from `Update()`. Presumably this means it now gets called all the time no matter what phase the Machine is in. Since the purpose of this patch was to set the owner reference on the BMH, which we've already agreed (in #1358) is wrong, maybe we should revert that whole patch.

Rozzii:
/hold
I am holding this based on @zaneb 's reply https://github.com/metal3-io/cluster-api-provider-metal3/pull/1008#issuecomment-1891113489 , could we discuss this @shibaPuppy ?

shibaPuppy:
@Rozzii @zaneb 
I am not sure about my opinion as I lack a deep understanding of the CAPM3.

I think the `power on` function of baremetal nodes through a call to `provisionHost()` is essential, but I don't think there is a need to force the state in metal3machine when it is in `provisioned` state.

shibaPuppy:
> I think I'd prefer to set it inside the same block that we are setting the image, rather than relying on the state, since that could be racy. But in practice I expect this is probably ok.

Does the image block match that line?

https://github.com/metal3-io/cluster-api-provider-metal3/pull/1008/files#diff-cd2cdab0bdcc6b6e359db724b479ca86f879705fbfbcdddba7dbc876941e89bbR1121-L1151

@kashifest @zaneb @Rozzii

zaneb:
> @Rozzii @zaneb I am not sure about my opinion as I lack a deep understanding of the CAPM3.
> 
> I think the `power on` function of baremetal nodes through a call to `provisionHost()` is essential, but I don't think there is a need to force the state in metal3machine when it is in `provisioned` state.

Yes, correct. It should do it once and leave it.

I think we should stop calling `setHostSpec()` from `Update()`, which would have the same effect by reducing complexity instead of adding to it.

We should also stop setting the owner reference, because we agreed in #1358 that that's a bug.

A straight revert of https://github.com/metal3-io/cluster-api-provider-metal3/commit/31d394f1ab4323c74d11ccbe559dc46d6287533e would achieve both.

shibaPuppy:
/retitle 🐛 Remove m3m ownerRef from bmh

shibaPuppy:
@Rozzii @zaneb 
[owerRef commit](https://github.com/metal3-io/cluster-api-provider-metal3/commit/31d394f1ab4323c74d11ccbe559dc46d6287533e) was not reverted, so I wrote a commit to remove it.

dtantsur:
Do we need to keep `DeleteOwnerRef` for the time being? I'm thinking of an upgrade scenario where someone already has hosts with OwnerReferences that now are going to stick forever.

kashifest:
We agreed to close this in favor of https://github.com/metal3-io/cluster-api-provider-metal3/pull/1742#issuecomment-2160751565 since this PR was opened initially with different purpose and removing ownerreference is dealt on the other one

PR #1009, opened 2023-05-31
:bug: Refactor taint usage and flow controll in node_reuse e2e test 

Rozzii:
This PR:
- introduces wait loops that wait for all pods to be running at the end of scale up and
  scale down cycles to make sure that the workloads of the cluster won't get stuck or kept
  in scheduling longer than absolutely needed
- disables control plane node tainting during the node_reuse test to get rid of the
  complexity of managing tains parallel to scale up/down async processes
- noSchedule taints are disabled for control plane nodes because of an other reason also,
  namely that during most of the test there are no other nodes than control plane nodes
  available for scheduling so it doesn't make sense tainting them

### DIFF
--- a/test/e2e/node_reuse.go
+++ b/test/e2e/node_reuse.go
@@ -33,7 +33,7 @@ type NodeReuseInput struct {
 }
 
 func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
-	Logf("Starting node reuse tests")
+	Logf("Starting node reuse tests [node_reuse]")
 	input := inputGetter()
 	targetClusterClient := input.TargetCluster.GetClient()
 	managementClusterClient := input.ManagementCluster.GetClient()
@@ -47,6 +47,7 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
 	var (
 		controlplaneTaints = []corev1.Taint{{Key: "node-role.kubernetes.io/control-plane", Effect: corev1.TaintEffectNoSchedule},
 			{Key: "node-role.kubernetes.io/master", Effect: corev1.TaintEffectNoSchedule}}
+		emptyTaint = []corev1.Taint{}
 	)
 
 	const (
@@ -66,14 +67,21 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
 	ListMachines(ctx, managementClusterClient, client.InNamespace(input.Namespace))
 	ListNodes(ctx, targetClusterClient)
 
-	By("Untaint all CP nodes before scaling down machinedeployment")
-	controlplaneNodes := getControlplaneNodes(ctx, clientSet)
-	untaintNodes(ctx, targetClusterClient, controlplaneNodes, controlplaneTaints)
+	By("Untaint all CP nodes before scaling down machinedeployment [node_reuse]")
+	for controlplaneNodesPresent := 0; controlplaneNodesPresent < numberOfControlplane; {
+		for untaintedNodeCount := 1; untaintedNodeCount > 0; {
+			controlplaneNodes := getControlplaneNodes(ctx, clientSet)
+			controlplaneNodesPresent = len(controlplaneNodes.Items)
+			untaintedNodeCount = untaintNodes(ctx, targetClusterClient, controlplaneNodes, controlplaneTaints)
+			time.Sleep(10 * time.Second)
+		}
+	}
 
-	By("Scale down MachineDeployment to 0")
+	By("Scale down MachineDeployment to 0 [node_reuse]")
+	// this removes the worker node(s)
 	ScaleMachineDeployment(ctx, managementClusterClient, input.ClusterName, input.Namespace, 0)
 
-	Byf("Wait until the worker is scaled down and %d BMH(s) Available", 1)
+	Byf("Wait until the worker is scaled down and %d BMH(s) Available [node_reuse]", 1)
 	WaitForNumBmhInState(ctx, bmov1alpha1.StateAvailable, WaitForNumInput{
 		Client:    managementClusterClient,
 		Options:   []client.ListOption{client.InNamespace(input.Namespace)},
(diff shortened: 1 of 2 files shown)


Rozzii:
test-centos-e2e-feature-main

test/e2e/node_reuse.go
```diff
@@ -66,33 +67,49 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
-	untaintNodes(ctx, targetClusterClient, controlplaneNodes, controlplaneTaints)
+	By("Untaint all CP nodes before scaling down machinedeployment [node_reuse]")
+	for controlplaneNodesPresent := 0; controlplaneNodesPresent < numberOfControlplane; {
+		for untaintedNodeCount := 1; untaintedNodeCount > 0; {
+			controlplaneNodes := getControlplaneNodes(ctx, clientSet)
+			controlplaneNodesPresent = len(controlplaneNodes.Items)
+			untaintedNodeCount = untaintNodes(ctx, targetClusterClient, controlplaneNodes, controlplaneTaints)
+			time.Sleep(10 * time.Second)
+		}
+	}
```
lentzi90, line 78: Fine for now, but I think we should follow up with a refactor where we change this to use `Eventually`, as in 
https://github.com/metal3-io/cluster-api-provider-metal3/blob/7004edc3b84de44551e86732437f1631ba7b8abe/test/e2e/common.go#L436-L443

test/e2e/node_reuse.go
```diff
@@ -66,33 +67,49 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
 	})
 
-	By("Get the provisioned BMH names and UUIDs")
+	By("Wait for all pods to be running after workers were scaled down to 0 [node_reuse]")
+	// Give time for the pods to settle before starting the upgrade process
+	framework.WaitForPodListCondition(ctx, framework.WaitForPodListConditionInput{
+		Lister:      targetClusterClient,
+		ListOptions: &client.ListOptions{LabelSelector: labels.Everything(), Namespace: "kube-system"},
+		Condition:   framework.PhasePodCondition(corev1.PodRunning),
+	}, input.E2EConfig.GetIntervals(input.SpecName, "wait-all-pod-to-be-running-on-target-cluster")...)
```
lentzi90, line 98: Nice!

test/e2e/node_reuse.go
```diff
@@ -103,9 +120,17 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
 	kcpObj.Spec.RolloutStrategy.RollingUpdate.MaxSurge.IntVal = 0
-	Expect(helper.Patch(ctx, kcpObj)).To(Succeed())
+	Logf("Disable tainting of CP nodes during the remainder of the node reuse test [node_reuse]")
+	kcpObj.Spec.KubeadmConfigSpec.JoinConfiguration.NodeRegistration.Taints = emptyTaint
+	kcpObj.Spec.KubeadmConfigSpec.InitConfiguration.NodeRegistration.Taints = emptyTaint
+	patchRequestError := helper.Patch(ctx, kcpObj)
+	if patchRequestError != nil {
+		Logf("Error while patching KCP: %s [node_reuse]", patchRequestError.Error())
+	}
+	Expect(patchRequestError).NotTo(HaveOccurred())
```
lentzi90, line 130: This seems a bit unnecessary. Is it just to get an explicit log message out if there was an error? If there was an error the `Expect` would anyway show it.

test/e2e/node_reuse.go
```diff
@@ -168,26 +193,37 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
 		Intervals: input.E2EConfig.GetIntervals(input.SpecName, "wait-machine-running"),
 	})
 
-	By("Get the provisioned BMH names and UUIDs after upgrade")
+	By("Wait for all CP nodes to be present after the CP upgrade has ended [node_reuse]")
+	for controlplaneNodesPresent := 0; controlplaneNodesPresent < numberOfControlplane; {
+		controlplaneNodes := getControlplaneNodes(ctx, clientSet)
+		controlplaneNodesPresent = len(controlplaneNodes.Items)
+		time.Sleep(5 * time.Second)
+	}
```
lentzi90, line 209: This one should also use `Eventually`. But fine for now. We can do it in a follow up.

test/e2e/node_reuse.go
```diff
@@ -204,26 +240,32 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
-	By("Scale the controlplane down to 1")
+	By("Scale the controlplane down to 1 [node_reuse]")
 	ScaleKubeadmControlPlane(ctx, managementClusterClient, client.ObjectKey{Namespace: input.Namespace, Name: input.ClusterName}, 1)
 
-	Byf("Wait until controlplane is scaled down and %d BMHs are Available", numberOfControlplane)
+	Byf("Wait until controlplane is scaled down and %d BMHs are Available [node_reuse]", (numberOfControlplane - 1 + numberOfWorkers))
 	WaitForNumBmhInState(ctx, bmov1alpha1.StateAvailable, WaitForNumInput{
 		Client:    managementClusterClient,
 		Options:   []client.ListOption{client.InNamespace(input.Namespace)},
 		Replicas:  numberOfControlplane,
```
lentzi90, line 250: `Replicas` should match the expression in the message. Otherwise this can get very confusing.

test/e2e/node_reuse.go
```diff
@@ -204,26 +240,32 @@ func nodeReuse(ctx context.Context, inputGetter func() NodeReuseInput) {
...
 		Intervals: input.E2EConfig.GetIntervals(input.SpecName, "wait-cp-available"),
 	})
 
+	By("Wait for 1 CP nodes to be present after CP was scaled to 1 [node_reuse]")
+	for controlplaneNodesPresent := 0; controlplaneNodesPresent != 1; {
+		controlplaneNodes := getControlplaneNodes(ctx, clientSet)
+		controlplaneNodesPresent = len(controlplaneNodes.Items)
+		time.Sleep(5 * time.Second)
+	}
```
lentzi90, line 259: This should also use `Eventually`

lentzi90 (commented):
Nice work on this @Rozzii !
Considering how long it takes to test this I'm willing to take it in as it is. But there are some things we should fix in a follow up.

We should use `Eventually` instead of custom loops with sleep for example. It gives us the timeout functionality so we don't get stuck in those loops forever, and it is already a standard way of doing the same. :slightly_smiling_face: 

/lgtm

test/e2e/node_reuse.go
Rozzii: That is the thing, it was failing without a log, it could have been for an other reason but the expect was failing and there was no log on multiple tests, which was strange...

Rozzii:
Thanks @lentzi90  for the comments, I will address them in a followup, I am also planning to review the some timeouts, especially stuff related to deletion because those are really tight.

test/e2e/node_reuse.go
lentzi90: Weird. Maybe it would need a comment passed to `NotTo`. But anyway, let's keep it like this for now.

kashifest (approved):

Sunnatillo (commented):
.

PR #1012, opened 2023-06-01
🐛 Fix race condition when associating BMH to M3M

lentzi90:
**What this PR does / why we need it**:

Add a mutex lock to the Associate function. This function is not thread safe since we work on a common pool of BMHs. One BMH could potentially be associated with multiple M3Ms without this lock.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1004

PR #1013, opened 2023-06-01
🐛 Fix race condition when associating BMH to M3M

lentzi90:

Manual backport of #1012 
Add a mutex lock to the Associate function. This function is not thread safe since we work on a common pool of BMHs. One BMH could potentially be associated with multiple M3Ms without this lock.


mboukhalfa (approved):

tuminoid (commented):
/lgtm

PR #1026, opened 2023-06-12
:bug: Increase machine deletion timeout during e2e tests

Rozzii:
In case of e2e feature tests, 7m has proven to be insufficient in situations when the machine marked for deletion is hosting all the Metal3, CAPI, cert manager and calico controllers .

7m seems to be enough for machines that have relatively few pods but it could happen that a heavily utilized node is being deleted and 7min is not enough.


Sunnatillo (commented):
LGTM

PR #1045, opened 2023-06-15
🌱 Include all gomods in different directories for Dependabot

kashifest:
Dependabot is currently not updating gomodules in api/, test/ and hack/tools dir. This PR updates it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,3 +26,45 @@ updates:
       prefix: ":seedling:"
     labels:
       - "ok-to-test"
+  - package-ecosystem: "gomod"
+    directory: "/api"
+    schedule:
+      interval: "weekly"
+    ignore:
+      # Ignore controller-runtime as its upgraded manually.
+      - dependency-name: "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -26,3 +26,45 @@ updates:
       prefix: ":seedling:"
     labels:
       - "ok-to-test"
+  - package-ecosystem: "gomod"
+    directory: "/api"
```
furkatgofurov7, line 30: would love to see if directory could take an array of paths to folders, but that seems to be a limitation https://github.com/dependabot/dependabot-core/issues/2178 at least for now

furkatgofurov7 (approved):
/retitle 🌱 Include all gomods in different directories for Dependabot
/approve

tuminoid (commented):
/lgtm

.github/dependabot.yml
kashifest: Yes. It’s a pity. Already  feeling annoyed since this would most probably open separate PRs

tuminoid: Let's see what sort of PR storm we're about to be hit with :)

PR #1058, opened 2023-06-20
:bug: Removed mariadb restart check in certRotation test

adilGhaffarDev:
Removed mariadb restart check in certRotation test. Because we do not run any tests with mariadb for 1.2 release.


lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #1063, opened 2023-06-26
:sparkles: Added tls configuration flags for CAPM3.

as20203:
Added tls configuration flags for CAPM3.

This enables flags --tls-min-version and --tls-max-version flags, which can be used to limit TLS versions server accepts. Choices are TLS12 and TLS13, so you can force TLS 1.2 only, TLS 1.3 only, or (default) TLS 1.2 and TLS 1.3 is negotiated between server and client.

This also disables TLS 1.0 and TLS 1.1 as they're insecure and obsolete and not supported by Go TLS library.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,10 +18,12 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"math/rand"
 	"os"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


as20203:
@tuminoid PR for release 1.4

tuminoid (commented):
/lgtm
/assign @lentzi90 @mboukhalfa

PR #1070, opened 2023-06-27
:bug: Disable TLS 1.0 and TLS 1.1

as20203:
We found out that TLS 1.0 and TLS 1.1 are enabled despite being insecure. These changes are backported to release-1.3 branch, but without the configuration flags. Min version is set to TLS12 and Max version set to TLS13, with no config to change those.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"math/rand"
@@ -91,6 +92,12 @@ func main() {
 	ctrl.SetLogger(klogr.New())
 	restConfig := ctrl.GetConfigOrDie()
 	restConfig.UserAgent = "cluster-api-provider-metal3-manager"
+
+	tlsOptionOverrides, err := GetTLSOptionOverrideFuncs()
+	if err != nil {
+		setupLog.Error(err, "unable to add TLS settings to the webhook server")
+		os.Exit(1)
+	}
 	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme:                     myscheme,
 		MetricsBindAddress:         metricsBindAddr,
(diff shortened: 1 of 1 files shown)


as20203:
@tuminoid do we allow setting cipher suites using the cipher suite flag for the hardcoded version release 1.3?

tuminoid:
> @tuminoid do we allow setting cipher suites using the cipher suite flag for the hardcoded version release 1.3?

No configuration flags are added here. Just apply the default TLS config.

tuminoid:
/retitle :bug: Disable TLS 1.0 and TLS 1.1

main.go
```diff
@@ -225,6 +253,13 @@ func initFlags(fs *pflag.FlagSet) {
 		":9440",
 		"The address the health endpoint binds to.",
 	)
+	tlsCipherPreferredValues := cliflag.PreferredTLSCipherNames()
+	tlsCipherInsecureValues := cliflag.InsecureTLSCipherNames()
+	fs.StringVar(&tlsOptions.TLSCipherSuites, "tls-cipher-suites", "",
```
tuminoid: No flags whatsoever should be added in this PR. We only want to disable TLS 1.0 and TLS 1.1. Cipher suites are staying as default like they were before.

main.go
```diff
@@ -376,3 +411,75 @@ func setupWebhooks(mgr ctrl.Manager) {
 		os.Exit(1)
 	}
 }
+
+// GetTLSOptionOverrideFuncs returns a list of TLS configuration overrides to be used
+// by the webhook server.
+func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error) {
```
tuminoid, line 411: We can clean up this function a lot, or maybe even remove it, since there is no options to check for.

main_test.go
tuminoid, line 1: We probably don't have anything to test, since we don't provide any options in this PR.

as20203: Yeah, I've removed those tests since the values are hardcoded.

main.go
as20203: I've cleaned up this function with only hardcoded tls values and also removed other code which is redundant now such as providing options to this function which are hardcoded inside it.

tuminoid (commented):
/lgtm

PR #1072, opened 2023-06-30
🌱 Rename upgrade managment test and remove unecessary passed msg

mboukhalfa:
Rename upgrade management cluster to upgrade clusterctl and add the correct upgrade from release to the describe text.
Also remove passed message from the aftereach as it shows that even when the test fail


lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #1073, opened 2023-06-30
🐛 Set M3M concurrency to 1 and add warning

lentzi90:

We currently have an issue with concurrency when associating a BMH to a M3M. A mutex was already added before to the Associate function. However, this does not completely solve the issue, because a caching client is used. In other words, the next M3M to associate can sometimes use cached data about the BMHs and pick one that is already associated.

This is a temporary workaround for #1004


Sunnatillo (commented):
/lgtm

PR #1109, opened 2023-07-26
:seedling: Fitting e2e to new ironic deployment

mquhuy:
This is needed once https://github.com/metal3-io/baremetal-operator/pull/1316 is in use


furkatgofurov7:
@mquhuy should we mark it as hold since as I understood, this is blocked by BMO PR?

PR #1124, opened 2023-08-10
 🌱Add log and manifest collection for clusterctl upgrade test

smoshiur1237:
This will help to debug logs more efficiently and also able to collect logs during failed test .

### DIFF
--- a/test/e2e/upgrade_clusterctl_test.go
+++ b/test/e2e/upgrade_clusterctl_test.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -49,6 +50,7 @@ var _ = Describe(fmt.Sprintf("When testing cluster upgrade from releases %s > cu
 					// Override capi/capm3 versions exported in preInit
 					os.Setenv("CAPI_VERSION", "v1beta1")
 					os.Setenv("CAPM3_VERSION", "v1beta1")
+					os.Setenv("KUBECONFIG_BOOTSTRAP", bootstrapClusterProxy.GetKubeconfigPath())
 				},
 				PreWaitForCluster:           preWaitForCluster,
 				PreUpgrade:                  preUpgrade,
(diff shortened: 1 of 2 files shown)


Rozzii:
Looks okay to me.
/lgtm

test/e2e/upgrade_clusterctl_test.go
```diff
@@ -197,6 +199,21 @@ func preInitFunc(clusterProxy framework.ClusterProxy) {
 		Expect(clusterProxy.Apply(ctx, certManagerYaml)).ShouldNot(HaveOccurred())
 	}
 
+	By("Fetch manifest for bootstrap cluster")
+	path := filepath.Join(os.Getenv("CAPM3PATH"), "scripts")
+	cmd := exec.Command("./fetch_manifests.sh") // #nosec G204:gosec
```
kashifest, line 204: is there a way to avoid this invoking of shell inside go code? As you can see you have to ignore the gosec error and we are trying to get rid of such invocations from other places too.

test/e2e/upgrade_clusterctl_test.go
```diff
@@ -295,10 +312,36 @@ func preUpgrade(clusterProxy framework.ClusterProxy) {
 func preCleanupManagementCluster(clusterProxy framework.ClusterProxy) {
 	// Abort the test in case of failure and keepTestEnv is true during keep VM trigger
 	if CurrentSpecReport().Failed() {
+		// Fetch logs in case of failure in management cluster
+		By("Fetch logs from management cluster")
+		path := filepath.Join(os.Getenv("CAPM3PATH"), "scripts")
+		cmd := exec.Command("./fetch_target_logs.sh") // #nosec G204:gosec
```
kashifest, line 318: same here

kashifest (commented):
Codewise looks good but not so much in favor of adding more gosec exceptions.

test/e2e/upgrade_clusterctl_test.go
smoshiur1237: Running of these script are part of e2e test and used  the approach in different place of e2e tests. I think we need to check for a  solution for that separately with new ticket.

kashifest: theres already ticket for that , I am concerned about adding new cases when old cases are being worked upon.

kashifest:
/approve
I have checked the tickets and it seems we will get rid of the fetch_ scripts and that would anyhow remove any exec commands here.

PR #1140, opened 2023-08-24
🌱 Bump CAPI to v1.3.10

kashifest:
Bump CAPI to v1.3.10 and update releasing doc to make sure we don't miss CAPI bumps before releasing. This PR also
fixes deprecated use of docker containerStop function


lentzi90 (commented):
/approve

PR #1179, opened 2023-09-04
🌱 Bump kind and kustomize and update bmh CRD

kashifest:
Apart from the bumps, this PR also removes the old way of installing kustomize which is through ensure-kustomize script. We have the kustomize dependency in hack/tools/go.mod and we can leverage that. This PR also updates the BMH crd in the example folder.


kashifest:
> There are some strange things in the `make help` output.
> 
> ```
> tools:
>   controller                                          Build controller-gen from tools folder.
>   golangci-                                           golangci-lint
>   moc                                                 Build mockgen from tools folder.
>   envs                                                Build envsubst from tools folder.
>   kustomize                                           Build a local copy of kustomize.
>   kusto                                               Build kustomize from tools folder.
>   envsubst                                            Build envsubst from tools folder.
> ```
> 
> I would say that some of these should not be in the help text at all because they are not meant for humans, just as necessary dependencies. For example, there is no point having a help text for both `kustomize` and `kusto`. The first issue is `kusto` is not a real target so it is not possible to do `make kusto`, and the second is that it is unclear which to use. Therefore I would say we should hide `kusto`. Same goes for controller-gen, mockgen and `envs` (short for `envsubst`) and `golangci-`. It could be useful to have a proper target for `golangci-lint` though, but having one that does not work in the help text is not useful.

true, checking

kashifest:
Fixed.  Keeping envsubst binary to be built locally for the user. 

```
tools:
  controller                                          Build controller-gen from tools folder.
  golangci-lint                                       Build a local copy of golangci-lint.
  kustomize                                           Build a local copy of kustomize.
  envsubst                                            Build envsubst from tools folder.
```

kashifest:
building clusterctl binary is broken since we do not have any dependancy for clusterctl in go.mod. Will fix it in a followup. https://github.com/metal3-io/cluster-api-provider-metal3/issues/1180

lentzi90 (commented):
/approve

PR #1196, opened 2023-09-07
🌱 Minor cleanups and added tests

lentzi90:

    This adds a few tests for ensureM3IPClaim, with and without
    EnableBMHNameBasedPreallocation set to true. Also made sure that the label
    map exists on the Metal3Data before trying to add labels to it.

- Fixes some naming in the unit tests.
- Adds a bit more logs and improves misleading messages.

Fixes #

### DIFF
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -619,6 +619,7 @@ func (m *DataManager) m3IPClaimObjectMeta(name, poolRefName string, preallocatio
 // ensureM3IPClaim ensures that a claim for a referenced pool exists.
 // It returns the claim and whether to fetch the claim again when fetching IP addresses.
 func (m *DataManager) ensureM3IPClaim(ctx context.Context, poolRef corev1.TypedLocalObjectReference) (reconciledClaim, error) {
+	m.Log.Info("Ensuring M3IPClaim for Metal3Data", "Metal3Data", m.Data.Name)
 	ipClaim, err := fetchM3IPClaim(ctx, m.client, m.Log, m.Data.Name+"-"+poolRef.Name, m.Data.Namespace)
 	if err == nil {
 		return reconciledClaim{m3Claim: ipClaim}, nil
@@ -638,7 +639,7 @@ func (m *DataManager) ensureM3IPClaim(ctx context.Context, poolRef corev1.TypedL
 	if m3dt == nil {
 		return reconciledClaim{m3Claim: ipClaim}, nil
 	}
-	m.Log.Info("Fetched Metal3DataTemplate")
+	m.Log.Info("Fetched Metal3DataTemplate", "Metal3DataTemplate", m3dt.Name)
 
 	// Fetch the Metal3Machine, to get the related info
 	m3m, err := m.getM3Machine(ctx, m3dt)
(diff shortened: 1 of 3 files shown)


baremetal/metal3data_manager.go
```diff
@@ -648,17 +652,17 @@ func (m *DataManager) ensureM3IPClaim(ctx context.Context, poolRef corev1.TypedL
...
+	m.Log.Info("Fetched Metal3Machine", "Metal3Machine", m3m.Name)
 
 	// Fetch the BMH associated with the M3M
 	bmh, err := getHost(ctx, m3m, m.client, m.Log)
 	if err != nil {
 		return reconciledClaim{m3Claim: ipClaim}, err
 	}
 	if bmh == nil {
-		return reconciledClaim{m3Claim: ipClaim}, WithTransientError(nil, requeueAfter)
+		return reconciledClaim{m3Claim: ipClaim}, WithTransientError(errors.New("no associated BMH yet"), requeueAfter)
```
lentzi90, line 663: We were already returning an error, so might as well make it a bit more descriptive...

baremetal/metal3data_manager.go
```diff
@@ -1419,7 +1425,7 @@ func fetchM3IPClaim(ctx context.Context, cl client.Client, mLog logr.Logger,
 	}
 	if err := cl.Get(ctx, metal3ClaimName, metal3IPClaim); err != nil {
 		if apierrors.IsNotFound(err) {
-			errMessage := "Address claim not found, requeuing"
+			errMessage := "Address claim not found"
```
lentzi90, line 1428: We are not deciding at this point if we are requeuing or not. In fact, we are calling this function several times in `ensureM3IPClaim` and not necessarily requeuing at all.

baremetal/metal3data_manager_test.go
```diff
@@ -91,7 +99,7 @@ var _ = Describe("Metal3Data manager", func() {
 		expectedErrorSet bool
 	}
 
-	DescribeTable("Test CreateSecret",
+	DescribeTable("Test Reconcile",
```
lentzi90, line 102: Test is for the `Reconcile` function and is using `testCaseReconcile` type. The name was just wrong

baremetal/metal3data_manager.go
```diff
@@ -596,6 +596,9 @@ func getReferencedPools(m3dt infrav1.Metal3DataTemplate) (map[string]corev1.Type
 // will be added to Data labels in case preallocation is enabled.
 func (m *DataManager) m3IPClaimObjectMeta(name, poolRefName string, preallocationEnabled bool) *metav1.ObjectMeta {
 	if preallocationEnabled {
+		if m.Data.Labels == nil {
+			m.Data.Labels = map[string]string{}
+		}
```
kashifest, line 601: I think the label is created much much before this, for example in this data manager code if you go back to the function calls `m3IPClaimObjectMeta` <- `ensureM3IPClaim` <-  `getAddressesFromPool` <-  `createSecrets` , the `createSecrets` will fetch m3DT object based on cluster label. Infact data manager wont even exist if the cluster label is missing https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/controllers/metal3data_controller.go#L84-L107. So I wonder if adding this nil check would add any value now. Ofcourse for unit test it might add some value.

kashifest (commented):
/approve
One comment inline which doesn't hamper merging.

baremetal/metal3data_manager.go
lentzi90: It is not just about unit tests, it is about reusable code. Someone may try to use this in another place in the future, where this assumption that something added the labels does not hold. I would say it is always a good idea to check. If we want to make the assumption explicit I would also be fine with returning an error if the labels are missing, but just causing a stack trace because of a nil map is not ok.

kashifest: I would rather return at error if a label is missing at this point based on the current assumption.

kashifest: However, I don’t have any strong opinion not this. So we can keep it like this as well

lentzi90: Yes I would be fine with that, although it would be a bit bigger change since it would require changing it so it can return an error

PR #1202, opened 2023-09-14
:seedling: Add healthcheck e2e test

huutomerkki:
This PR adds the healthcheck test as e2e test. Ansible version was removed in https://github.com/metal3-io/metal3-dev-env/pull/1252 .


test/e2e/common.go
```diff
@@ -670,6 +679,125 @@ func runCommand(logFolder, filename, machineIP, user, command string) error {
...
+		"nodepool": "nodepool-0",
+	}
+	healthcheck, err := DeployMachineHealthCheck(ctx, cli, namespace, clusterName, remediationTemplateName, healthCheckName, matchLabels)
+	if err != nil {
+		return nil, fmt.Errorf("creating worker healthcheck failed: %w", err)
+	}
+	return healthcheck, nil
+}
+
+func DeployMachineHealthCheck(ctx context.Context, cli client.Client, namespace, clusterName, remediationTemplateName, healthCheckName string, matchLabels map[string]string) (*clusterv1.MachineHealthCheck, error) {
```
lentzi90, line 726: For these new functions, could you please add some simple doc string explaining what they do. It doesn't have to be fancy, just something simple like:

```suggestion
// DeployMachineHealthCheck creates a Metal3RemediationTemplate and MachineHealthCheck.
func DeployMachineHealthCheck(ctx context.Context, cli client.Client, namespace, clusterName, remediationTemplateName, healthCheckName string, matchLabels map[string]string) (*clusterv1.MachineHealthCheck, error) {
```

I would probably rename the `Deploy*` functions to `Create*` but that is more a suggestion. It is just that they use `cli.Create` and we already have many other functions that `create` stuff, like `createTargetCluster`.

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,80 @@
...
+	BeforeEach(func() {
+		osType := strings.ToLower(os.Getenv("OS"))
+		Expect(osType).ToNot(Equal(""))
+		validateGlobals(specName)
+
+		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+	})
+
+	It("Should test healthchecking", func() {
```
lentzi90, line 24: Perhaps something more descriptive for this
```suggestion
	It("Should remediate unhealthy machines", func() {
```

lentzi90 (commented):
Nice job with this test! I added some comments but it is really just minor stuff.

test/e2e/common.go
```diff
@@ -655,7 +658,9 @@ func runCommand(logFolder, filename, machineIP, user, command string) error {
 	if err != nil {
 		return fmt.Errorf("couldn't open a new session: %w", err)
 	}
-	logFile := path.Join(logFolder, filename)
+	if err := session.Run("sudo " + command + "\n"); err != nil {
+		return fmt.Errorf("unable to send command %q: %w", "sudo "+command, err)
+	}
```
lentzi90, line 663: I may be missing something but isn't this doing the exact same thing as lines 668-670? Do we need both?

test/e2e/common.go
```diff
@@ -581,6 +582,7 @@ func MachineToVMName(ctx context.Context, cli client.Client, m *clusterv1.Machin
 	return "", fmt.Errorf("no matching Metal3Machine found for current Machine")
 }
 
+// Gets IPAddress based on machine, from machine -> m3machine -> m3data -> IPAddress.
```
lentzi90, line 585: Sorry for being nit picky. Could you make all of these comments follow the go doc standard so that they begin with the name of the function? For example this one should be
```suggestion
// MachineToIPAddress gets IPAddress based on machine, from machine -> m3machine -> m3data -> IPAddress.
```

I know it may seem redundant, but this way all kinds of developer tools or documentation generators can pick them up and make nice things happen :slightly_smiling_face: 
If you want to know more about go doc comments, check this nice page: https://go.dev/doc/comment

huutomerkki: Of course, I just hadn't realized that was the format :)

lentzi90 (commented):
/approve

lentzi90:
/hold
Let's check also that the test runs and passes. This is supposed to run as part of the normal e2e tests, right?

/test-ubuntu-integration-main

mboukhalfa (commented):
I appreciated the way you simplified the implementation by defining a function for each operation then call them in the `It` test block. I guess It would be also great if we followed the other test pattern and we defined the full test in a function in a separate file `healthcheck.go` then we call it from the `describe` block with main inputs [ e2e config, cluster, namespace...]

lentzi90:
I'm not seeing the healthcheck test run as part of the feature tests :slightly_frowning_face: 
Probably we will need to do something similar to this: https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/pivoting_based_feature_test.go#L91-L129
I.e. call the test from one of the "top" tests. That would unfortunately require a bit of refactor, basically what @mboukhalfa mentioned above.

lentzi90:
Alternatively, it may be possible to just add the relevant "tags" `[pivoting] [features]` so it is selected when we tell ginkgo to focus on these tests.

mboukhalfa:
This test is related to the current metal3remediation https://github.com/metal3-io/cluster-api-provider-metal3/blob/5135b76045c7f1da440272318395948c961378de/test/e2e/metal3remediation.go#L55 we have a todo comment there for adding the healthcheck test. I guess the existing metal3remediation function is slightly more generic and can include the current healthcheck, so probably we should rename the existing one to avoid confusion and call all these test from feature remediation test https://github.com/metal3-io/cluster-api-provider-metal3/blob/5135b76045c7f1da440272318395948c961378de/test/e2e/remediation_based_feature_test.go#L60

test/e2e/healthcheck_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("When testing healthcheck [healthcheck]", func() {
```
adilGhaffarDev, line 14:
```suggestion
var _ = Describe("When testing healthcheck [healthcheck] [features]", func() {
```
If you'd like to run it with the feature tests, add this.

adilGhaffarDev:
> I appreciated the way you simplified the implementation by defining a function for each operation then call them in the `It` test block. I guess It would be also great if we followed the other test pattern and we defined the full test in a function in a separate file `healthcheck.go` then we call it from the `describe` block with main inputs [ e2e config, cluster, namespace...]

I agree with that to some extent. However, there are benefits to adding the test as a separate Ginkgo focus. When someone needs to specifically test the health check for debugging purposes, having it as a focused test makes it very convenient to execute.

adilGhaffarDev:
looks good to me. @mboukhalfa can give the final lgtm if he is ok with this way of implementation.

kashifest:
/test-ubuntu-e2e-feature-main 
cc @mboukhalfa

mboukhalfa (approved):

smoshiur1237:
Thanks for adding the health check. Great work.
LGTM

test/e2e/healthcheck_test.go
smoshiur1237: This is resolves and decided to run healthcheck a part of remediation test.

PR #1214, opened 2023-09-21
🐛 fix ensure-golangci-lint digest for linux/amd64

tuminoid:
The digest for ensure-golangci-lint linux/amd64 is incorrect, and also the script fails to unbound variable when checking the SHA.

This fixes "make lint", and "make test", which are no longer utilized as CI tests, but only run locally.

Additionally, make the digest check also check the ARCH, which is the root cause of these mismatches.

Later on, I will add project-infra test to actually verify these scripts work.


lentzi90 (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #1213, opened 2023-09-20
🌱 bump gosec to 2.17.0, and fix gosec for submodules

tuminoid:
Gosec does not handle submodules well. Ignore them and run them separately so they run against correct dependencies.

Bump gosec to 2.17.0, and fix new errors it has discovered.
Remove G107 exclude, it is no longer present.

NOTE: In CI, the gosec image is coming via project-infra, gosec image change is only affecting local runs for now.

PR #1219, opened 2023-09-26
:bug:  fix too tight file permissions in e2e

tuminoid:
A different user is generating sha256sum for the images than is reading them back in. Change back to 544 as it was before #1213.


lentzi90 (commented):
/approve

PR #1225, opened 2023-09-27
🌱 Bumping CAPI to 1.5.2

adilGhaffarDev:
- Bump CAPI to v1.5.2
- Bump Cert-manager to v1.13.0
- Bump GO to 1.20.8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.20.7@sha256:741d6f9bcab778441efe05c8e4369d4f8ff56c9a635a97d77f55d8b0ec62f907
+ARG BUILD_IMAGE=docker.io/golang:1.20.8@sha256:6b29720c5598cb40b23dd38f522050a698d427c3a5696c5efe1ed22bd45b1396
 ARG BASE_IMAGE=gcr.io/distroless/static:nonroot@sha256:9ecc53c269509f63c69a266168e4a687c7eb8c0cfd753bd8bfcaa4f58a90876f
 
 # Build the manager binary on golang image
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ SHELL:=/usr/bin/env bash
 
 .DEFAULT_GOAL:=help
 
-GO_VERSION ?= 1.20.4
+GO_VERSION ?= 1.20.8
 # Use GOPROXY environment variable if set
 GOPROXY := $(shell go env GOPROXY)
 ifeq ($(GOPROXY),)
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -115,7 +115,7 @@ echo "Generated ${METAL3PLANE_GENERATED_FILE}"
 echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 
 # Get Cert-manager provider components file
-curl --fail -Ss -L -o "${COMPONENTS_CERT_MANAGER_GENERATED_FILE}" https://github.com/cert-manager/cert-manager/releases/download/v1.12.3/cert-manager.yaml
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ SHELL:=/usr/bin/env bash
 
 .DEFAULT_GOAL:=help
 
-GO_VERSION ?= 1.20.4
+GO_VERSION ?= 1.20.8
```
tuminoid, line 23: Same quesiton as in IPAM. This is not bumping Go.

examples/generate.sh
```diff
@@ -115,7 +115,7 @@ echo "Generated ${METAL3PLANE_GENERATED_FILE}"
 echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 
 # Get Cert-manager provider components file
-curl --fail -Ss -L -o "${COMPONENTS_CERT_MANAGER_GENERATED_FILE}" https://github.com/cert-manager/cert-manager/releases/download/v1.12.3/cert-manager.yaml
+curl --fail -Ss -L -o "${COMPONENTS_CERT_MANAGER_GENERATED_FILE}" https://github.com/cert-manager/cert-manager/releases/download/v1.3.0/cert-manager.yaml
```
kashifest, line 118:
```suggestion
curl --fail -Ss -L -o "${COMPONENTS_CERT_MANAGER_GENERATED_FILE}" https://github.com/cert-manager/cert-manager/releases/download/v1.13.0/cert-manager.yaml
```

kashifest (changes requested):

Makefile
adilGhaffarDev: I have updated the docker file, let me know if I have missed somewhere else. I will update ipam PRs too.

Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.20.7@sha256:741d6f9bcab778441efe05c8e4369d4f8ff56c9a635a97d77f55d8b0ec62f907
+ARG BUILD_IMAGE=docker.io/golang:1.20.8@sha256:1b46d1164aceec47750cc0c3a6f8c44ee8050a0745c07affcef73518544a2b1f
```
tuminoid, line 16: This SHA is incorrect. `6b29720c5598cb40b23dd38f522050a698d427c3a5696c5efe1ed22bd45b1396` is correct SHA for golang:1.20.8 for amd64.

adilGhaffarDev: I checked from here: https://hub.docker.com/layers/library/golang/1.20.8/images/sha256-1b46d1164aceec47750cc0c3a6f8c44ee8050a0745c07affcef73518544a2b1f?context=explore 
I will change it

kashifest: Yes its a bummer and pretty inconvenient that the digest on the hub doesn't match arch specific digest when you download, issue is still open as I can see https://github.com/docker/hub-feedback/issues/1925

kashifest (approved):
/approve

Dockerfile
tuminoid: It is just misleading, not mismatching. Digest on the page is manifest digest, not image digest for architecture.
Easiest way to get correct one is to just pull it on a arm64 linux machine.

kashifest: Thats why I called mismatch. The hub has os/arch tab and the digest under the image changes when you switch os/arch. So I thought the digest was image digest. Only after you pointed out, I came to know that digest is actually manifest digest. Anyways arm64 or you meant amd64?

tuminoid: I meant amd64 like I said. Arm64 would be incorrect for us.

kashifest: Ok, previous comment has arm64 so I got confused.

tuminoid: Oh, I did not even notice the latter one :) Kept staring the original comment, looking for the typo.

tuminoid (commented):
/lgtm

PR #1233, opened 2023-09-28
🌱 Bump capi/test module to v1.4.7

kashifest:
Also removes the unnecessary replaces which also was also wrongly pointing to v1.4.6


lentzi90 (commented):
/lgtm

PR #1238, opened 2023-10-02
:seedling: add hack/verify-release.sh script

tuminoid:
Script that can be used to verify release content.

```console
+# USAGE:
+#
+# This script aims to verify a release content per docs/releasing.md
+# is all done, all images are built and release in general good to go.
+# It can be executed before making a release tag to verify Go dependencies
+# and vulnerabilities are already fixed.
+#
+# Git setup:
+# This script expects to be executed in the root directory of CAPM3
+# repository, with the release commit/tag in question checked out.
+#
+# Command line arguments:
+# arg1: mandatory: version without leading v, eg. 1.5.0
+#
+# Environment variables:
+# GITHUB_TOKEN: mandatory: your bearer token that has access to the release
+# REMOTE: optional: use this git remote for tag checks: Default: autodetected
+# CONTAINER_RUNTIME: optional: container runtime binary. Default: docker
```

For example, check `release-1.5` content for upcoming release:

```console
$ cd $CAPM3_GIT
$ git checkout release-1.5
$ export GITHUB_TOKEN="some token"
$ ./hack/verify-release.sh 1.5.2  # non-existing tag/version
...
<some pre-tagging checks are run>
```
and after the release has been tagged, release notes formatted and what else `docs/releasing.md` tells us to do:

```console
$ git tag -s -a v1.5.2 -m v1.5.2
$ git push origin v1.5.2 
# and whatever releasing.md wants us to do
$ ./hack/verify-release.sh 1.5.2  # v1.5.2 is existing tag now
...
<many post-tagging checks are run>
<some pre-tagging checks are re-run>
```

Make `releasing.md` refer to this script.

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -1,0 +1,614 @@
+#!/usr/bin/env bash
+#
+# Copyright 2023 The Metal3 Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# USAGE:
+#
+# This script aims to verify a release content per docs/releasing.md
+# is all done, all images are built and release in general good to go.
+# It can be executed before making a release tag to verify Go dependencies
+# and vulnerabilities are already fixed.
+#
+# Git setup:
+# This script expects to be executed in the root directory of CAPM3
+# repository, with the release commit/tag in question checked out.
+#
+# Command line arguments:
+# arg1: mandatory: version without leading v, eg. 1.5.0
+#
+# Environment variables:
+# GITHUB_TOKEN: mandatory: your bearer token that has access to the release
+# REMOTE: optional: use this git remote for tag checks: Default: autodetected
+# CONTAINER_RUNTIME: optional: container runtime binary. Default: docker
+
+set -eu
+# we are using plenty of subshell pipes, and catch errors elsewhere
+set +o pipefail
+
+# enable support for **/go.mod
+shopt -s globstar
+
+# user input
+VERSION="${1:?release version missing, provide without leading v. Example: 1.5.0}"
+GITHUB_TOKEN="${GITHUB_TOKEN:?export GITHUB_TOKEN with permissions to read unpublished release notes}"
+
+# if CONTAINER_RUNTIME is set, we will use crane and osv-scanner from images
+# otherwise, we will expect them to be installed binaries. This allows some
+# flexibility for the Mac users, where Docker Desktop is a bit problematic.
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-}"
+# correct remote will be autodetected, if empty
+REMOTE="${REMOTE:-}"
+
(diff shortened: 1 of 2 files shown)


tuminoid:
/cc @adilGhaffarDev @lentzi90 @kashifest @smoshiur1237 

Please take a look and test, everything we discussed should be now implemented.

hack/verify-release.sh
```diff
@@ -0,0 +1,489 @@
...
+{
+    # we support origin (default) and upstrea (if cloned with "gh" CLI tool)
+    echo "Detecting remote ..."
+
+    if [[ -z "${REMOTE}" ]]; then
+        REMOTE="$(git remote -v | grep "${PROJECT} (fetch)" | awk '{print $1;}')"
+
+        if ! [[ "${REMOTE}" =~ ^(origin|upstream)$ ]]; then
+            echo "WARNING: detected remote '${REMOTE}' is not supported"
+        fi
```
lentzi90, line 170: The script failed for me here. I think the issue is that my remotes have `.git` at the end of the URL (exactly what you get if you copy the ssh URL from GitHub).
To solve this I suggest making the `grep` a bit more flexible, perhaps with a regex.

Somehow it just exited the script for me at this point. I do not get why it didn't echo the warning about unsupported remote :thinking:

lentzi90 (commented):
Awesome!
/approve

hack/verify-release.sh
tuminoid: Fixed

tuminoid:
/hold

One more improvement coming, verifying IPAM manifests are bumped with the go.mod. We have forgotten to do that few times now.

tuminoid:
/unhold

IPAM manifest check added.

tuminoid:
/override test-centos-e2e-integration-main
/override test-ubuntu-integration-main

Save the trees, don't run irrelevant tests.

hack/verify-release.sh
```diff
@@ -0,0 +1,587 @@
...
+
+    echo "Verifying container images ..."
+
+    for image_and_tag in "${container_images[@]}"; do
+        image="${image_and_tag/:*}"
+        tag="${image_and_tag/*:}"
+
+        # quay paginates 50 items at a time, so it is simpler to use gcrane
+        # to list all the tags, than DIY parse the pagination logic
+        if ! "${CONTAINER_RUNTIME}" run --rm gcr.io/go-containerregistry/gcrane \
```
tuminoid, line 362: Crane and osv-scanner are now executed via Docker, maybe I could just expect user to have them installed via "go install" instead. Docker is OK for Linux, but Mac users might have an issue with it.

/hold

tuminoid: OK. I changed it so that if `CONTAINER_RUNTIME` is defined, it will run gcrane and osv-scanner as images, otherwise it will check for pre-installed binaries. This caters for most users.

/unhold

kashifest:
I am getting this on MacOS

```
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
```

might be related to this https://stackoverflow.com/questions/31396985/why-is-mktemp-on-os-x-broken-with-a-command-that-worked-on-linux

tuminoid:
> I am getting this on MacOS
> 
> ```
> mktemp: illegal option -- -
> usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
>        mktemp [-d] [-q] [-u] -t prefix 
> ```
> 
> might be related to this https://stackoverflow.com/questions/31396985/why-is-mktemp-on-os-x-broken-with-a-command-that-worked-on-linux

Fixed, thanks for testing on Mac.

tuminoid:
I'll cherry-pick without releasing.md to branches.

PR #1246, opened 2023-10-05
🌱 fix module printout in hack/verify-release.sh

tuminoid:
In case module group versions mismatch, grep failed to print out the mismatching modules. Fix the grep with "-E", and remove unnecessary curlys around it as well, after grep was optimized.


tuminoid:
/override test-centos-e2e-integration-main 
/override test-ubuntu-integration-main 

Save the trees.

lentzi90 (commented):
/approve

smoshiur1237 (approved):
/lgtm

PR #1251, opened 2023-10-06
🐛 Fix IP-reuse deletion bug

### LINKED ISSUE #1250: IP reuse: Deleting M3D cascades to IPClaims across namespaces

1. Initialize CAPM3 with [IP reuse enabled](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/ip_reuse.md#how-to-enable-ip-reuse). This is very important to trigger the bug!
2. Create two identical clusters in separate namespaces.
   - They should have an IPPool each and allocate IPAddresses from them (no need to use preAllocations though).
   - Make sure that the M3DTs are named the same in each namespace.
3. Check that you have (at least) one M3D and IPClaim in each namespace
4. Delete one cluster. Alternatively: Delete just one M3D.
5. Check the IPClaims again. They are gone in both namespaces!

1. The IPClaim(s) in the other namespace should not be affected when the M3D is deleted from the first namespace!
2. (The M3D should not be deleted while there is a M3DC still using it. We can create a separate issue for this.)

We suspect that the issue is in the ownerRef of the IPClaims or the code around that. They do not include the namespace!

```yaml
  ownerReferences:
  - apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    controller: true
    kind: Metal3Data
    name: my-cluster-controlplane-template-0
    uid: ec4a98f2-f388-43b9-a643-c320ec21942f
```
The UID should still be different, but perhaps we are not checking this properly. I remember some comment about not checking UIDs since they can change when doing `clusterctl move`.

**Environment:**

- Cluster-api version: v1.4.7 and v1.5.2 tested
- Cluster-api-provider-metal3 version: v1.4.5 and v1.5.1 tested
- Environment (metal3-dev-env or other): other
- Kubernetes version: (use `kubectl version`): v1.27.3

/kind bug

lentzi90:

This limits the listing of IPClaims to the namespace of the Metal3Data. Without this limit, CAPM3 could delete IPClaims in other namespaces, provided that they have the same labels. The labels are based on the name of the IPPool and Metal3Data(Template), which can be the same in other namespaces.
This also adds a few unit tests that check specifically the behavior when deleting IPClaims when there are multiple, in multiple namespaces. One of these tests trigger the bug described in #1250.

Fixes #1250

### DIFF
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -1440,6 +1440,7 @@ func fetchM3IPClaim(ctx context.Context, cl client.Client, mLog logr.Logger,
 func (m *DataManager) fetchIPClaimsWithLabels(ctx context.Context, pool string) ([]ipamv1.IPClaim, error) {
 	allIPClaims := ipamv1.IPClaimList{}
 	opts := []client.ListOption{
+		&client.ListOptions{Namespace: m.Data.Namespace},
 		client.MatchingLabels{
 			DataLabelName: m.Data.Name,
 			PoolLabelName: pool,
--- a/baremetal/metal3data_manager_test.go
+++ b/baremetal/metal3data_manager_test.go
@@ -1554,6 +1554,169 @@ var _ = Describe("Metal3Data manager", func() {
 		}),
 	)
 
+	type testCaseMultiReleaseAddressFromM3Pool struct {
+		m3d      *infrav1.Metal3Data
+		poolRef  corev1.TypedLocalObjectReference
+		ipClaims []ipamv1.IPClaim
+	}
+
+	DescribeTable("Test releaseAddressFromM3Pool with multiple namespaces",
+		func(tc testCaseMultiReleaseAddressFromM3Pool) {
+			objects := []client.Object{}
+			for i := range tc.ipClaims {
+				// To make the test entries a bit smaller, we add the
+				// .spec.pool here based on the labels.
+				tc.ipClaims[i].Spec.Pool.Name = tc.ipClaims[i].Labels[PoolLabelName]
+				objects = append(objects, &tc.ipClaims[i])
+			}
+			fake := fake.NewClientBuilder().WithScheme(setupScheme()).WithObjects(objects...).Build()
+			dataMgr, err := NewDataManager(fake, tc.m3d,
+				logr.Discard(),
+			)
+			Expect(err).NotTo(HaveOccurred())
+			err = dataMgr.releaseAddressFromM3Pool(
+				context.TODO(), tc.poolRef,
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			for i := range tc.ipClaims {
+				capm3IPClaim := &ipamv1.IPClaim{}
+				claimNamespacedName := types.NamespacedName{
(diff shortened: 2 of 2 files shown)


baremetal/metal3data_manager.go
```diff
@@ -1440,6 +1440,7 @@ func fetchM3IPClaim(ctx context.Context, cl client.Client, mLog logr.Logger,
 func (m *DataManager) fetchIPClaimsWithLabels(ctx context.Context, pool string) ([]ipamv1.IPClaim, error) {
 	allIPClaims := ipamv1.IPClaimList{}
 	opts := []client.ListOption{
+		&client.ListOptions{Namespace: m.Data.Namespace},
```
lentzi90, line 1443: We need to ~~double~~ triple check that this is OK in all cases! Is there any case where we need to fetch IPClaims from other namespaces than the M3D?

kashifest: True, rather do we have a usecase where we create IPclaims for an M3D in separate namespaces

lentzi90:
/hold
We need to check if it is OK to only fetch the IPClaims in the same namespace as the Metal3Data!
I also want to improve the test a bit (it is basically duplicating another test with some small changes...)

baremetal/metal3data_manager.go
lentzi90: Alright, I have looked through the code a bit more and have some findings.

1. `fetchIPClaimsWithLabels` is only called from `releaseAddressFromM3Pool`
2. I assume the `DataManager` should never release addresses from any other namespaces than the M3D that it is working on at the moment.

Based on that I would say we are quite safe.

lentzi90:
/hold cancel
I'm now happy with the tests (have expanded them a bit). I have also checked the code for potential issues with limiting the listing of IPClaims to the M3D namespace and found no indication that this would be problematic.
I have confirmed locally with a full e2e test that this fixes #1250 (used same steps as in the issue).

lentzi90:
/cc @furkatgofurov7
I guess you are the one who knows this code best

tuminoid:
LGTM and loving the fact that you took the time to create proper tests for it too! :heart:

lentzi90:
> Worth of cherry-picking to release branches too?

Yes!

/cherry-pick release-1.5
/cherry-pick release-1.4

lentzi90:
/test-ubuntu-integration-main
Would it be too much to ask to run the tests? :raised_eyebrow:

baremetal/metal3data_manager.go
furkatgofurov7: Hey, thanks for the PR and sorry for the late reply on this one, I was on 2 weeks PTO. 

> 1. `fetchIPClaimsWithLabels` is only called from `releaseAddressFromM3Pool`
> 2. I assume the `DataManager` should never release addresses from any other namespaces than the M3D that it is working on at the moment.

Correct, M3D namespace is tied to IPClaim namespace in all use cases I could think of at the time of implementing the feature.

lentzi90: Thanks for confirming @furkatgofurov7 !

PR #1254, opened 2023-10-06
🌱 use GLOBIGNORE to ignore hack/tools/go.mod in verify-release.sh

tuminoid:
Using GLOBIGNORE makes the code simpler, and allows us to ignore the hack/tools/go.mod for sed as well, not just in grep. This fixes issue that module versions are reported as incorrect, if hack/tools/go.mod is the one with wrong version, yet the output then does now show the hack/tools/go.mod as culprit.

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -37,8 +37,9 @@ set -eu
 # we are using plenty of subshell pipes, and catch errors elsewhere
 set +o pipefail
 
-# enable support for **/go.mod
+# enable support for **/go.mod, and make it ignore hack/tools/go.mod
 shopt -s globstar
+GLOBIGNORE=./hack/tools/go.mod
 
 # user input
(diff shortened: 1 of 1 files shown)


tuminoid:
/override test-ubuntu-e2e-integration-release-1-4
Save the trees.

kashifest (commented):
/approve

lentzi90 (commented):
/lgtm

tuminoid:
/test build
Prow fails to communicate the success back to GH, it seems ....

tuminoid:
/test build
Maybe now the rate limit is refreshed?

tuminoid:
/override build

Build is stuck on different commit `e5d999` than others `1510c8`, doesn't react to anything when fixing it, and it is not relevant test here since we're only changing a hack script.

tuminoid:
/override test-ubuntu-e2e-integration-release-1-4 
Repushed for Prow issues.

PR #1255, opened 2023-10-09
🐛 IP reuse: Fix M3d/IPClaim deletion bug

lentzi90:
Manual cherry-pick of #1251

When listing the IPClaims, we currently only select by labels, so there is no limit on what namespace the IPClaim belongs to. The labels are based on the name of the IPPool and the name of the Metal3Data. Within the same namespace these will be unique, but not necessarily cluster wide.

This commit limits the list of IPClaims to the same namespace as the Metal3Data. It also adds a few unit tests to check the behavior when we have multiple IPClaims, potentially in separate namespaces and with identical labels.


kashifest (commented):
/approve

PR #1261, opened 2023-10-10
:seedling: Configurable go executable

fhaftmann:
This allows to select a specific go version if multiple go versions are installed on a system.

The assumption that the `go` executable refers to a go version suitable for compiling is not always valid; with having the `go` executable parametrized it is possible to choose a suitable executable from a set of local go installations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,9 @@ SHELL:=/usr/bin/env bash
 .DEFAULT_GOAL:=help
 
 GO_VERSION ?= 1.20.8
+GO := $(shell type -P go)
 # Use GOPROXY environment variable if set
-GOPROXY := $(shell go env GOPROXY)
+GOPROXY := $(shell $(GO) env GOPROXY)
 ifeq ($(GOPROXY),)
 GOPROXY := https://proxy.golang.org
 endif
@@ -41,7 +42,7 @@ TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
 BIN_DIR := bin
 
 # Set --output-base for conversion-gen if we are not within GOPATH
-ifneq ($(abspath $(ROOT_DIR)),$(shell go env GOPATH)/src/github.com/metal3-io/cluster-api-provider-metal3)
+ifneq ($(abspath $(ROOT_DIR)),$(shell $(GO) env GOPATH)/src/github.com/metal3-io/cluster-api-provider-metal3)
 	CONVERSION_GEN_OUTPUT_BASE := --output-base=$(ROOT_DIR)
 endif
 
@@ -60,7 +61,7 @@ SETUP_ENVTEST = $(TOOLS_BIN_DIR)/setup-envtest
 GINKGO := "$(ROOT_DIR)/$(TOOLS_BIN_DIR)/ginkgo"
 
 # Helper function to get dependency version from go.mod
-get_go_version = $(shell go list -m $1 | awk '{print $$2}')
+get_go_version = $(shell $(GO) list -m $1 | awk '{print $$2}')
 GINGKO_VER := $(call get_go_version,github.com/onsi/ginkgo/v2)
 ENVTEST_K8S_VERSION := 1.28.x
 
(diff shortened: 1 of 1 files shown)


tuminoid:
Hi @fhaftmann, thanks for the PR!

`type -P go` gives you path of first executable `go` in your `$PATH`, which is the same binary executed when you run `go`.  Which cases/configurations does this PR hence fix?

lentzi90 (commented):
/approve

tuminoid (commented):
We have this funny `GO_VERSION` on L23, that doesn't match up anymore if you change the GO version, but OTOH it doesn't react if your default Go is something else either. So not an issue created by this PR.

LGTM, but I'll let @mboukhalfa  or @kashifest take a look and LGTM. Thank you for the contribution!

kashifest:
I am trying to understand the motivation behind this , even go recommends to remove any prior go installation to have a fresh go tree. Multiple versions also means reproducibility and verification is difficult.

lentzi90:
> I am trying to understand the motivation behind this , even go recommends to remove any prior go installation to have a fresh go tree. Multiple versions also means reproducibility and verification is difficult.

I guess you refer to [this](https://go.dev/doc/install)? It is not really concerned about multiple go versions though. That page describes the "main" go installation, and that should only be one. However, it is perfectly fine to install more versions, and you can do it using go itself. Check this: https://go.dev/doc/manage-install

kashifest:
> I guess you refer to [this](https://go.dev/doc/install)? It is not really concerned about multiple go versions though. That page describes the "main" go installation, and that should only be one. However, it is perfectly fine to install more versions, and you can do it using go itself. Check this: https://go.dev/doc/manage-install

Yes and IMO in upstream code we should be only concerned about the main go installation. Users are ofcourse allowed to use multiple go versions locally, but configurable go versions upstream might give wrong signals IMO.

lentzi90:
> Yes and IMO in upstream code we should be only concerned about the main go installation. Users are ofcourse allowed to use multiple go versions locally, but configurable go versions upstream might give wrong signals IMO.

What would the use of multiple go versions be in that case, if the main installation is the only one we accept? I think this is a valid request that helps developers work on CAPM3 while also working on other projects that may have different go version requirements. It could even be useful when working on CAPM3 release branches.
Forcing developers to reinstall go just to be able to backport a change is not helpful.

fhaftmann:
> > Yes and IMO in upstream code we should be only concerned about the main go installation. Users are ofcourse allowed to use multiple go versions locally, but configurable go versions upstream might give wrong signals IMO.
> 
> What would the use of multiple go versions be in that case, if the main installation is the only one we accept? I think this is a valid request that helps developers work on CAPM3 while also working on other projects that may have different go version requirements. It could even be useful when working on CAPM3 release branches. Forcing developers to reinstall go just to be able to backport a change is not helpful.

That is exactly my intention: I want to compile CAPM3 with the go compiler as required by CAPM3 without fiddling much around. The MR is *not* about compiling different parts of CAPM3 with different go versions, but providing a simple way to choose exactly the required one explicitly.

kashifest:
> What would the use of multiple go versions be in that case, if the main installation is the only one we accept? I think this is a valid request that helps developers work on CAPM3 while also working on other projects that may have different go version requirements. It could even be useful when working on CAPM3 release branches. Forcing developers to reinstall go just to be able to backport a change is not helpful.

I have multiple go in my local env for different release branches of CAPM3, I use either the go1.x.x version to call the binaries or use an alias for go. I don't know how CAPM3 is forcing me to reinstall go without having this particular patch. 

I would let other reviewers decide the fate of this patch.

lentzi90:
> I have multiple go in my local env for different release branches of CAPM3, I use either the go1.x.x version to call the binaries or use an alias for go. 

This patch is allowing you to do that in a simpler way. If you want to use `go1.x.x` you would currently need to modify the Makefile, right? With this patch you would just need to set `GO=go1.x.x` and you can call any make commands like normal.

> I don't know how CAPM3 is forcing me to reinstall go without having this particular patch. 

I'm referring to what you said above: "IMO in upstream code we should be only concerned about the main go installation."
This implies that we should not accept the other versions of go that you may have installed, or in other words, you would have me reinstall go if I want a different version, since the "main" version is all we should care about.

fhaftmann:
> > I have multiple go in my local env for different release branches of CAPM3, I use either the go1.x.x version to call the binaries or use an alias for go.
> 
> This patch is allowing you to do that in a simpler way. If you want to use `go1.x.x` you would currently need to modify the Makefile, right? With this patch you would just need to set `GO=go1.x.x` and you can call any make commands like normal.

To be more explicit – with this patch, giving an *arbitrary* local go installation (e.g. simply the one coming with your distribution), you can bootstrap the desired go version simply using

```sh
GOBIN="$(pwd)/bin" go install golang.org/dl/go1.20.8@latest
bin/go1.20.8 download 
```

without any impact on your development system as such and then continue e. g.

```sh
make GO="$(pwd)/bin/go1.20.8" build 
```

As others have pointed out, AFAIU there is no way to accomplish this currently without manipulating you local development system or ad-hoc patching the Makefile.

kashifest:
> This implies that we should not accept the other versions of go that you may have installed, or in other words, you would have me reinstall go if I want a different version, since the "main" version is all we should care about.

 See my next line. " Users are ofcourse allowed to use multiple go versions locally" . 

> This patch is allowing you to do that in a simpler way. If you want to use `go1.x.x` you would currently need to modify the Makefile, right? With this patch you would just need to set `GO=go1.x.x` and you can call any make commands like normal.

Agree

>  I fail to see how this patch would break anything or even change anyone's wortkflow unless they want to change it by using it. Hence I'd be interested to see specific cases where this patch would have negative impact.

I didn't say it would break anything or would have negative impact. I asked about the necessity of it living in upstream code.  

> To be more explicit – with this patch, giving an _arbitrary_ local go installation (e.g. simply the one coming with your distribution), you can bootstrap the desired go version simply using
> 
> ```shell
> GOBIN="$(pwd)/bin" go install golang.org/dl/go1.20.8@latest
> bin/go1.20.8 download 
> ```
 
> without any impact on your development system as such and then continue e. g.
> 
> ```shell
> make GO="$(pwd)/bin/go1.20.8" build 
> ```
> As others have pointed out, AFAIU there is no way to accomplish this currently without manipulating you local development system or ad-hoc patching the Makefile.

Thanks. I can agree to this that it would make life simpler. 

/lgtm

fhaftmann:
Despite lgtm this is still on hold. Are there any objections to withdraw hold?

lentzi90:
Seems like we have reached consensus.
/hold cancel

PR #708, opened 2022-09-06
✨ Add bond hash policy

shibaPuppy:
https://github.com/openstack/ironic/blob/master/ironic/tests/unit/drivers/modules/network/json_samples/network_data.json#L31
Adds the bond hash policy feature added to Ironic.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1262, opened 2023-10-10
:sparkles: Enable mac address retrieved from annotations

### LINKED ISSUE #1208: Network devices specified in Metal3DataTemplate spec.networkData can be identified by an annotation of a machine object
**User Story**

As an operator I would like to use *spec.networkData* in *Metal3DataTemplate* to configure a machine’s network with network interfaces identified by a MAC address provided by an annotation of a machine object, similar as already possible using *spec.metaData.fromAnnotations*, to provide distinct MAC addresses for machines sharing the same *Metal3DataTemplate*.

**Detailed Description**

It is already possible to specify

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: Metal3DataTemplate
metadata:
  name: my-datatemplate
spec:
  clusterName: my-cluster
  metaData:
    fromAnnotations:
    - key: device_macaddress
      object: baremetalhost
      annotation: device_macaddress
```

to construct meta data from annotations (in this example, from *baremetalhost*s).

This should also carry over to sth. like

```yaml
kind: Metal3DataTemplate
metadata:
  name: my-datatemplate
spec:
  clusterName: my-cluster
  networkData:
    links:
      ethernets:
        - type: phy
           id: device
           macAddress:
             fromAnnotations:
               - key: device_macaddress
                 object: baremetalhost
                 annotation: device_macaddress
```

It seems feasible to use the existing implementation for *metaData* for *networkData* also.

Currently, *macAddress* can either be
a) a static string
b) derived from a network interface name (after inspection)

Both seem too inflexible in general:
a) seems only be reasonable if *my-datatemplate* refers to exactly one machine (which is a degenerate special case)
b) seems only be feasible if *my-datatemplate* refers to a bunch of machines which are similar enough to share the same schema for network device names.

/kind feature

fhaftmann:
This PR in principle enables users to assign mac addresses via networkData based on annotations to objects representing machines, in a fashion similiar what is already possible for metaData.

**Questions I was unable to resolve on my own**:

These issues should be resolved before merging:
* The data schema is extended conservatively, it is not clear to me whether this also requires an API version bump (e.g. to *v1beta2*).
* The fuzzy round trip conversion test complains about `FromAnnotation` on `v1beta1.NetworkLinkEthernetMac` going lost; I have found no clue how to instruct it that this is a genuinely new field with no representation in *v1alpha5*, although other schema extensions surely handled this in a similar way.

The follwing issues are also present on branch *main* (commit 3068f09caabb19e4b5b9b07e0ba091d867f8d22e), it is unclear what to do about these:
* `make lint` fails with
```
baremetal/remote/remote_test.go:36:4: validKubeConfig declared and not used (typecheck)
                        validKubeConfig              string
                        ^
controllers/metal3remediation_controller.go:105:10: r.Get undefined (type *Metal3RemediationReconciler has no field or method Get) (typecheck)
        err = r.Get(ctx, key, &metal3Machine)
```
* `make verify` complains about a bunch of wrong boilerplate headers and aborts.

Fixes #1208


tuminoid:
Hi @fhaftmann, thanks for the PR!

I don't get the failures with `make lint` or `make verify` as you do, but I recall I hit that linting issue earlier, and it was some sort of unrelated flake. While this PR is on top of fresh main, I think you might have something cached that'll cause it. My local `make lint` produces same errors as the golangci-lint test in this PR: https://github.com/metal3-io/cluster-api-provider-metal3/actions/runs/6469401835/job/17564710755?pr=1262 and not the ones you have in the description. 

Please try cleaning the repo (`git clean -dfx` for example) and linter cache `rm -rf ~/.cache/golangci-lint` and see if it helps.

fhaftmann:
> Please try cleaning the repo (`git clean -dfx` for example) and linter cache `rm -rf ~/.cache/golangci-lint` and see if it helps.

That did not help, but the feedback from the CI was enough to address the reported issues.

The failing unit test in https://prow.apps.test.metal3.io/view/s3/prow-logs/pr-logs/pull/metal3-io_cluster-api-provider-metal3/1262/unit/1711731905075875840 presumably is the failing round-trip conversion test mentioned in the PR where I am at a loss how to address it.

tuminoid:
> > Please try cleaning the repo (`git clean -dfx` for example) and linter cache `rm -rf ~/.cache/golangci-lint` and see if it helps.
> 
> That did not help, but the feedback from the CI was enough to address the reported issues.
> 
> The failing unit test in https://prow.apps.test.metal3.io/view/s3/prow-logs/pr-logs/pull/metal3-io_cluster-api-provider-metal3/1262/unit/1711731905075875840 presumably is the failing round-trip conversion test mentioned in the PR where I am at a loss how to address it.

/cc @mboukhalfa @lentzi90

lentzi90:
> make verify complains about a bunch of wrong boilerplate headers and aborts.

Please ignore this for now. We have not been running this in the CI for a long time. At some point we may want to, but until then you can safely ignore it.

Regarding `make lint`, could this be due to the go version? I saw #1261 so I'm guessing you are used to running with many different versions.
You can also try doing the same as the CI, i.e. run `hack/golint.sh` (set `CONTAINER_RUNTIME` to either `docker` or `podman` depending on what you have). That should give you the exact same behavior as seen in the CI.

lentzi90:
For the conversion, I think you are missing what we do [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/07d7eb67b6e70c43c596a19a821423e3522464d9/api/v1alpha5/conversion.go#L169-L211). The down conversion (v1beta1 -> v1alpha5) is good, but on upconversion to v1beta1 we need to restore the field from a "backup" annotation to get the roundtrip clean.
There is a recent good example in #708 that will hopefully help. :slightly_smiling_face:

fhaftmann:
> For the conversion, I think you are missing what we do [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/07d7eb67b6e70c43c596a19a821423e3522464d9/api/v1alpha5/conversion.go#L169-L211). The down conversion (v1beta1 -> v1alpha5) is good, but on upconversion to v1beta1 we need to restore the field from a "backup" annotation to get the roundtrip clean. There is a recent good example in #708 that will hopefully help. 🙂

Thanks, that was the missing hint.

fhaftmann:
> Regarding `make lint`, could this be due to the go version? I saw #1261 so I'm guessing you are used to running with many different versions. Alternatively the `golangci-lint` version. It is downloaded to `hack/tools/bin`. Deleting it from there should make sure it downloads the correct version the next time you run `make lint`.

Tried it, still same behaviour. Same behaviour without my changes. I am unsure whether this should be pursued further.

config/default/capm3/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/cluster-api-provider-metal3:main
+      - image: quay.io/metal3-io/cluster-api-provider-metal3-amd64:v1beta1
```
lentzi90, line 11: Remember to remote before merge!

baremetal/metal3data_manager.go
```diff
@@ -1373,6 +1377,32 @@ func getBMHMacByName(name string, bmh *bmov1alpha1.BareMetalHost) (string, error
...
+	case capimachine:
+		return machine.Annotations[annotation], nil
+	case host:
+		return bmh.Annotations[annotation], nil
+	default:
+		return "", errors.New("Unknown object type")
+	}
+}
+
+func getValueFromAnnotationStrict(object string, annotation string,
```
lentzi90, line 1394: I think we should just conclude that this is how we should handle annotations and get rid of `getValueFromAnnotation` (used when rendering metadata). Different behaviors for metadata and network data will just be confusing.

lentzi90 (commented):
Really great work on this! I especially appreciate the unit tests and docs!
Just two notes below

baremetal/metal3data_manager.go
lentzi90: /cc @kashifest

config/default/capm3/manager_image_patch.yaml
fhaftmann: Oops, I have often seen that side-effect of building but did not recognize it after issuing code generation. Corrected.

fhaftmann: But maybe this was not the purpose of your comment.

lentzi90: Yeah it is unfortunate that the current way of deploying and testing relies on making changes to these manifests :grimacing: 
I just wanted to make sure we remember to remove any changes to this line. Looks like you now accidentally removed the `:main` at the end.

baremetal/metal3data_manager.go
fhaftmann: During developing I found out that `getValueFromAnnotation` yields empty strings for non-existing annotations. (It is not clear to me whether Kubernetes itself distinguishes between empty or non-existing annotations.)

An empty MAC address is surely an error condition, and hence I decided to make that strict. (You could even argue to check the annotation further e.g. via a regular expression to allow early detection of errors.)

But for meta data it is perfectly valid to have empty values – we do just not know how the data is applied.

config/default/capm3/manager_image_patch.yaml
fhaftmann: Corrected now, sorry for the noice (again).

fhaftmann:
I am still confused by the golangci-lint / lint CI – in https://github.com/metal3-io/cluster-api-provider-metal3/actions/runs/6533034950/job/17737383459?pr=1262#step:5:12777 it plainly states
```
Running [/home/runner/golangci-lint-1.54.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: unused-parameter: parameter 'm3d' seems to be unused, consider removing or renaming it as _ (revive)
```
Unfortunately without any file / line indication.
I am also still unable to reproduce the lint run locally.
Is there any chance to make the CI explicit about location of issues?

baremetal/metal3data_manager.go
lentzi90: Thanks for the explanation, that makes sense!
Could you add doc strings to these functions with some description that makes it clear what the difference is?

lentzi90:
Here is something you can try. Create a codespace from your branch:
image
Then run `make lint` in there.

This is what I got when doing it:

```
baremetal/metal3data_manager.go:925:24: unused-parameter: parameter 'm3d' seems to be unused, consider removing or renaming it as _ (revive)
func renderNetworkData(m3d *infrav1.Metal3Data, m3dt *infrav1.Metal3DataTemplate,
                       ^
```

fhaftmann:
Thanks, that gave me the necessary information.

baremetal/metal3data_manager.go
fhaftmann: This discussion made be think about that matter again and I have now added an explicit regexp check for the resulting MAC address – regardless of its source. That appears far more conclusive then the rather erratic and not self-explanatory distinction between potentially empty and non-empty annotations.

That means a subtle change of the overall behaviour: non-well-formed MAC addresses will not raise issues later after provisioning, but already before.

baremetal/metal3data_manager.go
```diff
@@ -275,7 +276,7 @@ func (m *DataManager) createSecrets(ctx context.Context) error {
 	// The NetworkData secret must be created
 	if apierrors.IsNotFound(networkDataErr) {
 		m.Log.Info("Creating Networkdata secret")
-		networkData, err := renderNetworkData(m3dt, bmh, poolAddresses)
+		networkData, err := renderNetworkData(m3dt, m3m, capiMachine, bmh, poolAddresses)
```
lentzi90, line 279: nit:
```suggestion
		networkData, err := renderNetworkData(m3dt, m3m, machine, bmh, poolAddresses)
```
for consistency

lentzi90 (commented):
I'm happy with this
/approve

Would love to see some reviews by others though

baremetal/metal3data_manager.go
fhaftmann: The observation is valid, but this inconsistency is not introduced this MR.

There could surely be done something more consistent terminology concerning variable names etc., but this should IMHO go into a separate MR.

For example, a good starting point could be to make the symbolic identifiers

```go
const (
	m3machine = "metal3machine"
	host = "baremetalhost"
	capimachine = "machine"
)
```

less deviant and then make the variable terminology  more consistent.

Rozzii (approved):
/lgtm

PR #1268, opened 2023-10-13
:bug: Fix the "Could not find node with label" error message

mquhuy:
The current format results in error log which looks like this: "could not find node with labelnodelabelmetal3.io/uuid=b8d8272d-36d8-4eb3-a999-90bee3039b16"

Fixes #

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -1332,7 +1332,7 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, providerIDOnM3M 
 	if countNodesWithLabel == 0 {
 		// The node could either be still running cloud-init or have been
 		// deleted manually. TODO: handle a manual deletion case.
-		errMessage := fmt.Sprint("requeuing, could not find node with label", "nodelabel", nodeLabel)
+		errMessage := fmt.Sprintf("requeuing, could not find node with label: %s", nodeLabel)
 		m.Log.Info(errMessage)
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Weird that we log the error when we anyway return it then. I wonder if we should change it so that we just return the error and then let the caller take care of logging it. That would be a more common pattern I would say.
Anyway, this is clearly fixing a bug so
/lgtm

P.S. I can see why it is written like this. It would have been correct if passed to the logger, as in
```go
m.Log.Info("requeuing, could not find node with label", "nodelabel", nodeLabel)
```
But then we would need to repeat it for the return...

mquhuy:
> Weird that we log the error when we anyway return it then. I wonder if we should change it so that we just return the error and then let the caller take care of logging it. That would be a more common pattern I would say. Anyway, this is clearly fixing a bug so /lgtm
> 
> P.S. I can see why it is written like this. It would have been correct if passed to the logger, as in
> 
> ```go
> m.Log.Info("requeuing, could not find node with label", "nodelabel", nodeLabel)
> ```
> 
> But then we would need to repeat it for the return...

Yes. I noticed that the error should have been logged already by the inner function, but for some reason it's logged once again here, as an INFO log. Not sure what the reason was, but I would think as it was logged once, the second log doesn't need so much details.

mquhuy:
cc @Rozzii @smoshiur1237 @mboukhalfa @adilGhaffarDev

smoshiur1237 (approved):
/lgtm

Issue #1310, opened 2023-11-09
Don't assume KubeadmControlPlane is the controlplane provider 


hardys:
**What steps did you take and what happened:**

When testing in the [Sylva community ](https://gitlab.com/sylva-projects/sylva-core/-/issues/705) it was discovered that CAPM3 has a [hard-coded assumption](https://github.com/metal3-io/cluster-api-provider-metal3/blob/73e3789d3a2c58b5ce16aeb1598ded4839ef6dea/baremetal/metal3machine_manager.go#L1663C6-L1663C45) that `KubeadmControlPlane` is the controlplane provider.

It appears that this breaks when using the [nodeReuse feature](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/api.md#enabling-nodereuse-feature) with an alternative controlplane provider

**What did you expect to happen:**

The control plane reference can be any `controlplane.cluster.x-k8s.io` so we should remove the hard-coded assumptions - I expect CAPM3 to work with any control plane provider, not only Kubeadm.

**Environment:**

This was discovered in the Sylva test environment which optionally uses the RKE2 control plane provider, but I think this issue will be triggered when using any provider other than Kubeadm.

/kind bug

PR #1311, opened 2023-11-09
✨ Don't assume KubeadmControlPlane for nodeReuse

hardys:
It's valid to use other `controlplane.cluster.x-k8s.io` controlplane provider resources, so remove the explicit check for Kind

We [already check the Group ](https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/api/v1beta1/groupversion_info.go#L30) e.g `controlplane.cluster.x-k8s.io` which should be sufficient to ensure we're retrieving a controlplane resource name.

Fixes: #1310

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -578,7 +578,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			if m.Cluster.DeletionTimestamp.IsZero() {
 				// Fetch corresponding Metal3MachineTemplate, to see if nodeReuse
 				// feature is enabled. If set to true, check the machine role. In case
-				// machine role is ControlPlane, set nodeReuseLabelName to KubeadmControlPlane
+				// machine role is ControlPlane, set nodeReuseLabelName to ControlPlane
 				// name, otherwise to MachineDeployment name.
 				m.Log.Info("Getting Metal3MachineTemplate")
 				m3mt := &infrav1.Metal3MachineTemplate{}
@@ -609,15 +609,15 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 						}
 						// Check if machine is ControlPlane
 						if m.isControlPlane() {
-							// Fetch KubeadmControlPlane name for controlplane machine
-							m.Log.Info("Fetch KubeadmControlPlane name while deprovisioning host", "host", host.Name)
-							kcpName, err := m.getKubeadmControlPlaneName(ctx)
+							// Fetch ControlPlane name for controlplane machine
+							m.Log.Info("Fetch ControlPlane name while deprovisioning host", "host", host.Name)
+							cpName, err := m.getControlPlaneName(ctx)
 							if err != nil {
 								return err
 							}
-							// Set nodeReuseLabelName on the host to KubeadmControlPlane name
-							m.Log.Info("Setting nodeReuseLabelName in host to fetched KubeadmControlPlane", "host", host.Name, "kubeadmControlPlane", kcpName)
-							host.Labels[nodeReuseLabelName] = kcpName
+							// Set nodeReuseLabelName on the host to ControlPlane name
+							m.Log.Info("Setting nodeReuseLabelName in host to fetched ControlPlane", "host", host.Name, "controlPlane", cpName)
+							host.Labels[nodeReuseLabelName] = cpName
 						} else {
 							// Fetch MachineDeployment name for worker machine
 							m.Log.Info("Fetch MachineDeployment name while deprovisioning host", "host", host.Name)
@@ -949,7 +949,7 @@ func consumerRefMatches(consumer *corev1.ObjectReference, m3machine *infrav1.Met
 	return true
 }
 
-// nodeReuseLabelMatches returns true if nodeReuseLabelName matches KubeadmControlPlane or MachineDeployment name on the host.
+// nodeReuseLabelMatches returns true if nodeReuseLabelName matches ControlPlane or MachineDeployment name on the host.
 func (m *MachineManager) nodeReuseLabelMatches(ctx context.Context, host *bmov1alpha1.BareMetalHost) bool {
 	if host == nil {
 		return false
@@ -958,17 +958,17 @@ func (m *MachineManager) nodeReuseLabelMatches(ctx context.Context, host *bmov1a
 		return false
 	}
 	if m.isControlPlane() {
-		kcp, err := m.getKubeadmControlPlaneName(ctx)
+		cp, err := m.getControlPlaneName(ctx)
 		if err != nil {
 			return false
 		}
 		if host.Labels[nodeReuseLabelName] == "" {
 			return false
 		}
-		if host.Labels[nodeReuseLabelName] != kcp {
+		if host.Labels[nodeReuseLabelName] != cp {
 			return false
 		}
-		m.Log.Info("nodeReuseLabelName on the host matches KubeadmControlPlane name", "host", host.Name, "kubeadmControlPlane", kcp)
+		m.Log.Info("nodeReuseLabelName on the host matches ControlPlane name", "host", host.Name, "controlPlane", cp)
 		return true
 	}
 	md, err := m.getMachineDeploymentName(ctx)
@@ -1650,9 +1650,9 @@ func (m *MachineManager) DissociateM3Metadata(ctx context.Context) error {
 	return deleteObject(ctx, m.client, metal3DataClaim)
 }
 
-// getKubeadmControlPlaneName retrieves the KubeadmControlPlane object corresponding to the CAPI machine.
-func (m *MachineManager) getKubeadmControlPlaneName(_ context.Context) (string, error) {
-	m.Log.Info("Fetching KubeadmControlPlane name")
+// getControlPlaneName retrieves the ControlPlane object corresponding to the CAPI machine.
+func (m *MachineManager) getControlPlaneName(_ context.Context) (string, error) {
+	m.Log.Info("Fetching ControlPlane name")
 	if m.Machine == nil {
 		return "", errors.New("Could not find corresponding machine object")
 	}
@@ -1660,9 +1660,6 @@ func (m *MachineManager) getKubeadmControlPlaneName(_ context.Context) (string, 
 		return "", errors.New("Machine owner reference is not populated")
 	}
 	for _, mOwnerRef := range m.Machine.ObjectMeta.OwnerReferences {
-		if mOwnerRef.Kind != "KubeadmControlPlane" {
-			continue
-		}
 		aGV, err := schema.ParseGroupVersion(mOwnerRef.APIVersion)
 		if err != nil {
 			return "", errors.New("Failed to parse the group and version")
@@ -1670,12 +1667,12 @@ func (m *MachineManager) getKubeadmControlPlaneName(_ context.Context) (string, 
 		if aGV.Group != controlplanev1.GroupVersion.Group {
 			continue
 		}
-		// adding prefix to KubeadmControlPlane name in order to be able to differentiate
-		// KubeadmControlPlane and MachineDeployment in case they have the same names set in the cluster.
-		m.Log.Info("Fetched KubeadmControlPlane name", "kubeadmControlPlane", "kcp-"+mOwnerRef.Name)
-		return "kcp-" + mOwnerRef.Name, nil
+		// adding prefix to ControlPlane name in order to be able to differentiate
+		// ControlPlane and MachineDeployment in case they have the same names set in the cluster.
+		m.Log.Info("Fetched ControlPlane name", "controlPlane", "cp-"+mOwnerRef.Name)
+		return "cp-" + mOwnerRef.Name, nil
 	}
-	return "", errors.New("KubeadmControlPlane name is not found")
+	return "", errors.New("ControlPlane name is not found")
 }
 
 // getMachineDeploymentName retrieves the MachineDeployment object name corresponding to the MachineSet.
@@ -1704,7 +1701,7 @@ func (m *MachineManager) getMachineDeploymentName(ctx context.Context) (string, 
 			continue
 		}
 		// adding prefix to MachineDeployment name in order to be able to differentiate
-		// MachineDeployment and KubeadmControlPlane in case they have the same names set in the cluster.
+		// MachineDeployment and ControlPlane in case they have the same names set in the cluster.
 		m.Log.Info("Fetched MachineDeployment name", "machinedeployment", "md-"+msOwnerRef.Name)
 		return "md-" + msOwnerRef.Name, nil
 	}
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -52,7 +52,7 @@ const (
 	testUserDataSecretName    = "worker-user-data"
 	testMetaDataSecretName    = "worker-metadata"
 	testNetworkDataSecretName = "worker-network-data"
-	kcpName                   = "kcp-pool1"
+	cpName                    = "cp-pool1"
 )
 
 var Bmhuid = types.UID("4d25a2c2-46e4-11ec-81d3-0242ac130003")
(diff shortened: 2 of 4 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -1650,32 +1650,29 @@ func (m *MachineManager) DissociateM3Metadata(ctx context.Context) error {
...
 		if aGV.Group != controlplanev1.GroupVersion.Group {
 			continue
 		}
-		// adding prefix to KubeadmControlPlane name in order to be able to differentiate
-		// KubeadmControlPlane and MachineDeployment in case they have the same names set in the cluster.
-		m.Log.Info("Fetched KubeadmControlPlane name", "kubeadmControlPlane", "kcp-"+mOwnerRef.Name)
+		// adding prefix to ControlPlane name in order to be able to differentiate
+		// ControlPlane and MachineDeployment in case they have the same names set in the cluster.
+		m.Log.Info("Fetched ControlPlane name", "controlPlane", "kcp-"+mOwnerRef.Name)
 		return "kcp-" + mOwnerRef.Name, nil
```
furkatgofurov7, line 1673:
```suggestion
		return "cp-" + mOwnerRef.Name, nil
```

baremetal/metal3machine_manager_test.go
```diff
@@ -4193,21 +4193,23 @@ var _ = Describe("Metal3Machine manager", func() {
...
+							Kind:       "AnyControlPlane",
 							Name:       "test1",
 						},
 					},
 				},
 			},
-			expectError: true,
+			expectError:     false,
+			expectedKcp:     true,
+			expectedKcpName: "kcp-test1",
```
furkatgofurov7, line 4210: Let's not re-use the KCP specific field here? We could introduce new ones in `testCaseGetControlPlaneName` for any other Controlplane provider then KCP and use them here perhaps to differentiate the two OR rename the fields to be generic?

baremetal/metal3machine_manager_test.go
```diff
@@ -4146,15 +4146,15 @@ var _ = Describe("Metal3Machine manager", func() {
 		}),
 	)
 
-	type testCaseGetKubeadmControlPlaneName struct {
+	type testCaseGetControlPlaneName struct {
```
furkatgofurov7, line 4149: Since we generalized it, let's make sure we don't use `Kcp` anywhere throughout the tests?

baremetal/metal3machine_manager.go
```diff
@@ -609,15 +609,15 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 						}
 						// Check if machine is ControlPlane
 						if m.isControlPlane() {
```
furkatgofurov7, line 611: this I double-checked and it uses [this](https://github.com/kubernetes-sigs/cluster-api/blob/e55c6bfad66bb7c39d95037b7334c795e6201b23/util/util.go#L133) and should be fine as is

furkatgofurov7:
@hardys 👋🏼 Thanks for the PR. 

Some context on why only KCP was assumed: that is simply, we only used KCP everywhere as the ControlPlane provider (in all integration tests and other environments) since it is a de-facto control plane implementation used by CAPI and no other cp provider options were taken into the account by then. However, looking into the current proposed changes, I don't see any problems with it (not a breaking change, also probably not a bug 🐛 but rather 🌱 ) and it should give us more freedom of choice. 

There are few other places we need to make changes (replacing `KCP/KubeadmControlPlane` mentions with something generic, i.e `ControlPlane`) to align the docs + e2e tests with the current change:

1. notes in the current repo: https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/api.md#node-reuse-flow
2. the original proposal in the metal3-docs repo: https://github.com/metal3-io/metal3-docs/blob/main/design/cluster-api-provider-metal3/node_reuse.md
3. e2e tests we currently have, checks the feature logic with KCP which should be fine since we had to choose any CP provider to test it and that was KCP in the current e2e tests. I think it should be enough if we clarify that with something along the lines: "any other CP providers should also work in e2e, however, we are using KCP for testing node reuse feature in e2e" somewhere at the the top of node reuse e2e test description in https://github.com/metal3-io/cluster-api-provider-metal3/blob/73e3789d3a2c58b5ce16aeb1598ded4839ef6dea/test/e2e/pivoting_based_feature_test.go#L58

Additionally, let's run e2e tests on the PR to make sure we are not breaking the tests (@mboukhalfa or someone else can you please remind me what test trigger we are using to test the node reuse feature?)

furkatgofurov7 (commented):
@hardys 👋🏼 Thanks for the PR. 

Some context on why only KCP was assumed: that is simply, we only used KCP everywhere as the ControlPlane provider (in all integration tests and other environments) since it is a de-facto control plane implementation used by CAPI and no other cp provider options were taken into the account by then. However, looking into the current proposed changes, I don't see any problems with it (not a breaking change, also probably not a bug 🐛 but rather 🌱 ) and it should give us more freedom of choice. 

There are few other places we need to make changes (replacing `KCP/KubeadmControlPlane` mentions with something generic, i.e `ControlPlane`) to align the docs + e2e tests with the current change:

1. notes in the current repo: https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/api.md#node-reuse-flow
2. the original proposal in the metal3-docs repo: https://github.com/metal3-io/metal3-docs/blob/main/design/cluster-api-provider-metal3/node_reuse.md
3. e2e tests we currently have, checks the feature logic with KCP which should be fine since we had to choose any CP provider to test it and that was KCP in the current e2e tests. I think it should be enough if we clarify that with something along the lines: "any other CP providers should also work in e2e, however, we are using KCP for testing node reuse feature in e2e" somewhere at the the top of node reuse e2e test description in https://github.com/metal3-io/cluster-api-provider-metal3/blob/73e3789d3a2c58b5ce16aeb1598ded4839ef6dea/test/e2e/pivoting_based_feature_test.go#L58

Additionally, let's run e2e tests on the PR to make sure we are not breaking the tests (@mboukhalfa or someone else can you please remind me what test trigger we are using to test the node reuse feature?)

baremetal/metal3machine_manager.go
hardys: Thanks!

baremetal/metal3machine_manager_test.go
hardys: I replaced all the KCP naming to make it generic

baremetal/metal3machine_manager_test.go
hardys: Ack I renamed all-the-things to be more generic :)

baremetal/metal3machine_manager.go
hardys: Thanks - done and tests adjusted to match 👍

hardys:
Thanks for the review @furkatgofurov7 - I think I addressed all your comments 👍

furkatgofurov7 (commented):
Thanks, looking good from my side except a single nit:

furkatgofurov7:
> Done - perhaps in future we can consider if there's any way to extend this coverage, but for now we know there will be downstream testing that covers this scenario (such as in the CI run by the Sylva community mentioned above)

+1 on extending the coverage and adding upstream tests with other CP provider(s).

baremetal/metal3machine_manager.go
```diff
@@ -1650,32 +1650,29 @@ func (m *MachineManager) DissociateM3Metadata(ctx context.Context) error {
...
-		if mOwnerRef.Kind != "KubeadmControlPlane" {
-			continue
-		}
 		aGV, err := schema.ParseGroupVersion(mOwnerRef.APIVersion)
 		if err != nil {
 			return "", errors.New("Failed to parse the group and version")
 		}
 		if aGV.Group != controlplanev1.GroupVersion.Group {
 			continue
 		}
```
furkatgofurov7, line 1669: Looking into a bit more to this, this check should also block fetching CP name other than [KCP and not using v1beta1 APIs](https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/api/v1beta1/groupversion_info.go#L30), since we are referencing `controlplanev1.GroupVersion.Group` while comparing the API version..

hardys: I was wondering about that - this should match any `controlplane.cluster.x-k8s.io` resource I think - I agree that according to the capm3 support matrix it should probably also check the Version, but perhaps we might consider fixing that as a follow-up since it's not strictly related?

Also I note issue which triggered this is related to the RKE2 provider which is [not yet updated to v1beta1](https://github.com/rancher-sandbox/cluster-api-provider-rke2/issues/169) so it may be good if we can ensure that is aligned before enforcing the version here?

docs/api.md
```diff
@@ -535,7 +535,7 @@ When `spec.nodeReuse` field of metal3MachineTemplate is set to `True`, CAPM3
 Machine controller:
 
 - Sets `infrastructure.cluster.x-k8s.io/node-reuse` label to the corresponding
-  CAPI object name (`KubeadmControlPlane` or `MachineDeployment`) on the
+  CAPI object name (a `controlplane.cluster.x-k8s.io` object such as `KubeadmControlPlane` or a `MachineDeployment`) on the
```
tmmorin, line 538:
```suggestion
  type of owner object ("ControlPlane" for a `controlplane.cluster.x-k8s.io` object such as `KubeadmControlPlane` or a `MachineDeployment`) on the
```

tmmorin:
Thank you very much for bringing this fix @hardys :+1: 

As far as I can tell with my limited knowledge of Metal3, it looks good.

docs/api.md
furkatgofurov7, line 538: @tmmorin with ^, we are changing the fundamental logic behind the feature, we are not setting `infrastructure.cluster.x-k8s.io/node-reuse` label to the corresponding `type` of owner object on BMH, but the `name` of that object. For example, if CP name is called `cp-test1`, machine controller will set following label on BMH: `infrastructure.cluster.x-k8s.io/node-reuse: cp-test1`

furkatgofurov7: @tmmorin with ^, we are changing the fundamental logic behind the feature, we are not setting infrastructure.cluster.x-k8s.io/node-reuse label to the corresponding type of owner object on BMH, but the name of that object. For example, if CP name is called cp-test1, machine controller will set following label on BMH: infrastructure.cluster.x-k8s.io/node-reuse: cp-test1

baremetal/metal3machine_manager.go
furkatgofurov7: > I was wondering about that - this should allow any controlplane.cluster.x-k8s.io resource I think - I agree that according to the capm3 support matrix it should probably also check the Version, but perhaps we might consider fixing that as a follow-up since it's not strictly related?

@hardys ah right, it should be fine still if only group matches and not version 👍🏼 

> Also I note issue which triggered this is related to the RKE2 provider which is https://github.com/rancher-sandbox/cluster-api-provider-rke2/issues/169 so it may be good if we can ensure that is aligned before enforcing the version here?

I was commenting on that issue, perhaps you meant to bump API version in the RKE2 provider itself and not the CAPI version? We can discuss the details on that issue itself perhaps 🙂

docs/api.md
tmmorin, line 538: ah yes, indeed

the text could capture the `cp-` or `md-` prefix though

```suggestion
  CAPI object name (a `controlplane.cluster.x-k8s.io` object such as `KubeadmControlPlane` or a `MachineDeployment`), prefixed by `cp-` or`md-` (to distinguish control plane and machine deployments), on the
```

baremetal/metal3machine_manager.go
hardys: @furkatgofurov7 yes I was thinking the CAPI bump would impact the provider API version but it sounds like that's not the case, so provided we're only checking the group I think this should work OK

furkatgofurov7:
> test-centos-e2e-integration-main

looks like it was not triggered at all, I vaguely remember seeing CI problems in the provider channel recently which might be related. Let's try again:

furkatgofurov7:
> Additionally, let's run e2e tests on the PR to make sure we are not breaking them (@mboukhalfa or someone else can you please remind me what test trigger we are using to test the node reuse feature?)

Additionally, it should be:

/test-ubuntu-e2e-feature-main

furkatgofurov7:
>Additionally, it should be:
> test-ubuntu-e2e-feature-main

Looks like it did not trigger the tests at all, let's try again:

/test-ubuntu-e2e-feature-main

furkatgofurov7:
Unrelated to these changes `package update error` https://jenkins.nordix.org/job/metal3_capm3_main_e2e_feature_test_ubuntu/58/execution/node/70/log/

/test-ubuntu-e2e-feature-main

furkatgofurov7:
Alright, it is not a flake actually, but a wider issue discussed upstream, xref: https://github.com/containers/podman/issues/20690

hardys:
> Alright, it is not a flake actually, but a wider issue discussed upstream, xref: [containers/podman#20690](https://github.com/containers/podman/issues/20690)

This issue looks to be closed now so trying again

/test-ubuntu-e2e-feature-main

docs/api.md
lentzi90: That is a good point. It seems we have no protection from name collisions currently. I would suggest a separate PR for this though since it would be more of a change in behavior. We will need to consider upgrade scenarios also if we change it (probably just document that CAPM3 should not be upgraded at the same time as the MD or CP is rolling out).

lentzi90 (commented):
Looks good to me
/approve

furkatgofurov7:
broken pipe issue

/test-ubuntu-e2e-feature-main

docs/api.md
hardys: I think the behaviour in the event of a ControlPlane/MachineDeployment name collision is well defined, the [controlplane is checked first so will take precedence](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3machine_manager.go#L972), we could perhaps just document that to avoid any change in behaviour?

furkatgofurov7:
Mostly:

`client_loop: send disconnect: Broken pipe` failures and it is not reaching the node reuse tests to be run

Sunnatillo:
We have a issue on our CI for long running tests. We are currently working on it. 

Good job @hardys. LGTM

furkatgofurov7:
> We have a issue on our CI for long running tests. We are currently working on the fix.
> 
> Good job @hardys. LGTM

How about parallel tests, I assume they should be shorter than normal e2e feature tests?

/parallel-test-ubuntu-e2e-feature-main

hardys:
> Its a pity that the CI is still struggling with this broken pipe issue for so long. Lets try the luck one more time:

I'm not familiar with the details of that issue, is there somewhere I can track it and help collaborate on a fix?

Rozzii:
Could I ask you @hardys to please sign off  the commits, we plan to enable DCO enforcement. (https://github.com/apps/dco)

hardys:
> Could I ask you @hardys to please sign off the commits, we plan to enable DCO enforcement. (https://github.com/apps/dco)

Sure, rebased and amended 👍

kashifest:
The ubuntu based feature test was very close to pass https://jenkins.nordix.org/job/metal3_capm3_main_e2e_feature_test_ubuntu/83/ , It received a broken pipe after the node reuse was performed and repivot was done and the cluster was in deleting state. That is the closest it has went, so I think this should be good to go. We can try one last time, but I will anyhow override centos test
/override test-centos-e2e-feature-main
/test-ubuntu-e2e-feature-main

kashifest:
In the last run node-reuse passed https://jenkins.nordix.org/job/metal3_capm3_main_e2e_feature_test_ubuntu/84/, so if we combine the last two runs, we can safely say feature test is passing.

I am not sure if I would call it a bug though, when it was implemented as a feature we assumed we would use KCP only for this feature. So, IMO we can also consider it as an extension to that feature. 

/lgtm
/hold
cc @furkatgofurov7 @hardys
cc @

furkatgofurov7:
Thanks a lot @kashifest 👍🏼 

> I am not sure if I would call it a bug though, when it was implemented as a feature we assumed we would use KCP only for this feature. So, IMO we can also consider it as an extension to that feature.

+1, @hardys what do you think?

hardys:
> Thanks a lot @kashifest 👍🏼
> 
> > I am not sure if I would call it a bug though, when it was implemented as a feature we assumed we would use KCP only for this feature. So, IMO we can also consider it as an extension to that feature.
> 
> +1, @hardys what do you think?

From my perspective an infrastructure provider shouldn't really be coupled to a specific controlplane provider, but I'm certainly fine to retitle this if the consensus is to descibe it as an enhancement 👍

hardys:
/retitle ✨ Don't assume KubeadmControlPlane for nodeReuse

PR #1360, opened 2023-12-14
🌱  Use k8s apimachinery scheme builder instead of controller-runtime scheme builder

kashifest:
This is recommend for the CAPI providers as mentioned the v1.5 to v1.6 migration guide: https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.5-to-v1.6#suggested-changes-for-providers.

The original issue was discussed in https://github.com/kubernetes-sigs/cluster-api/issues/9011. The issue is using controller-runtime based scheme builder causes a lot of unnecessary dependancies added in go modules. Compared to that apimachinery was chosen as a better alternative.

### DIFF
--- a/api/v1beta1/groupversion_info.go
+++ b/api/v1beta1/groupversion_info.go
@@ -23,23 +23,29 @@ limitations under the License.
 package v1beta1
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
 var (
 	// GroupVersion is group version used to register these objects.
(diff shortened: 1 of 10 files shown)


lentzi90 (commented):
/approve

lentzi90:
I see this did not affect the go modules. Should we update/clean them also, or do we still depend on the controller-runtime scheme builder is some place?

smoshiur1237:
> I see this did not affect the go modules. Should we update/clean them also, or do we still depend on the controller-runtime scheme builder is some place?

As it is not affecting the go modules and test with gomod is passing. We can take this in.

smoshiur1237 (commented):
/lgtm

kashifest:
> I see this did not affect the go modules. Should we update/clean them also, or do we still depend on the controller-runtime scheme builder is some place?

Yes, I wondered the same. My understanding is we didnt directly imported any module named `sigs.k8s.io/controller-runtime/pkg/scheme` which we removed here, rather we use `sigs.k8s.io/controller-runtime ` module and that is something which is still used so no change in go module.

PR #1361, opened 2023-12-14
🐛 Fix race condition getting IPAddress for Metal3Data

### LINKED ISSUE #1354: IPClaims missing for Machines

It sometimes happens that a Metal3Machine appears without IPClaim after a rolling upgrade of a MachineDeployment.

**Update**

I think we have pinpointed the issue!

Code path:

- [getAddressesFromPool](https://github.com/metal3-io/cluster-api-provider-metal3/blob/24e63adf969bcaf46044d66698f882bcb43bc3ac/baremetal/metal3data_manager.go#L337)
  - [ensureM3IPClaim](https://github.com/metal3-io/cluster-api-provider-metal3/blob/24e63adf969bcaf46044d66698f882bcb43bc3ac/baremetal/metal3data_manager.go#L625)
    - [fetchM3IPClaim](https://github.com/metal3-io/cluster-api-provider-metal3/blob/24e63adf969bcaf46044d66698f882bcb43bc3ac/baremetal/metal3data_manager.go#L1448) checks only based on name!
  - [addressFromM3Claim](https://github.com/metal3-io/cluster-api-provider-metal3/blob/24e63adf969bcaf46044d66698f882bcb43bc3ac/baremetal/metal3data_manager.go#L707) checks deletionTimestamp, and if there is deletionTimestamp it also checks the ownerRef

If the new Metal3Data is created before the old IPClaim, it could take the old IPClaim and render the secret before the old IPClaim is deleted.

Reproduction criteria:

- The IPClaim name must be the same before and after rollout (this can happen in at least two cases) so that the controller finds it when looking for the new IPClaim.
  - If BMH name based pre-allocation is enabled, and the same BMH happens to be used for the old and new M3M. The IPClaim gets the name `<bmh-name>-<ippool-name>`.
  - If not using BMH name based pre-allocation, and the Metal3Data happens to get the [same index/name](https://github.com/metal3-io/cluster-api-provider-metal3/blob/24e63adf969bcaf46044d66698f882bcb43bc3ac/baremetal/metal3datatemplate_manager.go#L256-L258). The IPClaim gets the name `<m3d-name>-<ippool-name>`.
- IPClaim must not get the deletionTimestamp until after the new Metal3Data has gotten the IPAddress from it

**End of update**

**Second update**

After trying to find out why/if the IPClaim was not deleted before/together with the Metal3Data, I have concluded that [the IPClaim should always be deleted before the finalizers are removed from the Metal3Data](https://github.com/metal3-io/cluster-api-provider-metal3/blob/01e9b5977aa44ed0bd3283e0b5f98723a7d97915/controllers/metal3data_controller.go#L134-L145). So how come we see this issue still?
I think the answer is [cache](https://github.com/metal3-io/cluster-api-provider-metal3/blob/01e9b5977aa44ed0bd3283e0b5f98723a7d97915/main.go#L176-L184). Everything except BMHs, ConfigMaps and Secrets is cached, so we could simply get a cached IPClaim :facepalm: 

**End of second update**

Unfortunately we have not been able to reproduce the issue yet.
Here is what we know so far:

1. During an upgrade, a Metal3Machine comes up without IPClaims (or with some of them missing).
2. It is unclear if the IPClaim was there and then deleted or if it was never there.
3. It is unclear if the Metal3Data and Metal3DataClaim was also missing.
4. Before this we have heard that one Machine was stuck deleting and was forcefully removed, possibly by deleting the Node.
5. BMH name based pre-allocation is enabled but not used.
6. Node reuse is disabled.
7. Disk cleaning is disabled.
8. There are multiple MachineDeployments, but only one is rolled out at a time and using scale-in with max 1 unavailable.
9. There are many IPPools, about 6 addresses per Machine, but only used for metadata (no networkdata).
10. A new IPPool was added as part of the upgrade.
11. Each MachineDeployment has its own Metal3MachineTemplate and Metal3DataTemplate.
12. New Metal3MachineTemplates and Metal3DataTemplates are created and used when rolling out (needed because of the new IPPool).
13. The Metal3DataTemplates all use `templateReference`. The reference is unique per MachineDeployment and kept the same when rolling to a new Metal3DataTemplate.

All Metal3Machines that are configured to use IPClaims should have the proper IPClaims created.
Otherwise there is risk of IP clash and all kinds of strange things.

We are still investigating this issue and we are not 100 % sure that there is an issue in CAPM3, but it is our best lead so far.
I'm creating this issue now to keep track of findings and link to potentially relevant known bugs.

**Environment:**

- Cluster-api version: v1.5.3
- Cluster-api-provider-metal3 version: v1.5.2
- IP-address-manager: v1.5.2
- Environment (metal3-dev-env or other): other
- Kubernetes version: (use `kubectl version`): v1.28.3

**Discovered or related issues:**

- #1357
  - It is extremely unclear why/if Metal3DataTemplate `templateReference` is needed at all.
- #1356 - Deleting Metal3Data cascades to delete IPClaims and IPAddresses even when they are in use. (They are owned by the Metal3Data so perhaps expected... The real issue is that the Metal3Data is deletable.)
- #1358 
  - Deleting a Metal3Machine forcefully (removing finalizer) deletes the BareMetalHost(!) because it has an owner reference to the M3M.
  - A Metal3Machine owns both the IPClaim and IPAddress
  - A Metal3Machine owns both the Metal3DataClaim and Metal3Data

/kind bug

lentzi90:

In certain conditions `addressFromM3Claim` can accept an old IPClaim (about to be deleted) for a new Metal3Data. This commit fixes this by checking the UID of the owner reference instead of just trusting the name.

Conditions to trigger the issue during a rollout:

1. The IPClaim name must be the same before and after rollout. This can happen when BMH name based pre-allocations are used and the same BMH is used for the old and new M3M. Alternatively, if the Metal3Data happens to be named the same (i.e. it has the same index).
2. The IPClaim must not get the deletion timestamp until after the new Metal3Data has gotten the IPAddress from it (race). The reason for this is that we already check and avoid using IPClaims that has a deletion timestamp.

Fixes #1354

**NOTE**

I have heard that checking UID can be problematic after `clusterctl move`. This code is checking UID's, is it ok? Do we have any alternatives? The index or BMH name based naming of IPClaims and Metal3Data are very problematic here since they can very easily match for the old and new object.

Putting hold to make sure this is properly considered before merging.
/hold

Edit: Confirmed by checking CAPM3 logs that we do not end up with mismatching UIDs after `clusterctl move`.

### DIFF
--- a/baremetal/metal3data_manager.go
+++ b/baremetal/metal3data_manager.go
@@ -709,14 +709,15 @@ func (m *DataManager) addressFromM3Claim(ctx context.Context, poolRef corev1.Typ
 		return addressFromPool{}, true, errors.New("no claim provided")
 	}
 
-	if !ipClaim.DeletionTimestamp.IsZero() {
-		// Is it "our" ipClaim, or does it belong to an old and deleted Metal3Data with the same name?
-		matchingOwnerRef := false
-		for _, ownerRef := range ipClaim.OwnerReferences {
-			if ownerRef.UID == m.Data.GetUID() {
-				matchingOwnerRef = true
-			}
+	// Is it "our" ipClaim, or does it belong to an old and deleted Metal3Data with the same name?
+	matchingOwnerRef := false
+	for _, ownerRef := range ipClaim.OwnerReferences {
+		if ownerRef.UID == m.Data.GetUID() {
+			matchingOwnerRef = true
 		}
+	}
+
+	if !ipClaim.DeletionTimestamp.IsZero() {
 		if !matchingOwnerRef {
 			// It is not our IPClaim so we should not use it. Attempt to remove finalizer if it is still there.
 			m.Log.Info("Found old IPClaim with deletion timestamp. Attempting to clean up and requeue.", "IPClaim", ipClaim)
@@ -730,6 +731,12 @@ func (m *DataManager) addressFromM3Claim(ctx context.Context, poolRef corev1.Typ
 			return addressFromPool{}, true, nil
 		}
 		m.Log.Info("IPClaim has deletion timestamp but is still in use!", "IPClaim", ipClaim)
+	} else if !matchingOwnerRef {
+		// It is not our IPClaim, but it does not appear to be deleting either.
+		// This could happen due to misconfiguration (nameclash) or because the IPClaim
+		// just didn't get the deletionTimestamp before the new Metal3Data was created (race condition).
+		m.Log.Info("Found IPClaim with same name but different UID. Requeing and hoping it will go away.", "IPClaim", ipClaim)
+		return addressFromPool{}, true, nil
 	}
 
 	if ipClaim.Status.ErrorMessage != nil {
--- a/baremetal/metal3data_manager_test.go
+++ b/baremetal/metal3data_manager_test.go
@@ -1384,15 +1384,55 @@ var _ = Describe("Metal3Data manager", func() {
 			},
 			expectRequeue: false,
 		}),
+		Entry("Old IPClaim (wrong UID) without deletion timestamp", testCaseAddressFromM3Claim{
+			m3d: &infrav1.Metal3Data{
+				ObjectMeta: testObjectMeta(metal3DataName, namespaceName, "abc-def-ghi-jkl"),
+			},
+			poolName: testPoolName,
+			poolRef:  corev1.TypedLocalObjectReference{Name: testPoolName},
+			ipClaim: &ipamv1.IPClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       metal3DataName + "-" + testPoolName,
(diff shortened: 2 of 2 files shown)


lentzi90:
Confirmed in CAPM3 logs from the test that we do not end up with bad UIDs after `clusterctl move` :heavy_check_mark:

kashifest:
/approve
Thanks for fixing this. As a followup, we should redesign this whole architecture and relationships of objects IMO since as we started adding new features and use-cases, things got added on top of one another which has resulted in added complexities in the workflow.

smoshiur1237:
@lentzi90 thanks for the fix and good finding. 
Do we override the integration test ubuntu ? as it is having the CI issue with bmh. 
/lgtm

Issue #971, opened 2023-05-10
Removal of v1alpha5 apiversion


kashifest:
## Motivation

v1alpha5(v0.5.x) apiVersion have been EOL since September 2022 and the equivalent CAPI apiVersion v1alpha4(v0.4.x) has been EOL since April 2022. It is now a good time to plan to remove v1alpha5.

## Proposal
We can follow the following timeline:

- Release v1.5:
   -  Announce removal of apiVersion in v1.6 & v1.7
- Release v1.6:
   - Kubernetes API server will stop serving v1alpha5
      - We will set `served` to `false` for the apiVersion in all our CRDs.
      - Then users can’t read or write with the old apiVersion anymore.
      - The API server will still be able to read and convert the old apiVersion.
      - If necessary, users can easily enable the apiVersion again by reverting `served `back to `true` on the CRD.   
 - Release v1.7:
    - Remove v1alpha5 apiVersion
       - Drop all code related to v1alpha5 (API types, conversions, tests, …)
       - If we get feedback that folks need more time to migrate we can discuss delaying this stage by another minor release to give folks more time.

## Tasks

- [x] Open a PR to document removal plan
- [x] Check with CAPI maintainers if providers also need to implement changes as they have done in https://github.com/kubernetes-sigs/cluster-api/issues/7224 and https://github.com/kubernetes-sigs/cluster-api/pull/7995.

v1.5:

- [x] Mark v1alpha5 apiVersion and Go API types as deprecated
- [x] Release notes should highlight the imminent removal of the apiVersion in v1.6 & v1.7

v1.6:

- [x] Stop serving v1alpha5 via `kubebuilder:unservedversion`
- [x]  Release notes should reflect the changes

v1.7:

- [x] Drop all code related to v1alpha5 (API types, conversion, …)
   - Wait a bit for feedback after v1.6
- [x]  Release notes should reflect the changes

Misc:

- [x] Upgrade tests should be updated to cover v1.1 => v1.latest:
- [x] Document upgrade test coverage

/kind cleanup

Rozzii:
/cc @adilGhaffarDev 
@adilGhaffarDev has commented on https://github.com/kubernetes-sigs/cluster-api/issues/10051 thus linking it with https://github.com/kubernetes-sigs/cluster-api/issues/9688 FYI. I guess this requires a CAPI side fix anyways .

kashifest:
[...]

PR #1362, opened 2023-12-17
 ⚠️ Stop serving v1alpha5 API Versions

kashifest:
Set v1alpha5 API types unserved and deprecated in CRDs

Part of https://github.com/metal3-io/cluster-api-provider-metal3/issues/971

### DIFF
--- a/api/v1alpha5/metal3cluster_types.go
+++ b/api/v1alpha5/metal3cluster_types.go
@@ -84,6 +84,8 @@ type Metal3ClusterStatus struct {
 // +kubebuilder:resource:path=metal3clusters,scope=Namespaced,categories=cluster-api,shortName=m3c;m3cluster;m3clusters;metal3c;metal3cluster
 // +kubebuilder:subresource:status
 // +kubebuilder:object:root=true
+// +kubebuilder:unservedversion
+// +kubebuilder:deprecatedversion
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Time duration since creation of Metal3Cluster"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="metal3Cluster is Ready"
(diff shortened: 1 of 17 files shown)


kashifest:
This PR follows CAPI's PR which had stopped serving v1a4 API https://github.com/kubernetes-sigs/cluster-api/pull/8996. In that PR, CAPI has also dropped e2e upgrade tests from v1a4. For us, I couldn't find a simple way to drop it since they test seemed a bit convoluted to me because of parameterisation. That can be done as a followup after discussion.

lentzi90 (commented):
/approve

Just noting for the future, that it would be nicer to set the deprecation notice at least one release before stop serving. I.e. we could have set deprecation in v1.5.

tuminoid:
> /approve
> 
> Just noting for the future, that it would be nicer to set the deprecation notice at least one release before stop serving. I.e. we could have set deprecation in v1.5.

We have deprecation notice in 1.5.0: https://github.com/metal3-io/cluster-api-provider-metal3/releases/tag/v1.5.0

tuminoid (commented):
LGTM.

/cc @mboukhalfa

lentzi90:
> We have deprecation notice in 1.5.0: https://github.com/metal3-io/cluster-api-provider-metal3/releases/tag/v1.5.0

Good! I was more thinking about the `// +kubebuilder:deprecatedversion` comments though. They would mark the API as deprecated also in Kubernetes and pop warnings when you use it.

tuminoid:
> > We have deprecation notice in 1.5.0: https://github.com/metal3-io/cluster-api-provider-metal3/releases/tag/v1.5.0
> 
> Good! I was more thinking about the `// +kubebuilder:deprecatedversion` comments though. They would mark the API as deprecated also in Kubernetes and pop warnings when you use it.

Oh yes, that would be nice in the future. The deprecation notice in release notes is not visible in the repo at all, only in the notes if you know to look, so it would also provide that information.

tuminoid (commented):
/lgtm

PR #1403, opened 2024-01-16
🌱 bump ginkgo to v2.14.0

tuminoid:
Dependabot tries to bump ginkgo in #1401, #1402, but separately, which causes our Makefile "e2e-tests" go haywire. Ginkgo is actually missing from hack/tools/go.mod, and when "go get ginkgo..." is run, it dirties the go.sum/go.mod and the next build for kustomize explode.

Thus, add ginkgo to hack/tools/go.mod as well.


lentzi90 (commented):
/lgtm

PR #1411, opened 2024-01-17
🌱 Bump envtest to v1.29.x

kashifest:
as title says


tuminoid:
This is bumping envtest and not envsubst, right?

kashifest:
> This is bumping envtest and not envsubst, right?

Ah right, envtest

lentzi90 (commented):
/approve

PR #1432, opened 2024-01-29
🐛 Default k8s version in ensure-kubectl script

kashifest:
If we don't default to the kubernetes version used for the test, it will always use v1.29.0 and it will error out for upgrade tests where we use older k8s version to begin with.


smoshiur1237 (approved):
/lgtm

tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

PR #1436, opened 2024-01-30
🌱 Fix Permission Issue for process.txt in Tiltfile Configuration

maxrantil:
This PR fixes a setup issue:  `start.sh: line 34: can't create process.txt: Permission denied`. Modifications in the Tiltfile ensure process.txt is now correctly created with full permissions (0777).

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -133,7 +133,8 @@ FROM golang:1.20 as tilt-helper
 # Support live reloading with Tilt
 RUN wget --output-document /restart.sh --quiet https://raw.githubusercontent.com/windmilleng/rerun-process-wrapper/master/restart.sh  && \
     wget --output-document /start.sh --quiet https://raw.githubusercontent.com/windmilleng/rerun-process-wrapper/master/start.sh && \
-    chmod +x /start.sh && chmod +x /restart.sh
+    chmod +x /start.sh && chmod +x /restart.sh && \
+    touch /process.txt && chmod 0777 /process.txt
 """
 
 tilt_dockerfile_header = """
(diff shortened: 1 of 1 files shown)


tuminoid:
Q: non-root as in host side, or container side? For container side non-root we remove the securityContexts for Tilt, so it runs as root for Tilt development.

mquhuy:
Could you explain the commit message a little bit more? I'm having a hard time matching the description there with what gets changed.

smoshiur1237:
I am also having difficulties to understand the commit message and the changes. Also It seems to run as root.

maxrantil:
I have now updated the Title, commit message and explanation. I totally understand your confusion and the lack of explanation.

@mquhuy & @smoshiur1237 thanks for pointing that out. It was a mistake from my side, an earlier commit message was hanging.

@tuminoid  I'm not sure if it is container-side or host. I have notice what you mentioned about bypassing the securityContexts and that is something I will try to change next, so we are more similar to the CAPI setup. I got this error when trying to start up tilt: `/start.sh: line 34: can't create process.txt: Permission denied`
I then looked into CAPI and found out that this was the solution they used.

tuminoid:
> @tuminoid I'm not sure if it is container-side or host. I have notice what you mentioned about bypassing the securityContexts and that is something I will try to change next, so we are more similar to the CAPI setup. I got this error when trying to start up tilt: `/start.sh: line 34: can't create process.txt: Permission denied` I then looked into CAPI and found out that this was the solution they used.

OK, so the issue is that the `process.txt` does not pre-exist, and it cannot create it, but if it does exist, then it works. Sounds really weird, since we are running as `root` in the container. CAPI is doing the same `securityContext` removal, but in just different place in their Tilt setup. That said, there is no harm adding this.

tuminoid:
/override test-centos-e2e-integration-main test-ubuntu-integration-main

tuminoid (commented):
/lgtm

smoshiur1237 (commented):
LGTM

kashifest:
Not sure if I understood what we are trying to do here but lets see if it helps.
/approve

PR #1443, opened 2024-02-07
🌱 Enable more linters for golangci-lint

kashifest:
This PR enables a few more linters for golangci-lint, namely: 
  - asasalint
  - bidichk
  - containedctx
  - dupword
  - durationcheck
  - errchkjson
  - gci
  - ginkgolinter
  - goprintffuncname
  - loggercheck
  - nosprintfhostport
  - unparam
  - usestdlibvars


baremetal/metal3cluster_manager_test.go
```diff
@@ -378,7 +370,7 @@ var _ = Describe("Metal3Cluster manager", func() {
 	)
 })
 
-func newBMClusterSetup(tc testCaseBMClusterManager) (*ClusterManager, error) {
```
mquhuy, line 381: Is there any linter-related issue for this change? IMO returning an error is the better practice.

test/e2e/node_reuse.go
```diff
@@ -537,7 +537,10 @@ func deleteTaint(taints []corev1.Taint, taintsToDelete []corev1.Taint) ([]corev1
 	return newTaints, deleted
 }
 
-func createNewM3machineTemplate(ctx context.Context, namespace string, newM3machineTemplateName string, m3machineTemplateName string, clusterClient client.Client, imageURL string, imageChecksum string, checksumType string, imageFormat string) {
+func createNewM3machineTemplate(ctx context.Context, namespace string, newM3machineTemplateName string, m3machineTemplateName string, clusterClient client.Client, imageURL string, imageChecksum string) {
```
mquhuy, line 540: Since we're modifying this function anyway, let's change the name to a proper camelCase: `createNewM3MachineTemplate`.

mquhuy (commented):
A couple of questions/suggestions. Otherwise lgtm

baremetal/metal3cluster_manager_test.go
kashifest: Yes linter complains that this is always returning nill as an error, I agree its a good practice but the current code is not returning anything else. Hence I removed it.

test/e2e/node_reuse.go
tuminoid: If function name, then variable names too? :)

maxrantil:
LGTM

controllers/metal3cluster_controller.go
```diff
@@ -151,7 +150,7 @@ func patchMetal3Cluster(ctx context.Context, patchHelper *patch.Helper, metal3Cl
 	return patchHelper.Patch(ctx, metal3Cluster, options...)
 }
 
-func reconcileNormal(ctx context.Context, clusterMgr baremetal.ClusterManagerInterface) (ctrl.Result, error) {
+func reconcileNormal(ctx context.Context, clusterMgr baremetal.ClusterManagerInterface) (ctrl.Result, error) { //nolint:unparam
```
lentzi90, line 153: Is this because the ctrl.Result is always empty?

lentzi90 (commented):
/approve

controllers/metal3cluster_controller.go
kashifest: Precisely, I was tempted to remove it in the first glance, then thought otherwise since this is reconcile function, its better leave option for future user cases.

PR #1468, opened 2024-02-23
🌱 Bump CAPI to v1.5.6

adilGhaffarDev:
as title
Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ need to be manually installed. Example flow of installing Metal3 provider:
    level of the logging verbose with a positive integer number, ie. -v5.
 
    ```shell
-   clusterctl init --core cluster-api:v1.5.5 --bootstrap kubeadm:v1.5.5 \
-       --control-plane kubeadm:v1.5.5 -v5
+   clusterctl init --core cluster-api:v1.5.6 --bootstrap kubeadm:v1.5.6 \
+       --control-plane kubeadm:v1.5.6 -v5
    ```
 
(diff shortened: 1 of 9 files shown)


kashifest:
> Now this is failing because for release-1.5 PR we export the `CAPM3BRANCH` var as the PR commit hash, hence [this check ](https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L267) doesnt hold. Perhaps we should use `CAPM3RELEASEBRANCH` instead in dev-env.

The fix has landed and the new test moved forward. Now the failure is more related to your change, you have bumped `envsubst` to use v2 but somewhere (probably in Makefile) `envsubst` (v1) is still expected, you might need to change that as well. Here is the error 

```
rm -f hack/tools/bin/envsubst*
11:31:21  cd hack/tools && go build -tags=tools -o bin/envsubst-drone github.com/drone/envsubst/cmd/envsubst
11:31:21  no required module provides package github.com/drone/envsubst/cmd/envsubst; to add it:
11:31:21  	go get github.com/drone/envsubst/cmd/envsubst
11:31:21  make[1]: *** [Makefile:259: hack/tools/bin/envsubst-drone] Error 1
```

adilGhaffarDev:
> The fix has landed and the new test moved forward. Now the failure is more related to your change, you have bumped `envsubst` to use v2 but somewhere (probably in Makefile) `envsubst` (v1) is still expected, you might need to change that as well. Here is the error

Fixing it.

PR #1475, opened 2024-02-27
🌱 Bump CAPI to v1.6.2

adilGhaffarDev:
as title
Fixes #


hack/tools/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/drone/envsubst v1.0.3
 	github.com/golang/mock v1.6.0
 	github.com/jteeuwen/go-bindata v3.0.7+incompatible
-	k8s.io/code-generator v0.29.0
```
adilGhaffarDev, line 9: I am unsure why this was uplifted before. `v0.29.0 `requires `go 1.21` and with `k8s.io/code-generator v0.29.0` `go mod tidy` was changing the go version in hack tools to `1.21` and also adding `toolchains` . To make the go version consistent in all go.mods of the project I am changing it back to `k8s.io/code-generator v0.28.1`

kashifest: Interesting, that shouldn't have worked in the first place as you are explaining,  since it is already v0.29.0 in main branch without complains, anyhow keeping it to v0.28.1 is fine also.

kashifest (commented):
/approve

lentzi90 (commented):
/lgtm

PR #1479, opened 2024-02-29
🐛 EnsureImage download image fix

adilGhaffarDev:
When EnsureImage fails to download an image instead of giving the right error that the `image is not available` it downloads a failed page(https://artifactory.nordix.org/ui/native/metal3/images/k8s_v1.28.0/), and in provisioning ironic tries to use it.  This PR is fixing this by erroring out the right message and also failing early instead of trying to provision with non existent image.

Fixes #

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -22,6 +22,7 @@ import (
 	ipamv1 "github.com/metal3-io/ip-address-manager/api/v1alpha1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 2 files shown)


test/e2e/common.go
```diff
@@ -181,6 +182,9 @@ func DownloadFile(filePath string, url string) error {
 	if err != nil {
 		return err
 	}
+	if resp.StatusCode != http.StatusOK {
```
mquhuy, line 185: Shouldn't this block be after `defer`?

mquhuy:
lgtm

lentzi90 (commented):
/approve

test/e2e/common.go
```diff
@@ -181,6 +182,9 @@ func DownloadFile(filePath string, url string) error {
...
+		return errors.Errorf("failed to download image from %s got %d", filePath, resp.StatusCode)
```
tuminoid, line 186:
```suggestion
		return errors.Errorf("failed to download image from %q got %q", filePath, http.StatusText(resp.StatusCode))
```
Maybe quote the outputs, and unpack the status code to proper status text, ie `404` => `404 Not Found` ?

tuminoid (commented):
A suggestion.

test/e2e/common.go
kashifest: Good suggestion @tuminoid , +1

adilGhaffarDev: done

tuminoid (commented):
/lgtm

PR #1481, opened 2024-03-01
🌱 E2E: Remove inspector container

lentzi90:
**What this PR does / why we need it**:

Inspector has been removed in BMO so we can no longer get logs from it.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1487, opened 2024-03-01
🌱  Remove ironic-inspector from container list for e2e tests

kashifest:
Some leftovers were still there even after #1483 and #1481, which is only caught in feature test.


tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

PR #1488, opened 2024-03-01
🌱 try fix GH workflow for image building

tuminoid:
Image build action is not triggering for anything but main, which is the only trigger not having `*` in it. So, let's try removing the quotes as any of the [GH examples](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push)  do not use quotes unless it is mandatory (ie. leading `*`, or `[]`).


smoshiur1237 (commented):
/lgtm

PR #1507, opened 2024-03-08
🌱 Test CAPI in-cluster IPAM with CAPM3

peppi-lotta:
**What this PR does / why we need it**: Test if using CAPI's in-cluster IPAM provider will pass all the test. This will determine if CAMP3 could be able to move to use in-cluster and retire ip-address-manager. 

Fixes #


peppi-lotta:
Doesn't work.Close PR

PR #1606, opened 2024-04-17
:seedling: Use Logf and AnnotateBmh functions from BMO E2E

mquhuy:
This is a continuation of https://github.com/metal3-io/baremetal-operator/pull/1681. Should be edited and merged after the before-mentioned PR is merged.
/hold


test/e2e/common.go
```diff
@@ -60,14 +61,10 @@ func Byf(format string, a ...interface{}) {
 	By(fmt.Sprintf(format, a...))
 }
 
-func Logf(format string, a ...interface{}) {
-	fmt.Fprintf(GinkgoWriter, "INFO: "+format+"\n", a...)
-}
-
 func LogFromFile(logFile string) {
```
adilGhaffarDev, line 64: I am not sure about using BMO e2e `Logf`. `AnnotateBmh` makes sense but I think it's better to keep `Logf` here. This is something that can be different in BMO and CAPM3.

mquhuy: Yes, I think that makes sense, but could you elaborate on what could be the differences? This function is greatly basic, I don't see much space for diversion.

PR #1639, opened 2024-05-02
🌱 Set ARCH based on go env

lentzi90:

We had ARCH hard-coded to amd64. This commit changes it so that we take it from go env instead. That way we should automatically get the proper architecture on different devices.

Fixes #


Sunnatillo (commented):
/lgtm

PR #1641, opened 2024-05-03
🌱 Test: Reproduce parallel test issue

smoshiur1237:
This PR is for testing and giving reference the prow issue coming form parallel feature test run. Where it was giving wrong JOB pass even though test failed in our Jenkins CI.

The goal is to show a filed run in Jenkins where one of the three parallel test failed and at least one job succeed. But PROW is marking this test as success.


smoshiur1237:
> Do we still need this?

we can close this

PR #1624, opened 2024-04-24
🌱 Remove keep test env related vars and configs

kashifest:
These vars are removed from project-infra https://github.com/metal3-io/project-infra/pull/703 , now it is throwing errors form project-infra [PRs ](https://jenkins.nordix.org/view/Metal3/job/metal3-ubuntu-e2e-integration-test-main/63/consoleFull) since we still have traces of these vars here in CAPM3, removing these unused vars.

PR #1405, opened 2024-01-16
:seedling: Support for ClusterClass template

p-strusiewiczsurmacki-mobica:
**What this PR does / why we need it**:

This PR adds support for ClusterClass API as discussed in #1267 

**Which issue(s) this PR fixes**:
Fixes #1267

PR #1671, opened 2024-05-16
:bug: Clusterclass e2e test fix

p-strusiewiczsurmacki-mobica:
This PR is a followup to #1405

It fixes issues with redundant test env related vars and configs that couses the test to fail at start. Those were removed in #1624 and should be also removed for clusterclass.

Also, it tries to fix issue with `{{ ds.meta_data.hostname }}'` reference reported [in the comment here](https://github.com/metal3-io/cluster-api-provider-metal3/pull/1405#discussion_r1584187652).

It seems that,[ according to the cluster-api repository](https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/api/v1beta1/kubeadm_types.go#L231), `NodeRegistration.Name` defaults to hostname, so I believe referencing hostname explicitly here is redundant and minght be deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,13 +225,11 @@ e2e-clusterclass-tests: $(GINKGO) e2e-substitutions clusterclass-templates # Thi
 
 	$(GINKGO) --timeout=$(GINKGO_TIMEOUT) -v --trace --tags=e2e  \
 		--show-node-events --no-color=$(GINKGO_NOCOLOR) \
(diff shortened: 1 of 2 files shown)


kashifest:
/approve
Thanks a lot for the fix @p-strusiewiczsurmacki-mobica

tuminoid:
The link to the centos e2e points to completely different PR.

Anyways, ubuntu passed, 

PR #1708, opened 2024-05-17
🌱 Fix dependabot release branch config to bump only patch versions

kashifest:
With the recent introduction of release branch configuration for dependabot, it is pushing bumps for minor versions of dependancies as well. We don't want minor version bumps for release branches. Setting it to patch versions.
Signed-off-by: Kashif Khan <kashif.khan@est.tech>

PR #1713, opened 2024-05-17
🌱 Group dependabot k8s bumps for release branches and ignore major, minor bumps for all dependancies

kashifest:
Fixing patch version bumps via groups for release branches was probably a bad idea here https://github.com/metal3-io/cluster-api-provider-metal3/pull/1708. It has caused chaos and multiple different dependencies are now being grouped together in one PR. Current fix tries to tackle that by bringing back k8s group and then add an ignore rule for all dependancies in release branch to only allow patch bumps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -107,16 +107,11 @@ updates:
   target-branch: release-1.7
   ## group all dependencies with a k8s.io prefix into a single PR.
   groups:
-    all-go-mod-patch:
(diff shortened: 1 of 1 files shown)


Rozzii (commented):
/lgtm

tuminoid (commented):
/approve

This ignore trick is logical to me, but let's see what dependabot thinks of it.

tuminoid:

PR #1724, opened 2024-05-20
🌱 refactor TLS config

tuminoid:
TLS config code causes security linters to report false positive about TLS versions that can be configured.

This is porting over [CAPO PR 2037](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2037).


tuminoid:
Centos is unstable, Ubuntu + unit passed.

/cc @mboukhalfa @lentzi90

PR #1732, opened 2024-05-24
🌱 Allow minor version bumps for golang.org/x/* dependancies for release branch

kashifest:
golang.org/x/* libs only release minor versions. No patch versions. We need to allow minor version bumps for these libs for release branches for this reason.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -114,6 +114,9 @@ updates:
       patterns: ["k8s.io/*"]
   ignore:
   # Ignore major and minor bumps for release branch
+  # golang.org/x/* only releases minors no patches, so minors have to be allowed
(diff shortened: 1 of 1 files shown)


tuminoid:
Did you find any documentation that which order the config applies, ie. should the x/net be first or last to override the `*` ? If you did, can you link it here for reference?

kashifest:
> Did you find any documentation that which order the config applies, ie. should the x/net be first or last to override the `*` ? If you did, can you link it here for reference?

I didnt unfortunately. I am relying on the fact that the ipam ignore is working even though we have an `*` rule.

tuminoid:

tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #1741, opened 2024-05-24
🌱 remove kubesec workflow

tuminoid:
Kubesec action has not been maintained for 3+ years and is broken. Kubesec itself is maintained, and we'll see if we reintroduce it as prow test instead of action later, or if we have enough coverage from existing tools.


tuminoid:
/override metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main

kashifest (commented):
/approve
/cc @smoshiur1237

Sunnatillo (commented):
/lgtm

PR #1742, opened 2024-05-24
⚠️  Remove Metal3Machine owner reference from BMH

kashifest:
Metal3Machine should not own BMH. It should consume it and release it when not needed. we already put a M3M consumer reference on BMH. This also would mean that for pivoting use cases we must add the clusterctl move labels in the CRDs. Otherwise BMH wont be pivoted to target clusters.

⚠️ Now, BMH object/CRD needs to have The object has the `clusterctl.cluster.x-k8s.io/move` label or the `clusterctl.cluster.x-k8s.io/move-hierarchy ` label to make sure BMH is pivoted to target cluster and removed from the source. `clusterctl.cluster.x-k8s.io/move` and `clusterctl.cluster.x-k8s.io/move-hierarchy` labels could be applied to single objects or at the CRD level (the label applies to all the objects).

We still keep the removal of owner reference code for one minor release cycle to facilitate upgrade scenario where a BMH could still have an owner reference set from previous version. Starting from v1.11.x minor cycle we will remove this code also.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -458,6 +458,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 		if !consumerRefMatches(host.Spec.ConsumerRef, m.Metal3Machine) {
 			m.Log.Info("host already associated with another metal3 machine",
 				"host", host.Name)
+			// The following removal of ownerreference code will be removed in v1.11
 			// Remove the ownerreference to this machine, even if the consumer ref
 			// references another machine.
 			host.OwnerReferences, err = m.DeleteOwnerRef(host.OwnerReferences)
@@ -644,6 +645,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 
 		host.Spec.ConsumerRef = nil
 
+		// The following removal of ownerreference code will be removed in v1.11
 		// Remove the ownerreference to this machine.
 		host.OwnerReferences, err = m.DeleteOwnerRef(host.OwnerReferences)
 		if err != nil {
@@ -1098,13 +1100,6 @@ func (m *MachineManager) setHostConsumerRef(_ context.Context, host *bmov1alpha1
 		Namespace:  m.Metal3Machine.Namespace,
 		APIVersion: m.Metal3Machine.APIVersion,
 	}
-
-	// Set OwnerReferences.
-	hostOwnerReferences, err := m.SetOwnerRef(host.OwnerReferences, true)
-	if err != nil {
-		return err
-	}
-	host.OwnerReferences = hostOwnerReferences
 
 	// Delete nodeReuseLabelName from host.
 	m.Log.Info("Deleting nodeReuseLabelName from host, if any")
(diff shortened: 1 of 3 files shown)


kashifest:
/hold
This is a WIP

shibaPuppy:
@kashifest 
It seems to be PR for the same purpose.
can I close [my PR](https://github.com/metal3-io/cluster-api-provider-metal3/pull/1008)?

kashifest:
> @kashifest It seems to be PR for the same purpose. can I close [my PR](https://github.com/metal3-io/cluster-api-provider-metal3/pull/1008)?

Yes, I am doing the same thing here, I didn't notice you had the same thing in another PR, didnt mean to over step. I am currently testing this since it seems to not passing the pivot scenario so it would need more changes. I am also ok to close my one if you have progressed more.

shibaPuppy:
@kashifest 
i can't spend much time working, so I don't come often.
i will close my PR. :)

kashifest:
/hold
For discussion

lentzi90 (commented):
Thank you for doing this!
How should we handle upgrades? From what I can see, we would not automatically remove owner references that are already there, so it would be up to the user to remove them after upgrade I guess.
One alternative we could consider is to keep the code that removes the owner reference when deleting the M3M for one minor release cycle. Then we can say that users must do a rollout in order to get rid of the owner references, or they can do it manually if they prefer. What do you think?

kashifest:
> Thank you for doing this! How should we handle upgrades? From what I can see, we would not automatically remove owner references that are already there, so it would be up to the user to remove them after upgrade I guess. One alternative we could consider is to keep the code that removes the owner reference when deleting the M3M for one minor release cycle. Then we can say that users must do a rollout in order to get rid of the owner references, or they can do it manually if they prefer. What do you think?

Yes we can do that. I thought of writing a doc update and ask users to remove it manually. We can also keep the removal code for one minor cycle and ask for rollout.

tuminoid:
This will not make CAPM3 1.9, will it? @kashifest

tuminoid:
> This will not make CAPM3 1.9, will it? @kashifest

Discussed offline, this is only going to have a heads-up in 1.9, actual fixes will go to 1.10. Changed milestone accordingly.

tuminoid:
@kashifest can we finalize this and merge? Any breaking changes would be good to get merged early in minor cycle.

kashifest:
> @kashifest can we finalize this and merge? Any breaking changes would be good to get merged early in minor cycle.

Yes good point, I will recheck the PR and update it

kashifest:
/hold cancel
This is good to go in now @lentzi90 @tuminoid. I will push another PR and put it on hold for v1.11 minor release cycle where we get rid of removal of owner reference code also. 

> Yes good point, I will recheck the PR and update it

kashifest:
A followup to be landed in v1.11 release cycle is also created https://github.com/metal3-io/cluster-api-provider-metal3/pull/2238

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #1765, opened 2024-06-13
:bug: Changing errors in pause annotation functions from terminal to transient

adilGhaffarDev:
 Changing errors in pause annotation functions from terminal to transient. When BMH is rebooted there is a chance we get a terminal error from these functions which can cause machines to be in a failed state.


smoshiur1237 (commented):
/lgtm

PR #1800, opened 2024-06-19
🌱Add namespace directory in manifest collection

smoshiur1237:
There is a possibility to miss same kinds of manifest in different namespace. So this PR will add the namespace directory in manifest collection to gather all the manifests which will improve debugging capability.


smoshiur1237:
One test would be enough to check the added directory, so overriding the 2nd test.

tuminoid (commented):
/approve

mboukhalfa (approved):

Rozzii:
Okay, i have got my answer offline, probably in the future we should improve on this in terms of not creating all the "kind" directories for all the namespaces because many of them will be just empty dirs but this PR is a step in the right direction.
/hold cancel
/lgtm

PR #1807, opened 2024-06-26
📖 add openssf best practices badge to README

tuminoid:
We have achieved passing badge on OpenSSF Best Practices. Adding badge to key repositories to display this achievement. It is also checked by the CLOmonitor and is required to move up to incubated project in CNCF.


kashifest (commented):
/approve
/lgtm

PR #1810, opened 2024-06-28
🌱 Fix issues in running local hack scripts

tuminoid:
In CAPM3, there was only issues in codegen. Fix that, plus made it dynamically detect all the config and generated files, so they're included if they'd be generated elsewhere. It is in line with BMO generate.sh now.

For others, simply adding one shellcheck ignore for a optional check many people have enabled about, tired of seeing my editor complain :)

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+# shellcheck disable=SC2292
 
 set -eux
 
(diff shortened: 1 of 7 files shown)


tuminoid:

Rozzii (commented):
/lgtm

PR #1896, opened 2024-08-14
🌱 Silence fetch logs and fetch manifests scripts

Sunnatillo:
Silence fetch logs and fetch manifests scripts as they are too noisy in jenkins logs which makes debugging unpleasant.


tuminoid (commented):
/approve

PR #1997, opened 2024-09-23
:bug: Metal3DataTemplate: requeue if reconcileDelete did not clear finalizer

tmmorin:
**What this PR does / why we need it**:

This MR is a proposal to address #1994.

The reconcileDelete implementation of the Metal3DataTemplate controller needs to requeue if it decided that it can't yet unset the finalizer on the Metal3DataTemplate.  If no requeue is made, there's a scenario where Metal3Data still exists when reconcileDelete is first called, which results in not unsetting the finalize - in that situation, if no change of the Metal3DataTemplate triggers a new reconciliation, the finalizer never gets unset.

To finalize this PR I had to adjust the expectations of the unit test (expect requeue where this initially wasn't expected). 

This made we realize that there is another situation where we want to requeue: the case where `dataTemplateMgr.UpdateDatas` returns an error. The code was relying on checkReconcileError, which (a) can only return requeue results on baremetal.ReconcileError errors (not on other errors), and (b) even for those would not return a requeue for "Terminal" errors, which I is I think only a valid choice for when this function is called for reconcileNormal ; for deletes, it seems that requeues should always be done to avoid the resource remaining stuck on a finalizer.

PR #2000, opened 2024-09-26
:bug: Fix unterminating DataTemplate deletion

### LINKED ISSUE #1994: Metal3DataTemplate deletion stuck, finalizer never cleared
**What steps did you take and what happened**

When deleting all resources for a capm3 cluster a Metal3DataTemplate resource remains pending deletion, its `metal3datatemplate.infrastructure.cluster.x-k8s.io` finalizer never gets cleared.

No resource should have remained.

** TL;DR, me speculating on the cause**

_Disclaimer: my familiarity with capm3 is low, please take what follows as humble hypothesis ;)_

The Metal3DataTemplate controller reconcileDelete does not seem able to take into account that a Metal3DataClaim is not here anymore. More specifically [DataTemplateManager.UpdateDatas](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3datatemplate_manager.go#L174) blindly assumes that all Metal3DataClaims reported to exist by Metal3Data resource, actually exist.

**Context**

In the context of the [Sylva](https://gitlab.com/sylva-projects/sylva-core) project CI pipelines, since we upgraded to capm3 1.8.1, we've been regularly observing a case where a `Metal3DataTemplate` remains stuck pending deletion because its `metal3datatemplate.infrastructure.cluster.x-k8s.io` finalizer is never cleared (for reference: our related [gitlab issue](https://gitlab.com/sylva-projects/sylva-core/-/issues/1632))

The context is a **workload cluster deletion** scenario: we first delete the resources of the cluster,  CAPI Cluster object first, then once the Cluster resource is deleted, we delete all the CAPI resources that our system defines (KubeadmControlPlane, MDs, all the XXX template resources, etc.) and then the namespace.

(I realized when digging this issue that we were deleting the Machine3DataTemplate too early, possibly before the Metal3Machines using them would be deleted, so I suspect that this is possibly contributing to triggering this issue. We're aware that we possibly need to wait and delete resources in a more orderly way...)

I digged into one specific occurrence, and I'm sharing here my observations.

**Observations**

The observations shared are based on the dumps of resources and logs that we take after a CI job failure.

All the data can be found here: https://gitlab.com/sylva-projects/sylva-core/-/jobs/7852801513. The interesting bits are under `management-cluster-dump` which has a `clusterctl-describe.txt`, `Nodes.summary.txt`, `Metal3*` files with resource dumps (the relevant one are in `namespace:  kubeadm-capm3-virt`, and `capm3-system` directory has capm3 controller logs). I'm directly attaching the [Metal3DataTemplate resource dump](https://github.com/user-attachments/files/17078016/Metal3DataTemplates.wcluster.yaml.txt) and the [capm3 controller logs](https://github.com/user-attachments/files/17077957/logs.txt) to this issue.

At the point where the dump is taken we see the following:
* there remains no capm3 resource except our Metal3DataTemplate resource
* extract of the state of this resource:

```yaml
  kind: Metal3DataTemplate
  metadata:
    annotations:
      meta.helm.sh/release-name: cluster
      meta.helm.sh/release-namespace: kubeadm-capm3-virt
    creationTimestamp: "2024-09-18T08:24:11Z"
    deletionGracePeriodSeconds: 0
    deletionTimestamp: "2024-09-18T09:47:51Z"    # <<<<<<<<<<<<<<<<<<<<<<
    finalizers:
    - metal3datatemplate.infrastructure.cluster.x-k8s.io      # <<<<<<<<<
   ...
    name: wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f  # <<<<<<<<<<<<<
  
  managedFields:
...
    - apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
      fieldsType: FieldsV1
      fieldsV1:
        f:status:
          .: {}
          f:indexes:
            .: {}
            f:wc-1458191268-kubeadm-capm3-virt-control-plane-fn667: {}
          f:lastUpdated: {}
      manager: cluster-api-provider-metal3-manager
      operation: Update
      subresource: status
      time: "2024-09-18T09:51:51Z"

# ( this managedFields entry is the most recent one)
  
  status:
    indexes:
      wc-1458191268-kubeadm-capm3-virt-control-plane-fn667: 0  # <<<<<<<<<<<<<<<<<<<
    lastUpdated: "2024-09-18T09:51:51Z"
```
The `status.index` below is interesting, it's related to why the finalizer can't be removed. It designates a Meta3DataClaim resource (wc-1458191268-kubeadm-capm3-virt-control-plane-fn667) that is deleted very early.

* the namespace was deleted at 09:53:32 (see below the Metal3Data seems to be deleted at that point):    `deletionTimestamp: "2024-09-18T09:53:32Z"`

Grepping the logs of capm3 controllers :

```
I0918 09:47:50.684503       1 metal3datatemplate_manager.go:383] "msg"="Deleting Metal3DataClaim" "Metal3DataClaim"="wc-1458191268-kubeadm-capm3-virt-control-plane-8x9vs" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.713494       1 metal3datatemplate_manager.go:421] "msg"="Deleted Metal3Data" "Metal3Data"="wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-2" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.713522       1 metal3datatemplate_manager.go:435] "msg"="Deleted Metal3DataClaim" "Metal3DataClaim"="wc-1458191268-kubeadm-capm3-virt-control-plane-8x9vs" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.758422       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.785723       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.951225       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.960604       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.960713       1 metal3datatemplate_manager.go:383] "msg"="Deleting Metal3DataClaim" "Metal3DataClaim"="wc-1458191268-kubeadm-capm3-virt-control-plane-fn667" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.980052       1 metal3datatemplate_manager.go:421] "msg"="Deleted Metal3Data" "Metal3Data"="wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.980079       1 metal3datatemplate_manager.go:435] "msg"="Deleted Metal3DataClaim" "Metal3DataClaim"="wc-1458191268-kubeadm-capm3-virt-control-plane-fn667" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:50.998850       1 metal3datatemplate_manager.go:383] "msg"="Deleting Metal3DataClaim" "Metal3DataClaim"="wc-1458191268-kubeadm-capm3-virt-control-plane-7tjgn" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:51.023508       1 metal3datatemplate_manager.go:421] "msg"="Deleted Metal3Data" "Metal3Data"="wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-1" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:51.023536       1 metal3datatemplate_manager.go:435] "msg"="Deleted Metal3DataClaim" "Metal3DataClaim"="wc-1458191268-kubeadm-capm3-virt-control-plane-7tjgn" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:51.075542       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:51.083012       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:51.092095       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:47:51.360889       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:51:51.222374       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
I0918 09:51:51.234904       1 metal3datatemplate_manager.go:108] "msg"="Fetching Metal3Data objects" "logger"="controllers.Metal3DataTemplate.Metal3DataTemplate-controller" "metal3-datatemplate"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f"}
```

(no log  after `09:51:51` for our Metal3DataTemplate)

The above shows among other things that the Metal3DataClaims where deleted quite early (09:47:51)

• on related resources (same prefix)

```
I0918 09:53:40.305945       1 metal3data_manager.go:414] "msg"="Releasing address from IPPool" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"} "pool name"="wc-1458191268-kubeadm-capm3-virt-provisioning-pool"
I0918 09:53:40.327924       1 metal3data_manager.go:414] "msg"="Releasing address from IPPool" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"} "pool name"="wc-1458191268-kubeadm-capm3-virt-provisioning-pool"
I0918 09:53:40.343495       1 metal3data_manager.go:1468] "msg"="Address claim not found" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"}
I0918 09:53:40.343533       1 metal3data_manager.go:414] "msg"="Releasing address from IPPool" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"} "pool name"="wc-1458191268-kubeadm-capm3-virt-primary-pool"
I0918 09:53:40.364283       1 metal3data_manager.go:1468] "msg"="Address claim not found" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"}
I0918 09:53:40.391932       1 metal3data_manager.go:414] "msg"="Releasing address from IPPool" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"} "pool name"="wc-1458191268-kubeadm-capm3-virt-provisioning-pool"
I0918 09:53:40.399177       1 metal3data_manager.go:1468] "msg"="Address claim not found" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"}
I0918 09:53:40.399424       1 metal3data_manager.go:414] "msg"="Releasing address from IPPool" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"} "pool name"="wc-1458191268-kubeadm-capm3-virt-primary-pool"
I0918 09:53:40.406140       1 metal3data_manager.go:1468] "msg"="Address claim not found" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"}
I0918 09:53:40.412083       1 metal3data_controller.go:79] "msg"="failed to Patch Metal3Data" "logger"="controllers.Metal3Data.Metal3Data-controller" "metal3-data"={"Namespace":"kubeadm-capm3-virt","Name":"wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0"}
```

This shows that:
* until 09:53:40 the wc-1458191268-kubeadm-capm3-virt-cp-metadata-0204878f2f-0 Metal3Data was still around and reconciled
* (the dump tells us that at some point later it wasn't anymore, presumably when the ns was deleted)

**Tentative analysis**

I've been looking at the code to understand why the Metal3DataTemplate reconciliation can have been unable to remove the finalizer, and unable to remove the non-existing Metal3DataClaim from the status.indexes.

I'm trying to infer what the code would have seen between 09:47:51 (when the deletionTimestamp was set on Metal3DataTemplate, and when the Metal3DataClaims were deleted) and the deletion of the Metal3Data (which happened after 09:53:40).

So, in this windows:
* the Metal3DataTemplate is present
* the Metal3DataClaim wc-1458191268-kubeadm-capm3-virt-control-plane-fn667 is absent (all the relevant Metal3DataClaim are absent)
* the Metal3Data is present

The code in [DataTemplateManager.UpdateDatas](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3datatemplate_manager.go#L174) seems to:
* rely on [getIndexes](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3datatemplate_manager.go#L128) to get a map of indexes integers to Metal3DataClaim names, based on listing Metal3Data resources and looking at their Spec.Claim.Name
    * in our scenario this would result in finding the Metal3Data, and returning `map[0: wc-1458191268-kubeadm-capm3-virt-control-plane-fn667]`
* then UpdateDatas lists Metal3DataClaims and for those it finds, calls `indexes, err = m.updateData(ctx, &dataClaim, indexes` to update the indexes
    * in our scenario, since no Metal3DataClaim exists, it would never have reached `indexes, err = m.updateData(ctx, &dataClaim, indexes)`
* then, it returns the length of indexes
    * in our scenario, it seems that it would be 1 (what getIndexes returns, unmodified since the list objects loop isn't entered)

This is consistent with the observation that the allocationsCount seen by reconcileDelete is 1, that the finalizer is hence not removed, and the status.indexes we see on the Metal3DataTemplate resource.

It seems that UpdateDatas should check that the Metal3DataClaims names returned by getIndexes actually exist, and remove from 'indexes' the ones that are not.

I would speculate that #1478 would have made this latent issue occur more frequently. Indeed #1478 "Before cleaning up Metal3Data, check that the Metal3DataClaim is gone", merged in the 1.8.x development timeframe, would possibly create the problematic condition: Metal3DataClaim gone, but Metal3Data still points to it. 

**Worth noting**

The logs show that there are various issues occurring at differents points. I haven't digged those specifically, with the idea that despite transient errors, the processing should still be able to converge, and also because I suspect that the way Sylva tears down CAPI and CAPI providers resources may not be orderly enough and might cause these issues.

**Environment:**

- Cluster-api version: 1.7.6
- Cluster-api-provider-metal3 version: 1.8.1 specifically (is
- Environment (metal3-dev-env or other): Sylva project CI pipelines (more details below)
- Kubernetes version: (use `kubectl version`): 1.28.12

/kind bug

pierrecregut:

This PR fixes some non terminating Data/DataTemplate deletion. Some Metal3Data and Metal3DataTemplates were not destroyed  in rare circumstances due to missing reconcile events caused  by two distinct problems.

DataTemplate deletion has a watch over DataClaims but deletion  can only be performed when Data are completely removed (finalizer  executed) because Data resource requires the template.
    
We keep the facts that Data and DataClaims are still existing.
* When DataClaims exist, we do not remove the finalizer and wait   for reconcile events from the watch.
* If we have neither DataClaims or Data, we can safely  remove the finalizer. Deletion is complete.
* Otherwise we have to make a busy wait on Data deletion as there is no more claim to generate event but we still  need to wait until the Data finalizers have been executed.

The other source of problem is Metal3Data deletion : if Metal3Data is deleted before the Metal3DataClaim, we must actively wait for the claim deletion, as no other reconciliation event will be triggered.

**Alternative implementations**
Data resources may not need the template during deletion as we want to remove every ip claim associated to it. In that case,
we only need to wait for data claims deletion and we can remove the busy wait from templates over data resources.

Fixes #1994

### DIFF
--- a/baremetal/metal3datatemplate_manager.go
+++ b/baremetal/metal3datatemplate_manager.go
@@ -40,8 +40,8 @@ type DataTemplateManagerInterface interface {
 	UnsetFinalizer()
 	SetClusterOwnerRef(*clusterv1.Cluster) error
 	// UpdateDatas handles the Metal3DataClaims and creates or deletes Metal3Data accordingly.
-	// It returns the number of current allocations.
-	UpdateDatas(context.Context) (int, error)
+	// It returns if there are still Data object and undeleted DataClaims objects.
+	UpdateDatas(context.Context) (bool, bool, error)
 }
 
 // DataTemplateManager is responsible for performing machine reconciliation.
@@ -135,12 +135,7 @@ func (m *DataTemplateManager) getIndexes(ctx context.Context) (map[int]string, e
 			continue
 		}
 
-		// Get the claim Name, if unset use empty string, to still record the
-		// index being used, to avoid conflicts
-		claimName := ""
-		if dataObject.Spec.Claim.Name != "" {
-			claimName = dataObject.Spec.Claim.Name
-		}
+		claimName := dataObject.Spec.Claim.Name
 		m.DataTemplate.Status.Indexes[claimName] = dataObject.Spec.Index
 		indexes[dataObject.Spec.Index] = claimName
 	}
@@ -170,12 +165,13 @@ func (m *DataTemplateManager) updateStatusTimestamp() {
 }
 
 // UpdateDatas handles the Metal3DataClaims and creates or deletes Metal3Data accordingly.
-// It returns the number of current allocations.
-func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (int, error) {
+// It returns if there are still Data object and undeleted DataClaims objects.
+func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (bool, bool, error) {
 	indexes, err := m.getIndexes(ctx)
 	if err != nil {
-		return 0, err
+		return false, false, err
 	}
+	hasData := len(indexes) > 0
 
 	// get list of Metal3DataClaim objects
 	dataClaimObjects := infrav1.Metal3DataClaimList{}
@@ -186,9 +182,10 @@ func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (int, error) {
 
 	err = m.client.List(ctx, &dataClaimObjects, opts)
 	if err != nil {
-		return 0, err
+		return false, false, err
 	}
 
+	hasClaims := false
 	// Iterate over the Metal3Data objects to find all indexes and objects
 	for _, dataClaim := range dataClaimObjects.Items {
 		dataClaim := dataClaim
@@ -196,18 +193,20 @@ func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (int, error) {
 		if dataClaim.Spec.Template.Name != m.DataTemplate.Name {
 			continue
 		}
-
-		if dataClaim.Status.RenderedData != nil && dataClaim.DeletionTimestamp.IsZero() {
-			continue
+		if dataClaim.DeletionTimestamp.IsZero() {
+			hasClaims = true
+			if dataClaim.Status.RenderedData != nil {
+				continue
+			}
 		}
 
 		indexes, err = m.updateData(ctx, &dataClaim, indexes)
 		if err != nil {
-			return 0, err
+			return false, false, err
 		}
 	}
 	m.updateStatusTimestamp()
-	return len(indexes), nil
+	return hasData, hasClaims, nil
 }
 
 func (m *DataTemplateManager) updateData(ctx context.Context,
--- a/baremetal/metal3datatemplate_manager_test.go
+++ b/baremetal/metal3datatemplate_manager_test.go
@@ -256,7 +256,7 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 			)
 			Expect(err).NotTo(HaveOccurred())
 
-			nbIndexes, err := templateMgr.UpdateDatas(context.TODO())
+			hasData, _, err := templateMgr.UpdateDatas(context.TODO())
 			if tc.expectRequeue || tc.expectError {
 				Expect(err).To(HaveOccurred())
 				if tc.expectRequeue {
@@ -267,7 +267,7 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 			} else {
 				Expect(err).NotTo(HaveOccurred())
 			}
-			Expect(nbIndexes).To(Equal(tc.expectedNbIndexes))
+			Expect(hasData).To(Equal(tc.expectedNbIndexes > 0))
 			Expect(tc.template.Status.LastUpdated.IsZero()).To(BeFalse())
 			Expect(tc.template.Status.Indexes).To(Equal(tc.expectedIndexes))
 
--- a/baremetal/mocks/zz_generated.metal3datatemplate_manager.go
+++ b/baremetal/mocks/zz_generated.metal3datatemplate_manager.go
@@ -92,12 +92,13 @@ func (mr *MockDataTemplateManagerInterfaceMockRecorder) UnsetFinalizer() *gomock
 }
 
 // UpdateDatas mocks base method.
-func (m *MockDataTemplateManagerInterface) UpdateDatas(arg0 context.Context) (int, error) {
+func (m *MockDataTemplateManagerInterface) UpdateDatas(arg0 context.Context) (bool, bool, error) {
 	m.ctrl.T.Helper()
 	ret := m.ctrl.Call(m, "UpdateDatas", arg0)
-	ret0, _ := ret[0].(int)
-	ret1, _ := ret[1].(error)
-	return ret0, ret1
+	ret0, _ := ret[0].(bool)
+	ret1, _ := ret[1].(bool)
+	ret2, _ := ret[2].(error)
+	return ret0, ret1, ret2
 }
 
 // UpdateDatas indicates an expected call of UpdateDatas.
--- a/controllers/metal3data_controller.go
+++ b/controllers/metal3data_controller.go
@@ -115,11 +115,14 @@ func (r *Metal3DataReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	if !metal3Data.ObjectMeta.DeletionTimestamp.IsZero() {
 		// Check if the Metal3DataClaim is gone. We cannot clean up until it is.
 		err := r.Client.Get(ctx, types.NamespacedName{Name: metal3Data.Spec.Claim.Name, Namespace: metal3Data.Spec.Claim.Namespace}, &infrav1.Metal3DataClaim{})
-		if err != nil && apierrors.IsNotFound(err) {
-			return r.reconcileDelete(ctx, metadataMgr)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return r.reconcileDelete(ctx, metadataMgr)
+			}
+			return ctrl.Result{}, err
 		}
-		// We were unable to determine if the Metal3DataClaim is gone
-		return ctrl.Result{}, err
+		// Requeue until Metal3DataClaim is gone.
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil
 	}
 
 	// Handle non-deleted machines
--- a/controllers/metal3datatemplate_controller.go
+++ b/controllers/metal3datatemplate_controller.go
@@ -142,7 +142,7 @@ func (r *Metal3DataTemplateReconciler) reconcileNormal(ctx context.Context,
 	// If the Metal3DataTemplate doesn't have finalizer, add it.
 	dataTemplateMgr.SetFinalizer()
 
-	_, err := dataTemplateMgr.UpdateDatas(ctx)
+	_, _, err := dataTemplateMgr.UpdateDatas(ctx)
 	if err != nil {
 		return checkReconcileError(err, "Failed to recreate the status")
 	}
@@ -152,16 +152,19 @@ func (r *Metal3DataTemplateReconciler) reconcileNormal(ctx context.Context,
 func (r *Metal3DataTemplateReconciler) reconcileDelete(ctx context.Context,
 	dataTemplateMgr baremetal.DataTemplateManagerInterface,
 ) (ctrl.Result, error) {
-	allocationsCount, err := dataTemplateMgr.UpdateDatas(ctx)
+	hasData, hasClaims, err := dataTemplateMgr.UpdateDatas(ctx)
 	if err != nil {
 		return checkReconcileError(err, "Failed to recreate the status")
 	}
 
-	if allocationsCount == 0 {
-		// metal3datatemplate is marked for deletion and ready to be deleted,
-		// so remove the finalizer.
-		dataTemplateMgr.UnsetFinalizer()
+	if hasClaims {
+		return ctrl.Result{}, nil
 	}
+	if hasData {
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil
+	}
+
+	dataTemplateMgr.UnsetFinalizer()
 
 	return ctrl.Result{}, nil
 }
--- a/controllers/metal3datatemplate_controller_test.go
+++ b/controllers/metal3datatemplate_controller_test.go
@@ -83,9 +83,9 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 				}
 			}
 			if tc.m3dt != nil && !tc.m3dt.DeletionTimestamp.IsZero() && tc.reconcileDeleteError {
-				m.EXPECT().UpdateDatas(context.Background()).Return(0, errors.New(""))
+				m.EXPECT().UpdateDatas(context.Background()).Return(false, false, errors.New(""))
 			} else if tc.m3dt != nil && !tc.m3dt.DeletionTimestamp.IsZero() {
-				m.EXPECT().UpdateDatas(context.Background()).Return(0, nil)
+				m.EXPECT().UpdateDatas(context.Background()).Return(false, false, nil)
 				m.EXPECT().UnsetFinalizer()
 			}
 
@@ -93,9 +93,9 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 				tc.reconcileNormal {
 				m.EXPECT().SetFinalizer()
 				if tc.reconcileNormalError {
-					m.EXPECT().UpdateDatas(context.Background()).Return(0, errors.New(""))
+					m.EXPECT().UpdateDatas(context.Background()).Return(false, false, errors.New(""))
 				} else {
-					m.EXPECT().UpdateDatas(context.Background()).Return(1, nil)
+					m.EXPECT().UpdateDatas(context.Background()).Return(true, true, nil)
 				}
 			}
 
@@ -232,9 +232,9 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 			m.EXPECT().SetFinalizer()
 
 			if !tc.UpdateError {
-				m.EXPECT().UpdateDatas(context.TODO()).Return(1, nil)
+				m.EXPECT().UpdateDatas(context.TODO()).Return(true, true, nil)
 			} else {
-				m.EXPECT().UpdateDatas(context.TODO()).Return(0, errors.New(""))
+				m.EXPECT().UpdateDatas(context.TODO()).Return(false, false, errors.New(""))
 			}
 
 			res, err := r.reconcileNormal(context.TODO(), m)
@@ -284,12 +284,12 @@ var _ = Describe("Metal3DataTemplate manager", func() {
 			m := baremetal_mocks.NewMockDataTemplateManagerInterface(gomockCtrl)
 
 			if !tc.DeleteError && tc.DeleteReady {
-				m.EXPECT().UpdateDatas(context.TODO()).Return(0, nil)
+				m.EXPECT().UpdateDatas(context.TODO()).Return(false, false, nil)
 				m.EXPECT().UnsetFinalizer()
 			} else if !tc.DeleteError {
-				m.EXPECT().UpdateDatas(context.TODO()).Return(1, nil)
+				m.EXPECT().UpdateDatas(context.TODO()).Return(true, true, nil)
 			} else {
-				m.EXPECT().UpdateDatas(context.TODO()).Return(0, errors.New(""))
+				m.EXPECT().UpdateDatas(context.TODO()).Return(false, false, errors.New(""))
 			}
 
 			res, err := r.reconcileDelete(context.TODO(), m)


pierrecregut:
There was a bug in the first version (the status.Indexes map must be updated even when the claim is removed so that the mapping can be accessed and then removed during delete. Even if the new version will pass the tests, it opens in fact the question that there is a latent race already existing during deletion. the indexes from getIndexes are recomputed to remove the ones corresponding to the DataClaims being removed but we cannot be sure that the corresponding Data had the time to run their reconcile loop that removes the finalizer as [it needs the template to remove the IpClaims](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3data_manager.go#L306). If that is the case, the only really safe solution is closer to Thomas original proposal and last implementation alternative:
* compute boolean liveData as len(indexes) > 0 where indexes is the result of getIndexes() without the changes applied during updateData().
* compute boolean liveClaims to be true iff there is a claim listed with delete timestamp set to 0. 
* if liveClaims is true, just return without requeue as we wait for new events from the claims (and we may keep the template for a long time if nobody delete the claims).
* if liveClaims and liveData are both false, remove the finalizer, the deletion is finished
* if liveClaims is false and liveData is true, return with an explicit requeue. Requeue should be avoided but we *must* perform an active wait until the Data has been deleted but as we are waiting for finalizers, it should be relatively quick...

baremetal/metal3datatemplate_manager.go
```diff
@@ -135,17 +136,12 @@ func (m *DataTemplateManager) getIndexes(ctx context.Context) (map[int]string, e
...
-		// Get the claim Name, if unset use empty string, to still record the
-		// index being used, to avoid conflicts
-		claimName := ""
-		if dataObject.Spec.Claim.Name != "" {
-			claimName = dataObject.Spec.Claim.Name
-		}
-		m.DataTemplate.Status.Indexes[claimName] = dataObject.Spec.Index
+		claimName := dataObject.Spec.Claim.Name
+		statusIndexes[claimName] = dataObject.Spec.Index
 		indexes[dataObject.Spec.Index] = claimName
```
adilGhaffarDev, line 141: `indexes` and `statusIndexes` maps are the same, in case `indexes` `dataObject.Spec.Index` is key and value is`claimName` and `statusIndexes` has `claimName` as key and `dataObject.Spec.Index` as value. Can we not achieve what we are trying to do just by `indexes`?

adilGhaffarDev:
> There was a bug in the first version (the status.Indexes map must be updated even when the claim is removed so that the mapping can be accessed and then removed during delete. Even if the new version will pass the tests, it opens in fact the question that there is a latent race already existing during deletion. the indexes from getIndexes are recomputed to remove the ones corresponding to the DataClaims being removed but we cannot be sure that the corresponding Data had the time to run their reconcile loop that removes the finalizer as [it needs the template to remove the IpClaims](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3data_manager.go#L306). If that is the case, the only really safe solution is closer to Thomas original proposal and last implementation alternative:
> 
>     * compute boolean liveData as len(indexes) > 0 where indexes is the result of getIndexes() without the changes applied during updateData().
> 
>     * compute boolean liveClaims to be true iff there is a claim listed with delete timestamp set to 0.
> 
>     * if liveClaims is true, just return without requeue as we wait for new events from the claims (and we may keep the template for a long time if nobody delete the claims).
> 
>     * if liveClaims and liveData are both false, remove the finalizer, the deletion is finished
> 
>     * if liveClaims is false and liveData is true, return with an explicit requeue. Requeue should be avoided but we _must_ perform an active wait until the Data has been deleted but as we are waiting for finalizers, it should be relatively quick.

Yes, we need to check both liveClaims and liveData before deleting dataTemplate.
 I have tested Thomas's PR, and after provisioning and deprovisiong worker nodes for few times, I started seeing leftover metal3Datas and ipclaims. 
I agree with the solution that you are proposing. If you have time please open a PR or update this one, if not I can work on it.

baremetal/metal3datatemplate_manager.go
pierrecregut: They are "co-maps" if you want to access from index or from claims you will not use the same map. 
* statusIndexes "replaces" m.DataTemplate.Status.Indexes. We use it to get a filtered Indexes where we only keep those for which a claim exists
* indexes is in fact used as a set (value never really used). It is used to find which index can be used when a Data is created.
We could defer the creation of the indexes map. If there is no Data creation it would be faster but unless we are careful it may be slower if there are multiple Data created. This is an optimization (or not) unrelated to the issue.

pierrecregut:
@adilGhaffarDev I will do it by modifying this PR (no need for a third one). This is the safe solution

adilGhaffarDev:
cc @kashifest @mboukhalfa

tmmorin:
> @pierrecregut @tmmorin , thank you for working on it, also have you guys tested this on your side? is it solving your issue?

I haven't tested this one yet (I only tested #1997).

I plan to do a local build and a test later this week.

baremetal/metal3datatemplate_manager.go
```diff
@@ -186,28 +182,31 @@ func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (int, error) {
...
-		if dataClaim.Status.RenderedData != nil && dataClaim.DeletionTimestamp.IsZero() {
-			continue
+		if dataClaim.DeletionTimestamp.IsZero() {
+			hasClaims = true
+			if dataClaim.Status.RenderedData != nil {
+				continue
+			}
 		}
 
 		indexes, err = m.updateData(ctx, &dataClaim, indexes)
```
kashifest, line 203: what is the use of this indexes map here ? are we using it ?

kashifest (commented):
/approve
/hold
some comments inline

Sunnatillo (commented):
I have reviewed both PRs. Thank you @pierrecregut and @tmmorin.
I would lean towards this implementation. Thank you for working on it. 

LGTM

Sunnatillo:
Lets test this several times before merging it.

pierrecregut:
Thomas has tested the patch in the Sylva setting. It corrects some of the problems, but we encounter other issues not related to the logic of the DataTemplate controller, but rather the Data controller.

*  For context, in the Sylva test run, the namespace of the cluster is deleted before all the CAPI resources for the cluster have been removed. In this case, all objects are deleted in random order. If a Metal3Data is deleted before the claim, the deletion reconciliation is [aborted but not requeued](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/controllers/metal3data_controller.go#L117-L122). When the claim is deleted, garbage collection due to the owner reference may be triggered, but the Data is already marked as deleted, so nothing happens. If err is nil, a requeue must be generated.
*  On rare occasions, the [deferred patch fails](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/controllers/metal3data_controller.go#L117-L122). Unfortunately, the error is ignored, so no further reconciliation will be attempted if there was no error. The fix is simple: assign err to rerr if it is non-nil. This pattern is present in seven controllers; only the remediation controller is correct.

*  Finally, we have a design question regarding how Data is deleted. The delete operation lists the IP pools used in the template to find the IPClaim based on the labels, both on the Data name and the IP pool name. However, do we need to select based on the IP pool name if we will browse over all potential values? If we delete all the IPClaims with the correct label Data, then we do not need the template. Consequently, we do not need to block template deletion until the Data disappears, which would help us resolve the conflict of waiting for DataClaims and counting Data during template deletion.

pierrecregut:
So we propose to add the fix for first point to this PR. We think that point 2 is rare but more general and should be addressed somewhere else. Finally 3 could be a more major refactoring that simplifies the logic but needs much more care to be implemented and can be done later.

tmmorin:
Thanks for the recap @pierrecregut - and yes, I agree with that what you propose.

pierrecregut:
I have pushed #1 as a separate commit  in the PR and changed the comment on UpdateData. The pair should solve #1994 except for corner cases (point 2 above).

pierrecregut:
point 2 is issue #2002 for triaging with associated PR #2003

tmmorin:
I've run tests with a local build made from this branch, and I observed no resource deletion issue. :+1:

adilGhaffarDev:
I have tested it on my side, it's working fine.
/lgtm

adilGhaffarDev:
@pierrecregut can you squash the commits?

pierrecregut:
@adilGhaffarDev squash done but as commits solve two distinct issues, I wouldn't have done it on my own.

adilGhaffarDev:
> @adilGhaffarDev squash done but as commits solve two distinct issues, I wouldn't have done it on my own.
> 
> 

We try to keep one commit in one PR , it helps in keeping release notes tidy. Can you update the PR description too and add info regarding the metal3data reconciliation fix.

adilGhaffarDev:
> *  Finally, we have a design question regarding how Data is deleted. The delete operation lists the IP pools used in the template to find the IPClaim based on the labels, both on the Data name and the IP pool name. However, do we need to select based on the IP pool name if we will browse over all potential values? If we delete all the IPClaims with the correct label Data, then we do not need the template. Consequently, we do not need to block template deletion until the Data disappears, which would help us resolve the conflict of waiting for DataClaims and counting Data during template deletion.
> 
> 

Yes this seems like a unnecessary dependency on m3datatemplate in our deletion workflow, I am not sure what will be the impact if change this. Lets do it in separate PR and keep it on Main and not back-port to v1.8, we will test it on Main CI, and if everything goes well it will be in v1.9. 
For v1.8 this PR is good to go in.

adilGhaffarDev:
the test was already passed before the squash no need to run again.
/lgtm

adilGhaffarDev:
/unhold
Tests passed I am unholding.

PR #2003, opened 2024-10-02
:bug: Propagate errors during patch to reconcile

### LINKED ISSUE #2002: Silently Ignored Errors Leading to Non-Converging Reconciliation
In integration runs, we see message "Failed to patch XXXX" where XXXX is the name of a Metal3 resource. This will happen if the update was conflicting with another operation on the resource. If the patch was performed on an otherwise succesful reconcile, the changes are silently ignored and there will be no retry. 

As for other errors, the patch error should be reported in the output of reconcile.

metal3remediation_controller is the only correct controller.

**Environment:**

- Cluster-api version: -
- Cluster-api-provider-metal3 version: main
- Environment (metal3-dev-env or other): -
- Kubernetes version: (use `kubectl version`): -

/kind bug

pierrecregut:
If an otherwise succesful reconcile has a failure during the deferred patch process (typically conflict), the error is ignored, the resource is never modified and the reconciliation may never be retriggerred.

The fix propagates error err to global variable rerr.

Fixes #2002

### DIFF
--- a/controllers/metal3cluster_controller.go
+++ b/controllers/metal3cluster_controller.go
@@ -88,6 +88,7 @@ func (r *Metal3ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	defer func() {
 		if err := patchMetal3Cluster(ctx, patchHelper, metal3Cluster); err != nil {
 			clusterLog.Error(err, "failed to Patch metal3Cluster")
+			rerr = err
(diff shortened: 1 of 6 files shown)


Rozzii (commented):
/lgtm

adilGhaffarDev (approved):
/approve

PR #2010, opened 2024-10-04
🌱 Add scalability test

mboukhalfa:
Add test that apply scaled number of bmh on dev-env using fakeIPA.
by default the test apply 100 BMH by batches each 20 BMHs applied in the same time 

TODO:

- Merge prow trigger: https://gerrit.nordix.org/c/infra/cicd/+/22355
- Merge DevEnv running fakeIPA PR: https://github.com/metal3-io/metal3-dev-env/pull/1450

### DIFF
--- a/test/e2e/scalability_test.go
+++ b/test/e2e/scalability_test.go
@@ -1,0 +1,57 @@
+package e2e
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+/*
+ * Apply scaled number of BMHs per batches and wait for the them to become available
+ * The test pass when all the BMHs become available
+ */
+
+var _ = Describe("When testing scalability with fakeIPA [scalability]", Label("scalability"), func() {
+	scalabilityTest()
+})
+
+func scalabilityTest() {
+	It("Should apply BMHs per batches and wait for them to become available", func() {
+		numNodes, _ := strconv.Atoi(e2eConfig.GetVariable("NUM_NODES"))
+		batch, _ := strconv.Atoi(e2eConfig.GetVariable("BMH_BATCH_SIZE"))
+		Logf("Starting scalability test")
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+		ListBareMetalHosts(ctx, bootstrapClient, client.InNamespace(namespace))
+		applyBatchBmh := func(from int, to int) {
+			Logf("Apply BMH batch from node_%d to node_%d", from, to)
+			for i := from; i < to+1; i++ {
+				resource, err := os.ReadFile(filepath.Join(workDir, fmt.Sprintf("bmhs/node_%d.yaml", i)))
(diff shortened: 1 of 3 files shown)


adilGhaffarDev (commented):
/lgtm
/hold 
holding it because it is pointing to PR in dev-env

test/e2e/scalability_test.go
```diff
@@ -0,0 +1,57 @@
...
+ * Apply scaled number of BMHs per batches and wait for the them to become available
+ * The test pass when all the BMHs become available
+ */
+
+var _ = Describe("When testing scalability with fakeIPA [scalability]", Label("scalability"), func() {
+	scalabilityTest()
+})
+
+func scalabilityTest() {
+	It("Should apply per batches BMHs wait for them to become available", func() {
```
kashifest, line 25:
```suggestion
	It("Should apply BMHs per batches and wait for them to become available", func() {
```

test/e2e/scalability_test.go
```diff
@@ -0,0 +1,57 @@
...
+		numNodes, _ := strconv.Atoi(e2eConfig.GetVariable("NUM_NODES"))
+		batch, _ := strconv.Atoi(e2eConfig.GetVariable("BMH_BATCH_SIZE"))
+		Logf("Starting scalability test")
+		bootstrapClient := bootstrapClusterProxy.GetClient()
+		ListBareMetalHosts(ctx, bootstrapClient, client.InNamespace(namespace))
+		applyBatchBmh := func(from int, to int) {
+			Logf("Apply BMH batch from node_%d to node_%d", from, to)
```
kashifest, line 32: what is the reason that we are applying BMHs in batches? We have a `PROVISIONING_LIMIT` in [BMO](https://github.com/metal3-io/baremetal-operator/blob/main/docs/configuration.md?plain=1#L38). Can we utilize that?

lentzi90: I think it may be a good idea to have batches still. It gives us "checkpoints" that could be helpful when determining if we have a scalability issue or just some other random issue. E.g. if the first 60 BMHs were ok but after that it failed, then it could be a scalability issue. But if it fails directly for the first 20 then it is likely some other issue.

lentzi90 (commented):
This looks ok to me. Is the plan to do more in this test eventually, or why is it in CAPM3? As it is now it would make more sense in BMO (or dev-env) since it is just dealing with BMHs.
/lgtm

mboukhalfa:
> This looks ok to me. Is the plan to do more in this test eventually, or why is it in CAPM3? As it is now it would make more sense in BMO (or dev-env) since it is just dealing with BMHs. /lgtm

This is the first step of applying BMH but will add on top of this cluster scalability i.e applying capm3 cluster templates on fake-env with certain available bmh 

We might think of adding a scalability test that just test BMO or even just ironic somewhere

test/e2e/scalability_test.go
kashifest: That is true, I am ok with this. Since BMO already has a safety net there and I am trusting that https://github.com/metal3-io/metal3-docs/blob/main/design/baremetal-operator/limit-hosts-provisioning.md added the limit based on some statistics , I thought batching here wont actually add much value.

mboukhalfa: ok if @kashifest agree we can merge it

kashifest: /approve

mboukhalfa:
/hold 
waiting for pipeline https://gerrit.nordix.org/c/infra/cicd/+/22355

PR #2026, opened 2024-10-10
:seedling: make CAPI a pattern in dependabot

tuminoid:
Dependabot always bumps the CAPI main package and CAPI test package separately, which is waste of PRs and CPU time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,6 +29,8 @@ updates:
   groups:
     kubernetes:
       patterns: ["k8s.io/*"]
+    capi:
+      patterns: ["sigs.k8s.io/cluster-api", "sigs.k8s.io/cluster-api/test"]
   ignore:
   # Ignore controller-runtime major and minor bumps as its upgraded manually.
   - dependency-name: "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -29,6 +29,8 @@ updates:
   groups:
     kubernetes:
       patterns: ["k8s.io/*"]
+    capi:
+      patterns: ["sigs.k8s.io/cluster-api", "sigs.k8s.io/cluster-api/test"]
```
kashifest, line 33: This good already, we can also do the following as k8s above 
```suggestion
      patterns: ["sigs.k8s.io/cluster-api/*]
```

kashifest (commented):
/approve

.github/dependabot.yml
tuminoid: Does this actually work, as the cluster-api dep does not include trailing slash. so it does not match?

```
    sigs.k8s.io/cluster-api v1.8.3
```

kashifest: You have a point. I am not sure TBH. Let’s merge it like this

tuminoid:
/override metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main

PR #2039, opened 2024-10-14
🌱 delete the process of removing bmhs from default namespace

mboukhalfa:
Fixes # https://github.com/metal3-io/metal3-dev-env/issues/1383

PR #2046, opened 2024-10-19
🌱 Fixing issues in healthcheck test

adilGhaffarDev:
Adding more logs and fixing issues in healthcheck test.
This PR is 
- Making sure that Metal3RemediationTemplate is deleted after the healthcheck test is done.
- Wrapping "kubelet stop" command around Eventually so it does not fail. If eventually is not used the ssh connection fails sometimes, and failure is most of the time in the first "kubelet stop"(https://github.com/metal3-io/cluster-api-provider-metal3/blob/aa67ad5c84af874212be142c92aac195114556c9/test/e2e/healthchek.go#L51). Adding some wait eventually fixes the issue.
- Fixing FetchClusterLogs function, in case of any error FetchClusterLogs was returning and missing logs for remaining pods, that is fixed in this PR.
- Changing logs folder from "clusters" to "target_cluster_logs", in project infra we are only checking "target_cluster_logs".
- This PR is also fixing the failure "No Control Plane machines came into existence. " in remediation tests, it started happening after this PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2039/files

### DIFF
--- a/test/e2e/basic_integration_test.go
+++ b/test/e2e/basic_integration_test.go
@@ -16,7 +16,7 @@ var _ = Describe("When testing basic cluster creation [basic]", Label("basic"), 
 		validateGlobals(specName)
 
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
 	})
 
 	It("Should create a workload cluster", func() {
@@ -27,6 +27,6 @@ var _ = Describe("When testing basic cluster creation [basic]", Label("basic"), 
 	})
 
 	AfterEach(func() {
-		DumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, artifactFolder, namespace, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
+		DumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, targetCluster, artifactFolder, namespace, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
 	})
 })
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -125,19 +125,25 @@ func getSha256Hash(filename string) ([]byte, error) {
 	return hash.Sum(nil), nil
 }
 
-func DumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterProxy framework.ClusterProxy, artifactFolder string, namespace string, intervalsGetter func(spec, key string) []interface{}, clusterName, clusterctlLogFolder string, skipCleanup bool) {
+// TODO change this function to handle multiple workload(target) clusters.
+func DumpSpecResourcesAndCleanup(ctx context.Context, specName string, bootstrapClusterProxy framework.ClusterProxy, targetClusterProxy framework.ClusterProxy, artifactFolder string, namespace string, intervalsGetter func(spec, key string) []interface{}, clusterName, clusterctlLogFolder string, skipCleanup bool) {
 	Expect(os.RemoveAll(clusterctlLogFolder)).Should(Succeed())
-	clusterClient := clusterProxy.GetClient()
+	clusterClient := bootstrapClusterProxy.GetClient()
 
-	clusterProxy.CollectWorkloadClusterLogs(ctx, namespace, clusterName, artifactFolder)
+	bootstrapClusterProxy.CollectWorkloadClusterLogs(ctx, namespace, clusterName, artifactFolder)
 
+	By("Fetch logs from target cluster")
+	err := FetchClusterLogs(targetClusterProxy, clusterLogCollectionBasePath)
+	if err != nil {
+		Logf("Error: %v", err)
+	}
 	// Dumps all the resources in the spec namespace, then cleanups the cluster object and the spec namespace itself.
 	By(fmt.Sprintf("Dumping all the Cluster API resources in the %q namespace", namespace))
 	// Dump all Cluster API related resources to artifacts before deleting them.
 	framework.DumpAllResources(ctx, framework.DumpAllResourcesInput{
 		Lister:    clusterClient,
 		Namespace: namespace,
-		LogPath:   filepath.Join(artifactFolder, "clusters", clusterProxy.GetName(), "resources"),
+		LogPath:   filepath.Join(artifactFolder, bootstrapClusterProxy.GetName(), "resources"),
 	})
 
 	if !skipCleanup {
--- a/test/e2e/ip_reuse_test.go
+++ b/test/e2e/ip_reuse_test.go
@@ -18,7 +18,7 @@ var _ = Describe("When testing ip reuse [ip-reuse] [features]", Label("ip-reuse"
 		validateGlobals(specName)
 
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
 	})
 	It("Should create a workload cluster then verify ip allocation reuse while upgrading k8s", func() {
 		IPReuse(ctx, func() IPReuseInput {
@@ -39,6 +39,6 @@ var _ = Describe("When testing ip reuse [ip-reuse] [features]", Label("ip-reuse"
 		ListMetal3Machines(ctx, bootstrapClusterProxy.GetClient(), client.InNamespace(namespace))
 		ListMachines(ctx, bootstrapClusterProxy.GetClient(), client.InNamespace(namespace))
 		ListNodes(ctx, targetCluster.GetClient())
-		DumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, artifactFolder, namespace, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
+		DumpSpecResourcesAndCleanup(ctx, specName, bootstrapClusterProxy, targetCluster, artifactFolder, namespace, e2eConfig.GetIntervals, clusterName, clusterctlLogFolder, skipCleanup)
 	})
 })
--- a/test/e2e/md_remediations_test.go
+++ b/test/e2e/md_remediations_test.go
@@ -15,13 +15,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	bmhCrsFile = "bmhosts_crs.yaml"
+)
+
 var _ = Describe("When testing MachineDeployment remediation [healthcheck] [remediation] [features]", Label("healthcheck", "remediation", "features"), func() {
 	BeforeEach(func() {
 		osType := strings.ToLower(os.Getenv("OS"))
 		Expect(osType).ToNot(Equal(""))
 		validateGlobals(specName)
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
+
+		Logf("Removing existing BMHs from source")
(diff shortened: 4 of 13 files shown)


test/e2e/basic_integration_test.go
```diff
@@ -16,7 +16,7 @@ var _ = Describe("When testing basic cluster creation [basic]", Label("basic"),
 		validateGlobals(specName)
 
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
```
mquhuy, line 19: It should be `target_clusters_logs`, if there could be more than 1 target cluster. Or we can omit `logs` (which makes it `target_clusters`), as it's clearly a log directory.

And why do we include bootstrap cluster name here? Shouldn't it be the target cluster name?

test/e2e/common.go
```diff
@@ -125,19 +125,24 @@ func getSha256Hash(filename string) ([]byte, error) {
...
+		Logf("Error: %v", err)
+	}
 	// Dumps all the resources in the spec namespace, then cleanups the cluster object and the spec namespace itself.
 	By(fmt.Sprintf("Dumping all the Cluster API resources in the %q namespace", namespace))
 	// Dump all Cluster API related resources to artifacts before deleting them.
 	framework.DumpAllResources(ctx, framework.DumpAllResourcesInput{
 		Lister:    clusterClient,
 		Namespace: namespace,
-		LogPath:   filepath.Join(artifactFolder, "clusters", clusterProxy.GetName(), "resources"),
+		LogPath:   filepath.Join(artifactFolder, "clusters", bootstrapClusterProxy.GetName(), "resources"),
```
mquhuy, line 145: If we have `target_clusters` log dir, shouldn't this be `bootstrap_cluster` (instead of `clusters`, we also only have 1 bootstrap cluster, I suppose?)

And do we need the cluster name?

test/e2e/healthchek.go
```diff
@@ -23,94 +25,120 @@ const (
...
+	Expect(bootstrapClusterClient.Delete(ctx, controlplaneRemediationTemplate)).To(Succeed(), "should delete controlplane Metal3RemediationTemplate CR")
+
+	By("Make sure Metal3RemediationTemplate CRs are actually deleted")
+	Eventually(func() bool {
+		cpM3MremediationTemplate := &infrav1.Metal3RemediationTemplate{}
+		workerM3MremediationTemplate := &infrav1.Metal3RemediationTemplate{}
+		cpErr := bootstrapClusterClient.Get(ctx, client.ObjectKeyFromObject(controlplaneRemediationTemplate), cpM3MremediationTemplate)
+		workerErr := bootstrapClusterClient.Get(ctx, client.ObjectKeyFromObject(workerRemediationTemplate), workerM3MremediationTemplate)
+		return apierrors.IsNotFound(cpErr) && apierrors.IsNotFound(workerErr)
+	}, 2*time.Minute, 10*time.Second).Should(BeTrue(), "Metal3RemediationTemplate should have been deleted")
```
mquhuy, line 108: Should these intervals be added to the e2e_conf.yaml instead?

test/e2e/ip_reuse_test.go
```diff
@@ -18,7 +18,7 @@ var _ = Describe("When testing ip reuse [ip-reuse] [features]", Label("ip-reuse"
 		validateGlobals(specName)
 
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
```
mquhuy, line 21: Same as above

test/e2e/md_remediations_test.go
```diff
@@ -21,7 +21,14 @@ var _ = Describe("When testing MachineDeployment remediation [healthcheck] [reme
 		Expect(osType).ToNot(Equal(""))
 		validateGlobals(specName)
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
+
+		Logf("remove existing bmh from source")
+		cmd := exec.Command("bash", "-c", "kubectl delete -f bmhosts_crs.yaml  -n metal3")
```
mquhuy, line 27: Should we use https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/common.go#L948 instead?

mquhuy (commented):
A few nits

test/e2e/basic_integration_test.go
adilGhaffarDev: Project Infra checks only this directory and copies whatever is in it. We will need to make changes there too, I can open a separate PR which will just be fixing the naming of these folders in capm3 and project-infra.

test/e2e/common.go
smoshiur1237: Here `taget_cluster_logs` log directory was used to align the previous format of log collection and from project infra it is looking for same directory and copy it from here.  

Also I think `bootstrapClusterProxy ` is just a rename of `clusterProxy`, In remediation it involves the bootstrap cluster.

adilGhaffarDev: Just "resources" is fine here we don't need to mention the cluster name. I have changed it. please check.

test/e2e/md_remediations_test.go
adilGhaffarDev: done

test/e2e/healthchek.go
mquhuy: Could you reference them here, instead of hardcoding?

adilGhaffarDev: sorry it was missed, now fixed

test/e2e/pivoting.go
```diff
@@ -423,8 +429,8 @@ func rePivoting(ctx context.Context, inputGetter func() RePivotingInput) {
 	numberOfControlplane := int(*input.E2EConfig.GetInt32PtrVariable("CONTROL_PLANE_MACHINE_COUNT"))
 	numberOfAllBmh := numberOfWorkers + numberOfControlplane
 
-	By("Fetch logs from target cluster")
-	err := FetchClusterLogs(input.TargetCluster, clusterLogCollectionBasePath)
+	By("Fetch logs from target cluster before pivot")
```
smoshiur1237, line 432:
```suggestion
	By("Fetch logs from target cluster after pivot")
```

adilGhaffarDev: done

test/e2e/md_remediations_test.go
```diff
@@ -15,13 +15,25 @@ import (
...
 		osType := strings.ToLower(os.Getenv("OS"))
 		Expect(osType).ToNot(Equal(""))
 		validateGlobals(specName)
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "clusters", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
+
+		Logf("Removing existing BMHs from source")
+		bmhData, err := os.ReadFile(filepath.Join(workDir, bmhCrsFile))
+		Expect(err).ToNot(HaveOccurred(), "No BMH CRs file not found")
```
lentzi90, line 32:
```suggestion
		Expect(err).ToNot(HaveOccurred(), "BMH CRs file not found")
```

test/e2e/common.go
```diff
@@ -125,19 +125,24 @@ func getSha256Hash(filename string) ([]byte, error) {
 	return hash.Sum(nil), nil
 }
 
-func DumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterProxy framework.ClusterProxy, artifactFolder string, namespace string, intervalsGetter func(spec, key string) []interface{}, clusterName, clusterctlLogFolder string, skipCleanup bool) {
+func DumpSpecResourcesAndCleanup(ctx context.Context, specName string, bootstrapClusterProxy framework.ClusterProxy, targetClusterProxy framework.ClusterProxy, artifactFolder string, namespace string, intervalsGetter func(spec, key string) []interface{}, clusterName, clusterctlLogFolder string, skipCleanup bool) {
```
lentzi90, line 128: I'm a bit torn about the change here since it implies that there will always be exactly one target cluster.
What if we have tests with multiple workload clusters at some point?
Not a huge issue, but something to think about. Perhaps we could instead detect all workload clusters automatically?

test/e2e/common.go
```diff
@@ -125,19 +125,24 @@ func getSha256Hash(filename string) ([]byte, error) {
...
+		Logf("Error: %v", err)
+	}
 	// Dumps all the resources in the spec namespace, then cleanups the cluster object and the spec namespace itself.
 	By(fmt.Sprintf("Dumping all the Cluster API resources in the %q namespace", namespace))
 	// Dump all Cluster API related resources to artifacts before deleting them.
 	framework.DumpAllResources(ctx, framework.DumpAllResourcesInput{
 		Lister:    clusterClient,
 		Namespace: namespace,
-		LogPath:   filepath.Join(artifactFolder, "clusters", clusterProxy.GetName(), "resources"),
+		LogPath:   filepath.Join(artifactFolder, "resources"),
```
lentzi90, line 145: Now that the function takes care of both bootstrap and workload cluster logs I think this must also be changed. I think my original idea was to call the function for all cluster proxise separately and then the name of the cluster proxy would separate the logs.
I would suggest that we at minimum put this in a `bootstrapCluster` folder. Perhaps even better to call `DumpAllResources` separately for both bootstrap and workload clusters and put the resources in separate folders.

lentzi90 (commented):
Great work! This looks like mostly solid improvements.

Added some comments inline. In general I would also encourage to avoid "target cluster" and use "workload cluster" instead. It is not a big issue, but "target cluster" can be quite confusing. If we want to be precise it is no longer a target cluster after the pivot, so continuing to call it that sounds strange.

test/e2e/common.go
adilGhaffarDev: I agree, I can open a separate PR with this change. Can we merge this one without this change? I will add the to-do to the function.

lentzi90: Yes I'm fine to merge like this and work on improving it later :slightly_smiling_face:

test/e2e/common.go
adilGhaffarDev: now dumping resources to `bootstrapCluster`

lentzi90 (commented):
/approve

mboukhalfa:
> In general I would also encourage to avoid "target cluster" and use "workload cluster" instead. It is not a big issue, but "target cluster" can be quite confusing. If we want to be precise it is no longer a target cluster after the pivot, so continuing to call it that sounds strange.

I think if we pivoted we can we still call it workload as well since it becomes management cluster !

lentzi90:
> > In general I would also encourage to avoid "target cluster" and use "workload cluster" instead. It is not a big issue, but "target cluster" can be quite confusing. If we want to be precise it is no longer a target cluster after the pivot, so continuing to call it that sounds strange.
> 
> If we pivoted can we still call it workload since it becomes management cluster !

Yes, after pivot it is a [workload cluster](https://cluster-api.sigs.k8s.io/user/concepts#workload-cluster) that is also a [management cluster](https://cluster-api.sigs.k8s.io/user/concepts#management-cluster), also known as a self hosted cluster

mboukhalfa:
> > > In general I would also encourage to avoid "target cluster" and use "workload cluster" instead. It is not a big issue, but "target cluster" can be quite confusing. If we want to be precise it is no longer a target cluster after the pivot, so continuing to call it that sounds strange.
> > 
> > 
> > If we pivoted can we still call it workload since it becomes management cluster !
> 
> Yes, after pivot it is a [workload cluster](https://cluster-api.sigs.k8s.io/user/concepts#workload-cluster) that is also a [management cluster](https://cluster-api.sigs.k8s.io/user/concepts#management-cluster), also known as a self hosted cluster

Maybe we could clarify this further in CAPI glossary, as the current definition of a [workload cluster](https://cluster-api.sigs.k8s.io/reference/glossary#workload-cluster) doesn't account for clusters running CAPI tooling, which may cause confusion when a workload cluster also becomes a management cluster post-pivot

lentzi90:
Yes we could probably add "self-hosted cluster" to the glossary. They use this terminology in the [test spec](https://github.com/kubernetes-sigs/cluster-api/blob/f1b11741f3d63356baa492fdf44662f4c7179046/test/e2e/self_hosted.go#L84) so I believe it would be easily accepted.

kashifest (commented):
/lgtm

PR #2056, opened 2024-10-28
🌱 Redundant IP claims and related RBAC policies

ksahil12:
Removed RBAC rules are not being utilized by operator and hence can be cleaned up. It includes:
1. ipclaims
2. ipaddresses
3. ipaddressclaims

Fixes #

### DIFF
--- a/controllers/metal3datatemplate_controller.go
+++ b/controllers/metal3datatemplate_controller.go
@@ -56,14 +56,6 @@ type Metal3DataTemplateReconciler struct {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=metal3datas/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=metal3dataclaims,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=metal3dataclaims/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups=ipam.metal3.io,resources=ipclaims,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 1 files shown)


tuminoid:
/ok-to-test

Please sign off the commit to pass DCO.

/cc @kashifest

Rozzii (changes requested):
I don't think this is actually correct.
/hold

adilGhaffarDev:
> I don't think this is actually correct.
> These comments are not in the file that I would expect them to be but they are also no repeated elswhere.
> Simply deleting them won't be enough, I will look into this in a bit more detail.

@ksahil12 you have to run "make generate" after removing the comments, this will update rbacs in the config. However, I think this change will break CAPM3 because we need these rbacs in Metal3Data.

Rozzii:
After looking into it a bit more, these are package level kubebuilder tags: https://book.kubebuilder.io/reference/markers/rbac.html .
Functionally nothing would change if we would move these commands to a different file, and these tags are certainly not redundant this is the only place where they are listed.

This change would only make sense if the tags would be moved to an other .go file within the package where the IPAM related resources are actually accessed  but even then it would be just a syntax change no functional change .

Also just by removing the kubebuilder tag commands without running "make generate" on the repo, the actual RBAC boilerplate coud wouldn't be generated.

adilGhaffarDev:
> After looking into it a bit more, these are package level kubebuilder tags: https://book.kubebuilder.io/reference/markers/rbac.html . Functionally nothing would change if we would move these commands to a different file, and these tags are certainly not redundant this is the only place where they are listed.
> 
> This change would only make sense if the tags would be moved to an other .go file within the package where the IPAM related resources are actually accessed but even then it would be just a syntax change no functional change .
> 
> Also just by removing the kubebuilder tag commands without running "make generate" on the repo, the actual RBAC boilerplate code wouldn't be re-generated.

Closing this in support of this

PR #2065, opened 2024-10-31
🌱Enable TLS 1.3 flag in CAPM3

smoshiur1237:
PR will add the --tls-min-version flag value to VersionTLS13 to use TLS 1.3.


tuminoid (commented):
/lgtm

PR #2083, opened 2024-11-07
🌱 Use controllerutils to add/remove finalizers

kashifest:
Using controllerutils to add and remove finalizers. Using controllerutil is thread safe, handles race conditions and has few other advantegous.
Signed-off-by: Kashif Khan <kashif.khan@est.tech>


Rozzii (commented):
/lgtm

kashifest:
> Looks good! Can we get rid of the [Filter function](https://github.com/metal3-io/cluster-api-provider-metal3/blob/d535044f8cb59eedb90dd27b25a22ac8a2577e83/baremetal/utils.go#L46) completely? I see that `Contains` is used for release notes, but `Filter` looks like we may be able to delete it.

Good point, forgot to check that

PR #2086, opened 2024-11-12
:bug: bootstrap: add support for data secret

honza:
We can configure bootstrap by setting `ConfigRef` or `DataSecretName`. Here, we correct the logic that determines if bootstrapping is ready to allow the data secret method as well as `ConfigRef`.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -171,7 +171,11 @@ func (m *MachineManager) IsProvisioned() bool {
 
 // IsBootstrapReady checks if the machine is given Bootstrap data.
 func (m *MachineManager) IsBootstrapReady() bool {
-	return m.Machine.Status.BootstrapReady
+	if m.Machine.Spec.Bootstrap.ConfigRef != nil {
+		return m.Machine.Status.BootstrapReady
+	}
+
+	return m.Machine.Spec.Bootstrap.DataSecretName != nil
 }
 
 // isControlPlane returns true if the machine is a control plane.
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -429,6 +429,8 @@ var _ = Describe("Metal3Machine manager", func() {
 		ExpectTrue bool
 	}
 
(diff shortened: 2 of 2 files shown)


Sunnatillo:
I think we should not do logic of CABPK provider. 
It should be handled [here](https://github.com/kubernetes-sigs/cluster-api/blob/1cd15431ef105cbd73d26aa8d1ec62ff89317af7/exp/internal/controllers/machinepool_controller_phases.go#L180) and we should just read the from it.

honza:
> It is already handled [here](https://github.com/kubernetes-sigs/cluster-api/blob/1cd15431ef105cbd73d26aa8d1ec62ff89317af7/exp/internal/controllers/machinepool_controller_phases.go#L180) and capm3 should just read from it.

This is referencing `MachinePool`s.

The [Machine reconciler doesn't have the same logic](https://github.com/kubernetes-sigs/cluster-api/blob/3cce0d973682f11ab0f0ba1c2522eba66dac2d91/internal/controllers/machine/machine_controller_phases.go#L188-L191) upstream.  It will never set `BootstrapReady` if `ConfigRef` is empty.

Sunnatillo:
> > It is already handled [here](https://github.com/kubernetes-sigs/cluster-api/blob/1cd15431ef105cbd73d26aa8d1ec62ff89317af7/exp/internal/controllers/machinepool_controller_phases.go#L180) and capm3 should just read from it.
> 
> This is referencing `MachinePool`s.
> 
> The [Machine reconciler doesn't have the same logic](https://github.com/kubernetes-sigs/cluster-api/blob/3cce0d973682f11ab0f0ba1c2522eba66dac2d91/internal/controllers/machine/machine_controller_phases.go#L188-L191) upstream. It will never set `BootstrapReady` if `ConfigRef` is empty.

My bad,  in that case
/approve

tuminoid:
/retest
Come on Prow ...

zaneb:
@Sunnatillo I think you were right that it _should_ work that way though. I opened a bug fix upstream: https://github.com/kubernetes-sigs/cluster-api/pull/11459
(Even if that merges there will be lots of broken CAPIs around for a while, so we still need this.)

PR #2087, opened 2024-11-12
:seedling: bump golang to 1.23.3 and golangci-lint to 1.60.3

tuminoid:
- [x] Bump Golang to 1.23.3.
- [x] Bump Golangci-lint to 1.60.3 to support Go 1.23.
    1. Get rid of deprecated golangci-lint configs
    2. add timeout to the action, otherwise it'll be timed out in 1 minute. 15 minutes is consistent with "make lint" timeout.
- [x] Fix issues coming via new golangci-lint
    1. Fix new linter errors by removing the return value which is always nil (unparam linter), and thus meaningless.                                                                              
    2. "exportloopref" is not needed from Go 1.22 onwards, replaced by "copyloopvar" in golangci-lint 1.60.2.                                                                                  
    3. Ambiguous string warnings in Wrapf fixed.
    4. Fix gosec issue about int32 overflow

Fixes: #2047

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -329,15 +329,9 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 
 	// A machine bootstrap not ready case is caught in the controller
 	// ReconcileNormal function
-	err = m.getUserDataSecretName(ctx)
-	if err != nil {
-		return err
-	}
+	m.getUserDataSecretName(ctx)
 
-	err = m.setHostLabel(ctx, host)
-	if err != nil {
-		return err
-	}
+	m.setHostLabel(ctx, host)
 
 	err = m.setHostConsumerRef(ctx, host)
 	if err != nil {
@@ -418,9 +412,9 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 // for the BareMetalHost through Metal3Machine. The UserDataSecretName might already be in a secret with
 // CABPK v0.3.0+, but if it is in a different namespace than the BareMetalHost,
 // then we need to create the secret.
-func (m *MachineManager) getUserDataSecretName(_ context.Context) error {
+func (m *MachineManager) getUserDataSecretName(_ context.Context) {
 	if m.Metal3Machine.Status.UserData != nil {
-		return nil
+		return
 	}
 
 	if m.Metal3Machine.Spec.UserData != nil {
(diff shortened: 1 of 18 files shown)


tuminoid:
Manual cherry-pick for 1.8 here: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2099

Release 1.7 does not need one as it is going out of support along with 1.9 and Golang 1.22 is supported for the rest of its life.

baremetal/metal3machine_manager.go
```diff
@@ -418,9 +412,9 @@ func (m *MachineManager) Associate(ctx context.Context) error {
 // for the BareMetalHost through Metal3Machine. The UserDataSecretName might already be in a secret with
 // CABPK v0.3.0+, but if it is in a different namespace than the BareMetalHost,
 // then we need to create the secret.
-func (m *MachineManager) getUserDataSecretName(_ context.Context) error {
+func (m *MachineManager) getUserDataSecretName(_ context.Context) {
```
tuminoid, line 415: We have two functions like this here now.

There is a choice to be made:
1. Remove the return value as PR now suggests.
2. Is there error handling missing, and the fix is to add an error condition so return value is not always nil
3. 4. Leave as is and disable linter

@kashifest @lentzi90 opinions?

test/e2e/remediation.go
```diff
@@ -136,7 +136,7 @@ func remediation(ctx context.Context, inputGetter func() RemediationInput) {
 
 	By("Testing unhealthy and inspection annotations")
 	By("Scaling down KCP to 1 replica")
-	newReplicaCount := 1
+	newReplicaCount := int32(1)
```
tuminoid, line 139: This is the lone test case that has variable instead of just harcoded number. 

It is questionable why it is int32 param in the first place, why not uint32 or just uint? If we change that, its going to be follow-up tho.

tuminoid (commented):
Left some self-comments on the PR that we should consider.

baremetal/metal3machine_manager.go
lentzi90: This (1) looks correct to me.
(The function did not do at all what I expected based on the name and description though :see_no_evil: )

lentzi90 (commented):
/approve

baremetal/metal3machine_manager.go
tuminoid: Yeah, it was surprising one. Some tech debt to pay here.

kashifest (commented):
/lgtm

tuminoid:
If this aborts again due to timeout, I'll override it, as the log is filled with unrelated Jenkins agent issues and network flakes. Ubuntu has passed long time ago.

PR #2110, opened 2024-11-26
 🌱 Use wget while downloading image

kashifest:
Recently we have experienced net.OpError error or http2streamerror error while downloading image quite frequently. This PR uses wget instead which is more resilient to http/network issues. We should change it to httpGet when network improves again.


kashifest:
> Just noting that implementing Go with HTTP/1 might be worth a try. Wget on Ubuntu is still not built with HTTP/2 by default, so it might be more robust choice for Go client as well. Go by default is using HTTP/2 and as you wrote on the summary, seems to be source of the issues.

Yep for now, lets take the CI back, I will do a followup soon

tuminoid (commented):
/lgtm

PR #2135, opened 2024-12-05
:bug: Use GOPROXY to download releases

peppi-lotta:
Some e2e CI test are experiencing failure due to 
`./scripts/ci-e2e.sh: line 82: CAPM3RELEASEPATH: unbound variable`
`./scripts/ci-e2e.sh: line 92: CAPIRELEASEPATH: unbound variable`

script/ci-e2e.sh is using function `get_latest_release` that used to be set in the Metal-dev-env code base. In [this PR ](https://github.com/metal3-io/metal3-dev-env/pull/1474). Moving to use the function `get_latest_release_from_goproxy` that replaced the function should fix the problem. This function is no longer in CAPM3 release 1.8 but need to be patched to  releases 1.6 and 1.7.

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -9,6 +9,8 @@ export WORKING_DIR=/opt/metal3-dev-env
 FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
 
 export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
+CAPIGOPROXY="${CAPIGOPROXY:-https://proxy.golang.org/sigs.k8s.io/cluster-api/@v/list}"
+CAPM3GOPROXY="${CAPM3GOPROXY:-https://proxy.golang.org/github.com/metal3-io/cluster-api-provider-metal3/@v/list}"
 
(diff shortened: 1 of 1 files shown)


scripts/ci-e2e.sh
```diff
@@ -9,6 +9,10 @@ export WORKING_DIR=/opt/metal3-dev-env
 FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
 
 export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
+export CAPI_BASE_URL="${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}"
+export CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
```
kashifest, line 13: can't we just do this since it will have the default value from previous line? 
```suggestion
export CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL}/releases}"
```

kashifest (commented):
/approve

Rozzii (commented):
looks okay now
/hold cancel

PR #2136, opened 2024-12-05
:bug: Set unbound env variables 1.7

peppi-lotta:
**What this PR does / why we need it**: script/ci-e2e.sh is using environment variables CAPIRELEASEPATH and CAPM3RELEASEPATH that used to be set in the Metal-dev-env code base. In https://github.com/metal3-io/metal3-dev-env/pull/1474CAPIRELEASEPATH was removed which leads to the variable being unbound. CAPM3RELEASEPATH has been removed somewhere along the way as-well.


kashifest (commented):
/approve

kashifest:
It seems like the issue is still visible?
/approve cancel

kashifest:
Test has moved passed the problematic zone, so the fix works.

/approve

tuminoid (commented):
/retitle 🐛 Use GOPROXY to download releases
/lgtm

PR #2156, opened 2024-12-12
🌱 Removes depreacted klogr.New() and rand.Seed()

Sunnatillo:
As of Go 1.20 there is no reason to call Seed with a random value 
klogr.New()  is deprecated, klog.Background will automatically use the right logger

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -32,7 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
-	"k8s.io/klog/v2/klogr"
+	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


zhouhao3:
It's better to modify controllers/suite_test.go as well.

Sunnatillo:
> It's better to modify controllers/suite_test.go as well.

Can you elaborate your suggestion please?

zhouhao3:
> Can you elaborate your suggestion please?

There is the same usage in controllers/suite_test.go(klogr.New()), which should also be changed.

Sunnatillo:
> > Can you elaborate your suggestion please?
> 
> There is the same usage in controllers/suite_test.go(klogr.New()), which should also be changed.

Oh, I missed it. Thank you for pointing out. Fixed now.

kashifest (commented):
/approve

PR #2161, opened 2024-12-17
:seedling: Add release github workflow

mquhuy:
Now that the release workflow is working for IPAM, we could add it to CAPM3.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -7,18 +7,39 @@ on:
     - 'release-*'
     tags:
     - 'v*'
+  workflow_run:
+    workflows:
+    - "Create_Release"
+    types:
+    - completed
 
 permissions:
   contents: read
 
 jobs:
+  set_ref:
+    runs-on: ubuntu-latest
+    outputs:
+      github_ref: ${{ steps.set_ref.outputs.github_ref }}
+    steps:
+    - name: Set ref
+      id: set_ref
+      run: |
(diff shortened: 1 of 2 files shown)


tuminoid:
We were supposed to add it after 1.9 so we can iron out all the issues to avoid fixing multiple repos. I don't think we have a single clean run in IPAM yet, nor have done patch releases with it etc.
/hold

mquhuy:
> We were supposed to add it after 1.9 so we can iron out all the issues to avoid fixing multiple repos. I don't think we have a single clean run in IPAM yet, nor have done patch releases with it etc.
> 
> We also agreed to make reusable workflow in project-infra before taking it in other repos. /hold

My intention is to have this in if we have IPAM release working. In case it doesn't work out, we should fix this PR and then test it out. IMO without taking this into use in other repos, it will take a long time to have a fully working solution.

The current setup in IPAM only fails in the build image step, which can be relatively easy to patch.

tuminoid:
> The current setup in IPAM only fails in the build image step, which can be relatively easy to patch.

Can we then patch it right now? We'll have IPAM 1.9.1 tomorrow (after CAPI 1.9.1 bump and k8s 1.32.0 bump), so thats opportunity to fix both image build and test patch release.  If those pass, then we can either generalize the workflow and then it in CAPM3 and BMO as well. If it still doesn't work, I don't want to merge broken automation and then fix it three times in different repos many times.

tuminoid:
This has been completed in another PR.
/close

PR #2163, opened 2024-12-18
:bug: bump x/crypto to v0.31.0

tuminoid:
https://osv.dev/GO-2024-3321
It is not called, but appears in scanning, so closing it off.


kashifest (commented):
/approve

PR #2201, opened 2024-12-27
🌱 Remove CP scaling from e2e-feature-test

adilGhaffarDev:
Tests failing with following error after 1.9 because MachineSetPreflightChecks is moved to beta and the default value for it is set to true, before it was false:
```
I1227 11:18:04.605619    1 machineset_preflight.go:139] "Scale up on hold because MachineSet version (1.32.0) and ControlPlane version (1.31.2) do not conform to the kubernetes version skew policy as MachineSet version is higher than ControlPlane version (\"KubernetesVersionSkew\" preflight check failed); MachineSet version (1.32.0) and ControlPlane version (1.31.2) do not conform to kubeadm version skew policy as kubeadm only supports joining with the same major+minor version as the control plane (\"KubeadmVersionSkew\" preflight check failed). The operation will continue after the preflight check(s) pass" controller="machineset" controllerGroup="cluster.x-k8s.io" controllerKind="MachineSet" MachineSet=
```
 in some tests, we are upgrading MDs only and not KCPs which is causing this issue.

This PR is fixing the test by also upgrading KCPs. Now flow of test is like that:
- 3 KCP, 2 worker 1.31 k8s
- Upgrade KCP to 1.32
- From IPPOOL fetch allocated IPs so we check them later
- Upgrade worker nodes to 1.32
- Check New IPs with old ones they should be the same

In addition to that this PR is also moving CreateTargetCluster to common.go

Fixes #

### DIFF
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -78,6 +78,13 @@ if [[ ${GINKGO_FOCUS:-} == "integration" || ${GINKGO_FOCUS:-} == "basic" ]]; the
   export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"1"}
 fi
 
+# IPReuse feature test environment vars and config
+if [[ ${GINKGO_FOCUS:-} == "features" && ${GINKGO_SKIP:-} == "pivoting remediation" ]]; then
+  export NUM_NODES="5"
+  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"3"}
+  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"2"}
+fi
+
 # Exported to the cluster templates
 # Generate user ssh key
 if [ ! -f "${HOME}/.ssh/id_rsa" ]; then
--- a/test/e2e/basic_integration_test.go
+++ b/test/e2e/basic_integration_test.go
@@ -23,7 +23,21 @@ var _ = Describe("When testing basic cluster creation [basic]", Label("basic"), 
 		By("Fetching cluster configuration")
 		k8sVersion := e2eConfig.GetVariable("KUBERNETES_VERSION")
 		By("Provision Workload cluster")
-		targetCluster, _ = createTargetCluster(k8sVersion)
+		targetCluster, _ = CreateTargetCluster(ctx, func() CreateTargetClusterInput {
+			return CreateTargetClusterInput{
+				E2EConfig:             e2eConfig,
+				BootstrapClusterProxy: bootstrapClusterProxy,
+				SpecName:              specName,
+				ClusterName:           clusterName,
+				K8sVersion:            k8sVersion,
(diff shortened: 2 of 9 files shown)


scripts/environment.sh
```diff
@@ -78,6 +78,13 @@ if [[ ${GINKGO_FOCUS:-} == "integration" || ${GINKGO_FOCUS:-} == "basic" ]]; the
   export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"1"}
 fi
 
+# IPReuse feature test environment vars and config
+if [[ ${GINKGO_FOCUS:-} == "features" && ${GINKGO_SKIP:-} == "pivoting remediation" ]]; then
```
mquhuy, line 82: This looks weird. "GINKGO_FOCUS == features" would render "GINKGO_SKIP" unnecessary. Did you mean to use || instead of &&?

mquhuy (commented):
I haven't checked the code very carefully, but based on your PR comment, can't the issue be solved by `export EXP_MACHINE_SET_PREFLIGHT_CHECKS = false`?

Also, an inline comment. That line seems not related to the fix though :-?

scripts/environment.sh
adilGhaffarDev: We following tests under GINKGO_FOCUS == features

- [ip-reuse] [features]
- [healthcheck] [remediation] [features]
- [pivoting] [features]
- [remediation] [features]

to set env variables only for IPReuse I need to make sure GINKGO_FOCUS == features && GINKGO_SKIP == pivoting remediation
-e2e-feature-test-main-features is only running IPReuse, perhaps we should change the name of the test.

adilGhaffarDev:
> I haven't checked the code very carefully, but based on your PR comment, can't the issue be solved by `export EXP_MACHINE_SET_PREFLIGHT_CHECKS = false`?

yes, but I think we should align our tests as close to how they should be in a production environment. Previously in this test, we were only upgrading worker nodes and skipping KCP which I think seems wrong.

mquhuy:
> > I haven't checked the code very carefully, but based on your PR comment, can't the issue be solved by `export EXP_MACHINE_SET_PREFLIGHT_CHECKS = false`?
> 
> yes, but I think we should align our tests as close to how they should be in a production environment. Previously in this test, we were only upgrading worker nodes and skipping KCP which I think seems wrong.

I'm not opposing to that, but in that case let's edit the PR title and content to reflect your reasons.

adilGhaffarDev:
> I'm not opposing to that, but in that case let's edit the PR title and content to reflect your reasons.

done

mquhuy:
> > I'm not opposing to that, but in that case let's edit the PR title and content to reflect your reasons.
> 
> done

I mean the title should be something like "Remove CP scaling from e2e-feature-test", as that's the main reason for this change.

adilGhaffarDev:
cc @Rozzii

Rozzii (commented):
/lgtm

Issue #2112, opened 2024-11-27
Adapt e2e tests to account for Ironic deployed by IrSO


dtantsur:
Currently, the tests assume that Ironic is either started locally or is deployed by BMO's deploy.sh. When deployed by IrSO, it looks differently, so at least re-pivoting tests fail.

/kind ci
/help

tuminoid:
We also need some keywords to run tests with IRSO. Create https://github.com/metal3-io/project-infra/issues/920 in project infra for that.

PR #2203, opened 2024-12-30
🌱 e2e: account for ironic-standalone-operator in pivoting tests

dtantsur:
BMO's deploy.sh is still used after pivoting. This should be addressed
separately.

Fixes: #2112
Signed-off-by: Dmitry Tantsur <dtantsur@protonmail.com>


test/e2e/pivoting.go
```diff
@@ -340,24 +344,41 @@ func installIronicBMO(ctx context.Context, inputGetter func() installIronicBMOIn
 	}
 }
 
+type DeploymentType string
```
adilGhaffarDev, line 347:
```suggestion
type IronicDeploymentType string
```

test/e2e/pivoting.go
```diff
@@ -340,24 +344,41 @@ func installIronicBMO(ctx context.Context, inputGetter func() installIronicBMOIn
...
+
+const (
+	DeploymentTypeLocal DeploymentType = "local"
+	DeploymentTypeBMO   DeploymentType = "deploy.sh"
+	DeploymentTypeIrSO  DeploymentType = "irso"
```
adilGhaffarDev, line 352: There is one more deployment type, we are deploying Ironic using Kustomize in the `clusterctl upgrade tests` and not relying on BMO `deploy.sh`, as seen in the following code: https://github.com/metal3-io/cluster-api-provider-metal3/blob/6cad2fb3e1ab8274ef077c8476b75226e95e2f7b/test/e2e/upgrade_clusterctl_test.go#L236.

The primary reason for this approach was to simplify the process of installing different versions of Ironic. The plan is to use the same approach in other tests and get rid of `deploy.sh` but I haven't had the chance to work on it yet. I'll try to address this soon.

I’d like to ask how the `IrSO` manages the versioning of Ironic deployments. If it's straightforward(easy to configure), we could potentially eliminate `Kustomize` and rely solely on `IrSO`.

adilGhaffarDev (commented):
Thanks for working on it. I think this will be a great addition and simplify our tests. 
I have a question and a small suggestion, but overall, it looks good.

dtantsur:
> Are you planning to add deployment with IrSO in CAPM3 too?

Absolutely. My final goal is to have IrSO as the main way to deploy Ironic. But I'm taking small steps because I'm completely unfamiliar with the e2e testing code.

test/e2e/pivoting.go
dtantsur: Will change

test/e2e/pivoting.go
dtantsur: Do you suggest to rename `deploy.sh` to something else or to create a new deployment type?

> I’d like to ask how the IrSO manages the versioning of Ironic deployments.

The `Ironic` CRD has a `version` field which chooses the branch of ironic-image (e.g. `27.0`) or `latest`.

adilGhaffarDev: > Do you suggest to rename `deploy.sh` to something else or to create a new deployment type?

Create a new type:
`	DeploymentTypeKustomize IronicDeploymentType = "kustomize"`
Its fine not add it now, I need to do some refactoring I can add it then.



> The `Ironic` CRD has a `version` field which chooses the branch of ironic-image (e.g. `27.0`) or `latest`.

Great once this PR is merged, I will add ironic deployment IrSO in CAPM3.

adilGhaffarDev:
> Absolutely. My final goal is to have IrSO as the main way to deploy Ironic. But I'm taking small steps because I'm completely unfamiliar with the e2e testing code.

Great, I can help with that.

test/e2e/common.go
```diff
@@ -61,6 +61,8 @@ const (
 	osTypeCentos           = "centos"
 	osTypeUbuntu           = "ubuntu"
 	ironicSuffix           = "-ironic"
+	// Credit for existence of this variable goes to golangci-lint.
+	True = "true"
```
mquhuy, line 65: Instead of having this hack, maybe you could add something similar to this
https://github.com/metal3-io/baremetal-operator/blob/main/test/e2e/e2e_config.go#L161

test/e2e/pivoting.go
```diff
@@ -109,14 +110,16 @@ func pivoting(ctx context.Context, inputGetter func() PivotingInput) {
 	Expect(er).ToNot(HaveOccurred(), "Cannot fetch target cluster kubeconfig")
 
 	By("Remove Ironic containers from the source cluster")
-	isIronicDeployment := true
+	ironicDeploymentType := DeploymentTypeBMO
 	if ephemeralCluster == Kind {
-		isIronicDeployment = false
+		ironicDeploymentType = DeploymentTypeLocal
+	} else if os.Getenv("USE_IRSO") == True {
```
mquhuy, line 116: Use `e2e_config.GetVariable()` instead of `os.Getenv()`

mquhuy: Also, consider using `switch case` for better readability.

dtantsur: Is it the same thing? I see some places use both `os.GetEnv` and `GetVariable`.

dtantsur: Mmm, looking at the source code, I cannot use it. GetVariable expects it to be always present, USE_IRSO is not.

dtantsur: I guess I can put it in `e2e_conf.yaml`. Will it work?

mquhuy: Yes, you can put a default value in e2e_config. It will also prioritize the env var if it was exported before.

test/e2e/pivoting.go
```diff
@@ -340,16 +344,24 @@ func installIronicBMO(ctx context.Context, inputGetter func() installIronicBMOIn
 	}
 }
 
+type IronicDeploymentType string
+
+const (
+	IronicDeploymentTypeLocal     IronicDeploymentType = "local"
+	IronicDeploymentTypeKustomize IronicDeploymentType = "kustomize"
```
adilGhaffarDev, line 351:
```suggestion
	IronicDeploymentTypeKustomize IronicDeploymentType = "kustomize"
	IronicDeploymentTypeBMO IronicDeploymentType = "deploy.sh"
```
Lets put both options here later once irso is working we can decide what to remove.

dtantsur: I don't understand the difference between Kustomize and deploy.sh. Is it not the same thing? Where do we use what?

adilGhaffarDev: Sorry my bad I checked the code again, IronicDeploymentType is only used to remove ironic that is deployed by metal3-dev-env, so it's either going to be IrSO or deploy.sh, we don't need Kustomize here. Let's keep `IronicDeploymentTypeBMO IronicDeploymentType = "deploy.sh"` and remove `IronicDeploymentTypeKustomize IronicDeploymentType = "kustomize"`. 
Sorry for the confusion, we do use Kustomize to deploy ironic in some tests but that is not related to this flow.

PR #2237, opened 2025-01-13
🌱 bump x/net to v0.34.0

tuminoid:
Shuts up vulnerability scanners about [x/net vuln](https://osv.dev/GHSA-w32m-9786-jp63). Vulnerable path is not called in CAPM3.


Rozzii (commented):
/lgtm

PR #2238, opened 2025-01-13
✨ Remove the code which removes M3M owner reference from BMH 

kashifest:
We have removed the code which sets M3M owner reference in v1.10 release cycle. Now we also remove the code which removes the owner reference from BMH.

This PR should land in v1.11 release cycle.

Signed-off-by: Kashif Khan <kashif.khan@est.tech>


kashifest:
/hold
This PR should land in v1.11 release cycle 

This is currently rebased on top of https://github.com/metal3-io/cluster-api-provider-metal3/pull/1742 which should land in v1.10.

Issue #1357, opened 2023-12-12
Metal3DataTemplate with templateReference cannot be deleted while Metal3Data with matching name exists


lentzi90:
**What steps did you take and what happened:**

1. Create a Cluster with Metal3 for infra and make sure it has a MachineDeployment that uses a Metal3DataTemplate with a `templateReference`.
2. Create another Metal3DataTemplate and Metal3MachineTemplate that references it.
3. Update the MachineDeployment to use the new Metal3MachineTemplate and Metal3DataTemplate so that a rollout is triggered.
4. When the rollout is complete, delete the old Metal3DataTemplate that is no longer used.
5. **The Metal3DataTemplate cannot be deleted because it thinks it is still in use**

**What did you expect to happen:**

Old Metal3DataTemplates should be possible to delete when they are no longer used.

**Anything else you would like to add:**

The issue happens only when using `templateReference`. It is unclear to me what the function of this field is.
I have been able to do rollouts without issues when not using the `templateReference` and can then delete the old Metal3DataTemplate.

**Environment:**

- Cluster-api version: v1.5.3
- Cluster-api-provider-metal3 version: v1.5.2
- Environment (metal3-dev-env or other): other
- Kubernetes version: (use `kubectl version`): v1.28.3

/kind bug

peppi-lotta:
# templateReference investigation

## Questions to answer

- Is templateReference actually needed for anything?
- Does having templateReference prevent deletion of resources in any situation where it shouldn’t?
- Do we fix the bug or remove templateReference all together?

## Beginning

| Resource           |  Name                    | templateReference to  |
|--------------------|--------------------------|-----------------------|
| Metal3DataTemplate | test1-workers-template   | -                     |
| Metal3Data         | test1-workers-template-0 | 🠋                     |

```
kubectl get metal3data -n metal3 test1-workers-template-0 -o yaml
...
spec:
  claim:
    name: test1-qtmqg-5br6d
    namespace: metal3
  metaData:
    name: test1-qtmqg-5br6d-metadata
    namespace: metal3
  networkData:
    name: test1-qtmqg-5br6d-networdata
    namespace: metal3
  template:
    name: test1-workers-template
    namespace: metal3
```

New secrets were created. Metal3DataTemplate test1-workers-template can not be deleted because it is in use.

## Do rolling update where new Metal3DataTemplate's templateReference is set to current Metal3DataTemplate.

| Resource           |  Name                    | templateReference to   |
[...]

PR #190, opened 2021-04-08
:sparkles: Add unit test for TemplateReference

kashifest:

PR #184, opened 2021-03-26
Add templateReference logic in Metal3DataTemplate

kashifest:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2265, opened 2025-01-23
:warning: Remove metal3datatemplate template reference

peppi-lotta:
**What this PR does / why we need it**: This reverts all logic changes made in https://github.com/metal3-io/cluster-api-provider-metal3/pull/184 https://github.com/metal3-io/cluster-api-provider-metal3/pull/190 and any other mention of the templateReference. It's unclear if there is any use cases for templateReference. If there are none it should be deleted. 

Deprecation notice: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2326

I'm leaving the field in the API. That will be remove in v1beta2

Fixes # https://github.com/metal3-io/cluster-api-provider-metal3/issues/1357


Rozzii:
@peppi-lotta I have added this to the 1.11 milestone and to the backlog to have a better visibility.
Could you also please rebase and write a short update about the state of this PR?

Pinging some folks for who this might be relevant to take a look:
@kashifest @zaneb @lentzi90

Rozzii (commented):
I will come back to review after the CI is passing.

peppi-lotta:
I'm now running all kind of test to see if removal of templateReference breaks any of the tests. 

My manual tests seem to indicate that everything works the same with or without the templateReference. The manual testing showed that by setting the templateReference the name of a m3d object can stay the same after upgrade. I figure this could be useful in some case to keep the m3d objects name the same. The test is not testing this though. 

The test seems to just query whether there is a m3d object with a templateReference in general. There is no test regarding whether the template reference is correct in any way. I'm not sure what would be the "correct" way to use the TemplateReference as there is no clear use cases in the upstream. 

If none of the test fail due to this change I would say we can remove this feature.

zaneb:
Unfortunately I don't think you can ever remove a field from an existing CRD (except when bumping the version) without breaking users. This API is at v1beta1 already and the field has been there since 2021, so I think this has to wait for v1beta2.

lentzi90:
Hmm that is a good point. But we can still remove the logic and ignore the field if it anyway doesn't do anything. Then removing it i v1beta2 would be just cosmetics

peppi-lotta:
That's a good point @zaneb. Thank you for bringing this up.

Based on this I'm thinking 
  - Deprecation notice in release 1.10 
  - Logic could be removed 1.11 but leave the API field (Because everything seems to be working the same with or without this value)
  - The API field can be removed in v1beta2 then.

peppi-lotta:
/hold
I'll put this on hold for now. Let's get this merged this once 1.10 is out like I suggested in my last comment.

peppi-lotta:
Release 1.10 is published so let's get this in and it will get max amount of testing before 1.11. 
/hold cancel 

/test metal3-centos-e2e-basic-test-main
/test metal3-centos-e2e-feature-test-main-remediation
/test metal3-centos-e2e-feature-test-main-scalability
/test metal3-e2e-1-32-1-33-upgrade-test-main
/test metal3-ubuntu-e2e-basic-test-main
/test metal3-ubuntu-e2e-feature-test-main-features

Rozzii:
/hold cancel
scalability is not relevant here as much that we should hold this, and the CI is greener now.

tuminoid:
This was last pushed like a month ago, please rebase  and rerun tests.

tuminoid:
These remediation failures are not related, they're present in periodics as well, same with scalability.

tuminoid (commented):
/approve

PR #2267, opened 2025-01-23
:bug: Fix live-iso serial console boot verification in e2e 

Rozzii:
This commit:
 - Changes the target string looked for via the serial logs by the ginkgo live-iso test code to verify that the live iso image has been booted properly.
 - Removes the redundant use of the "cat" command from the boot verification process in the live-iso test
 -  Changes the name of the live-iso test image to the new image that is uefi compatible.

This change is needed because the release1.8 e2e feature test are failing the live-iso boot verification even when the live-iso boot menu is visible in the serial logs.

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -67,7 +67,7 @@ source "${M3_DEV_ENV_PATH}/lib/ironic_basic_auth.sh"
 source "${M3_DEV_ENV_PATH}/lib/ironic_tls_setup.sh"
 
 # image for live iso testing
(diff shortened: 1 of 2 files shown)


Rozzii:
/test metal3-centos-e2e-integration-test-release-1-8
Networking flake (well known TCP/IP issue)

tuminoid (commented):
/hold

We'll rename the images in ARM to reflect this change, so it isn't as invisible and add the LIVE_ISO_IMAGE update this this PR.

Rozzii:
> /hold
> 
> We'll rename the images in ARM to reflect this change, so it isn't as invisible and add the LIVE_ISO_IMAGE update this this PR.

Done.

PR #2271, opened 2025-01-23
:bug: Fix: pulling unused images in the middle of tests

peppi-lotta:
**What this PR does / why we need it**: Let's take `metal3-periodic-ubuntu-e2e-integration-test-release-1-9` as an example. This test runs `make` in `metal3-dev-env` where container images (with correct tags) get pulled. Some time later the test moves into `cluster-api-provider-metal3` folder and runs `make e2e-test`. At this time CAPM3, BMO and IPAM images will get pulled again but this time the tag is set to latest. This is unnecessary. This fix will pull the correct images if they are missing instead of always bulling images that are unused in many cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,8 +159,6 @@ ARTIFACTS ?= $(ROOT_DIR)/_artifacts
 E2E_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/e2e_conf.yaml
 E2E_OUT_DIR ?= $(ROOT_DIR)/test/e2e/_out
 E2E_CONF_FILE_ENVSUBST ?= $(E2E_OUT_DIR)/$(notdir $(E2E_CONF_FILE))
-E2E_CONTAINERS ?= quay.io/metal3-io/cluster-api-provider-metal3 quay.io/metal3-io/baremetal-operator quay.io/metal3-io/ip-address-manager
-
 SKIP_CLEANUP ?= false
 EPHEMERAL_TEST ?= false
 SKIP_CREATE_MGMT_CLUSTER ?= true
@@ -211,10 +209,6 @@ e2e-tests: CONTAINER_RUNTIME?=docker # Env variable can override this default
 export CONTAINER_RUNTIME
 
 e2e-tests: $(GINKGO) e2e-substitutions cluster-templates # This target should be called from scripts/ci-e2e.sh
-	for image in $(E2E_CONTAINERS); do \
-		$(CONTAINER_RUNTIME) pull $$image; \
-	done
-
 	$(GINKGO) --timeout=$(GINKGO_TIMEOUT) -v --trace --tags=e2e  \
 		--show-node-events --no-color=$(GINKGO_NOCOLOR) \
 		--junit-report="junit.e2e_suite.1.xml" \
@@ -232,10 +226,6 @@ e2e-clusterclass-tests: CONTAINER_RUNTIME?=docker # Env variable can override th
 export CONTAINER_RUNTIME
 
 e2e-clusterclass-tests: $(GINKGO) e2e-substitutions clusterclass-templates # This target should be called from scripts/ci-e2e.sh
-	for image in $(E2E_CONTAINERS); do \
-		$(CONTAINER_RUNTIME) pull $$image; \
-	done
-
 	$(GINKGO) --timeout=$(GINKGO_TIMEOUT) -v --trace --tags=e2e  \
 		--show-node-events --no-color=$(GINKGO_NOCOLOR) \
 		--junit-report="junit.e2e_suite.1.xml" \


Makefile
```diff
@@ -159,7 +159,10 @@ ARTIFACTS ?= $(ROOT_DIR)/_artifacts
 E2E_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/e2e_conf.yaml
 E2E_OUT_DIR ?= $(ROOT_DIR)/test/e2e/_out
 E2E_CONF_FILE_ENVSUBST ?= $(E2E_OUT_DIR)/$(notdir $(E2E_CONF_FILE))
-E2E_CONTAINERS ?= quay.io/metal3-io/cluster-api-provider-metal3 quay.io/metal3-io/baremetal-operator quay.io/metal3-io/ip-address-manager
+capm3_image = "${CONTAINER_REGISTRY}/metal3-io/cluster-api-provider-metal3:${CAPM3RELEASE:-latest}"
```
tuminoid, line 162: Why put them in lowercase?

Makefile
```diff
@@ -159,7 +159,10 @@ ARTIFACTS ?= $(ROOT_DIR)/_artifacts
 E2E_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/e2e_conf.yaml
 E2E_OUT_DIR ?= $(ROOT_DIR)/test/e2e/_out
 E2E_CONF_FILE_ENVSUBST ?= $(E2E_OUT_DIR)/$(notdir $(E2E_CONF_FILE))
-E2E_CONTAINERS ?= quay.io/metal3-io/cluster-api-provider-metal3 quay.io/metal3-io/baremetal-operator quay.io/metal3-io/ip-address-manager
+CAPM3_IMAGE ?= ${CONTAINER_REGISTRY}/metal3-io/cluster-api-provider-metal3:${CAPM3RELEASE:-latest}
```
tuminoid, line 162: Who initially defines CONTAINER_REGISTRY here?

In CAPM3, it is only mentioned in e2e_conf.yaml, so it is probably empty here, unless it was already defined in dev-env first. With CAPM3 e2e, the test cycle starts on CAPM3 side, not from dev-env, so does this actually work?

In this Makefile, registry variables defined are REGISTRY, STAGING_REGISTRY and PROD_REGISTRY?

tuminoid (commented):
Question:

tuminoid:
There also seems to be some actual issue triggered by this:

```text
[2025-01-24T07:05:44.054Z] for image in registry.nordix.org/quay-io-proxy/metal3-io/cluster-api-provider-metal3: registry.nordix.org/quay-io-proxy/metal3-io/ip-address-manager: "registry.nordix.org/quay-io-proxy/metal3-io/baremetal-operator:; do \
[2025-01-24T07:05:44.054Z] 	podman pull $image; \
[2025-01-24T07:05:44.054Z] done
[2025-01-24T07:05:44.054Z] bash: -c: line 1: unexpected EOF while looking for matching `"'
[2025-01-24T07:05:44.054Z] bash: -c: line 4: syntax error: unexpected end of file
```

https://jenkins.nordix.org/blue/organizations/jenkins/metal3-centos-e2e-integration-test-main/detail/metal3-centos-e2e-integration-test-main/884/pipeline

Makefile
peppi-lotta: CONTAINER_REGISTRY can be defined in project-infra (and is defined there for e2e-integration-tests) or gets a default value from dev-env. I can change `${CONTAINER_REGISTRY} `to `${CONTAINER_REGISTRY:quay.io}` just to be sure. 

I've been looking at [e2e-integration-test on release-1.9 ](https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-ubuntu-e2e-integration-test-release-1-9/40/consoleFull) when making this change.

Here are some copied from the log I've linked above. Times are system time. As you can see from here, yes we starts from CAPM3 e2e but the tests very quickly moves to execute in dev-env. Prepulling of images happens there and we come back to CAPM3 later where the some images get pulled again with latest tag (which is not correct for release-1.9 tests). 
```
01:31:43  Running the tests
01:31:43  + make test-e2e
...
01:31:45  + pushd /opt/metal3-dev-env/metal3-dev-env
01:31:45  /opt/metal3-dev-env/metal3-dev-env ~/tested_repo
01:31:45  + make
01:31:45  make[1]: Entering directory '/opt/metal3-dev-env/metal3-dev-env'
...
01:37:25  ++ sudo docker pull registry.nordix.org/quay-io-proxy/metal3-io/baremetal-operator:v0.9.0
01:37:27  v0.9.0: Pulling from quay-io-proxy/metal3-io/baremetal-operator
...
01:56:45  + make e2e-tests
01:56:45  make[1]: Entering directory '/home/metal3ci/tested_repo'
...
01:56:50  Using default tag: latest
01:56:52  latest: Pulling from metal3-io/baremetal-operator
01:56:52  dd5ad9c9c29f: Already exists
01:56:52  960043b8858c: Already exists
01:56:52  b4ca4c215f48: Already exists
01:56:52  eebb06941f3e: Already exists
01:56:52  02cd68c0cbf6: Already exists
01:56:52  d3c894b5b2b0: Already exists
01:56:52  b40161cd83fc: Already exists
01:56:52  46ba3f23f1d3: Already exists
01:56:52  4fa131a1b726: Already exists
01:56:52  fc8b0dbaf292: Pulling fs layer
01:56:53  fc8b0dbaf292: Verifying Checksum
01:56:53  fc8b0dbaf292: Download complete
01:56:54  fc8b0dbaf292: Pull complete
```

tuminoid: Note that this is Makefile and not bash, so the variable syntax is not `{}` but `()`.

Also you need to make fallbacks differently.

```suggestion
CAPM3RELEASE ?= latest
CAPM3_IMAGE ?= $(CONTAINER_REGISTRY)/metal3-io/cluster-api-provider-metal3:$(CAPM3RELEASE)
```

Makefile
```diff
@@ -159,7 +159,10 @@ ARTIFACTS ?= $(ROOT_DIR)/_artifacts
...
+IPAM_IMAGE ?= ${CONTAINER_REGISTRY}/metal3-io/ip-address-manager:${IPAMRELEASE:-latest}
```
tuminoid, line 163: same

Makefile
```diff
@@ -159,7 +159,10 @@ ARTIFACTS ?= $(ROOT_DIR)/_artifacts
...
+BMO_IMAGE ?= ${CONTAINER_REGISTRY}/metal3-io/baremetal-operator:${BARE_METAL_OPERATOR_TAG:-latest}
```
tuminoid, line 164: same

Makefile
```diff
@@ -159,7 +159,15 @@ ARTIFACTS ?= $(ROOT_DIR)/_artifacts
 E2E_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/e2e_conf.yaml
 E2E_OUT_DIR ?= $(ROOT_DIR)/test/e2e/_out
 E2E_CONF_FILE_ENVSUBST ?= $(E2E_OUT_DIR)/$(notdir $(E2E_CONF_FILE))
-E2E_CONTAINERS ?= quay.io/metal3-io/cluster-api-provider-metal3 quay.io/metal3-io/baremetal-operator quay.io/metal3-io/ip-address-manager
+
+CONTAINER_REGISTRY ?= quay.io
+CAPM3RELEASE ?= latest
```
tuminoid, line 164: Group the tag variables and image variables. Makes this more readable.

We also need to cherry-pick this to release branches to have any effect there.

tuminoid (commented):
/approve

mquhuy:
This won't work. For e.g. check https://jenkins.nordix.org/blue/rest/organizations/jenkins/pipelines/metal3-centos-e2e-integration-test-main/runs/887/nodes/21/steps/35/log/?start=0

```
export CAPM3_IMAGE=registry.nordix.org/quay-io-proxy/metal3-io/cluster-api-provider-metal3:main
```

=> This is devenv's default. It should have taken into use the value we define here in CAPM3 e2e (tag should be `latest`).

These values should have been exported inside [ci-e2e.sh](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/scripts/ci-e2e.sh), also please check in that file those default values, like the  CAPM3RELEASE, which is defined based on the CAPM3RELEASEBRANCH (same redefinition will happen inside devenv), so the redefinition we do here won't likely take any effects.

mquhuy:
/hold 
until the issue above is sorted out.

peppi-lotta:
> This won't work. For e.g. check https://jenkins.nordix.org/blue/rest/organizations/jenkins/pipelines/metal3-centos-e2e-integration-test-main/runs/887/nodes/21/steps/35/log/?start=0
> 
> ```
> export CAPM3_IMAGE=registry.nordix.org/quay-io-proxy/metal3-io/cluster-api-provider-metal3:main
> ```
> 
> => This is devenv's default. It should have taken into use the value we define here in CAPM3 e2e (tag should be `latest`).
> 
> These values should have been exported inside [ci-e2e.sh](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/scripts/ci-e2e.sh), also please check that there are some default values, like the CAPM3RELEASE, which is defined based on the CAPM3RELEASEBRANCH (same redefinition will happen inside devenv), so the redefinition of that var we do here won't likely take any effects.

I changed the code a little bit. These values are set and used after dev-env `make` has ran. Here we want to use the values that have been defined in dev-env or default to latest. 

I'm not trying the set these values for the beginning of the test I'm just trying to make sure the same images are used trough out the whole test process. 

Values set: https://github.com/metal3-io/cluster-api-provider-metal3/blob/7aaf7b896c7939776198e9a3d19f2ec5a963c2b0/Makefile#L162
Values used: https://github.com/metal3-io/cluster-api-provider-metal3/blob/7aaf7b896c7939776198e9a3d19f2ec5a963c2b0/Makefile#L235

mquhuy:
> I changed the code a little bit. These values are set and used after dev-env `make` has ran. Here we want to use the values that have been defined in dev-env or default to latest.
> 
> I'm not trying the set these values for the beginning of the testing process I'm just trying to make sure the same images are used trough each part of a given test.

It also wouldn't work in that case. Devenv `make` is ran inside a subshell, so whatever values it has set won't be picked by either ci-e2e.sh or make. The only way you could make sure the two processes use the same images is by setting the images here towards devenv (via ci-e2e.sh), not vice verse.

peppi-lotta:
> > I changed the code a little bit. These values are set and used after dev-env `make` has ran. Here we want to use the values that have been defined in dev-env or default to latest.
> > I'm not trying the set these values for the beginning of the testing process I'm just trying to make sure the same images are used trough each part of a given test.
> 
> It also wouldn't work in that case. Devenv `make` is ran inside a subshell, so whatever values it has set won't be picked by either ci-e2e.sh or make. The only way you could make sure the two processes use the same images is by setting the images here towards devenv (via ci-e2e.sh), not vice verse.

Oh I see you pont! I've made the change now so that it is using values that should be defined and coming from project-infra.

peppi-lotta:
All right! All the images that are pulled here have already been pulled in previous phases of testing. This pull causes an unnecessary unused images to be pull when test are run on any release branch. There is no need to pull the images here. I suggest we remove this pull altogether. 

I'm abandoning the "defaulting" to correct values because I don't want to add many lines to `Makefile` when the problem can be solved by deleting :)

adilGhaffarDev:
> I'm abandoning the "defaulting" to correct values because I don't want to add many lines to `Makefile` when the problem can be solved by deleting :)

Are images pulled in `dev-env`? If so, let's implement the defaulting and pulling in `ci-e2e.sh` as suggested by @mquhuy. This means images will be pulled in `ci-e2e.sh`, while `dev-env` will first check for existing images before attempting to pull them.

Another reason for following this approach is to reduce dependency on `dev-env`, aligning with our goal of gradually removing it. This PR will contribute to that effort if we do that.

I am also fine to merge this PR just like that and do that in a separate PR maybe later. But that is something we will have to do eventually.

adilGhaffarDev:
> `Metal3-dev-env` has logic for defining the correct image and pulling them in `common.sh`, `releases.sh` and `image_prepull.sh`. `Metal-dev-env` pulls both container and node images. Is it actually smart at this point to duplicate so much code from from `metal3-dev-env` in to this project? Would it be smarter to do this once more decoupling is done? If we do this now we will have the same code twice without any gain.

yes, we can do it later. 
LGTM from my side.
if no objection from @mquhuy we can merge.

kashifest:
@mquhuy this is waiting for your confirmation

mquhuy:
> `Metal3-dev-env` has logic for defining the correct image and pulling them in `common.sh`, `releases.sh` and `image_prepull.sh`. `Metal-dev-env` pulls both container and node images. Is it actually smart at this point to duplicate so much code from from `metal3-dev-env` in to this project? Would it be smarter to do this once more decoupling is done? If we do this now we will have the same code twice without any gain.

Yes, I think you made a good point. We will need to handle this issue when the dev-env is removed from E2E anyway.

/lgtm

PR #2108, opened 2024-11-22
✨  Deprecate NoCloudProvider and add CloudProviderEnabled

adilGhaffarDev:
**What this PR does / why we need it**:
Deprecate NoCloudProvider add CloudProviderEnabled, and set the default value of CloudProviderEnabled to true.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of https://github.com/metal3-io/cluster-api-provider-metal3/issues/2255

PR #2276, opened 2025-01-24
🌱 remove  kubebuilder:default=true for CloudProviderEnabled and set defaults in defaulting webhook

adilGhaffarDev:
Instead of using `// +kubebuilder:default=true` to set a default value for `CloudProviderEnabled`, we are configuring defaults through a defaulting webhook. This approach avoids conflicts with the `NoCloudProvider` field in the validation webhook.

Using `kubebuilder:default` makes it impossible to distinguish whether the value of `CloudProviderEnabled` was explicitly set by the user or automatically assigned by the default. This distinction is critical because our validation webhook ensures that `CloudProviderEnabled` and `NoCloudProvider` do not conflict when both are set. If we rely on `kubebuilder:default`, the validation webhook cannot function as intended, as it cannot determine the source of the `CloudProviderEnabled` value. By shifting defaulting logic to the webhook, we retain the ability to enforce these validation rules effectively.

This PR is a follow-up of this PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2108 

Fixes #

### DIFF
--- a/api/v1beta1/metal3cluster_types.go
+++ b/api/v1beta1/metal3cluster_types.go
@@ -47,7 +47,7 @@ type Metal3ClusterSpec struct {
 	// If set to false, CAPM3 will use node labels to set providerID on the kubernetes nodes.
 	// If set to true, providerID is set on nodes by other entities and CAPM3 uses the value of the providerID on the m3m resource.
 	// TODO: Change the default value to false in release 1.12. Ref: https://github.com/metal3-io/cluster-api-provider-metal3/issues/2255
-	// +kubebuilder:default=true
+	// Default value is true, it is set in the webhook.
 	// +optional
 	CloudProviderEnabled *bool `json:"cloudProviderEnabled,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


Sunnatillo (commented):
/lgtm

tuminoid:
Actual test need to be done via test PR in dev-env, targeting this branch.

/cc @kashifest
PTAL.

tuminoid (commented):
IMO this seems fine.

adilGhaffarDev:
> It is still not clear to me what issue is it solving, and why the current way of setting the default using kubebuilder tag is an issue, can you describe the issue in a bit more detail ?

i have updated the description of this PR and also the previous PR please check.

kashifest:
> i have updated the description of this PR and also the previous PR please check.

Thanks a lot. We had offline discussion as well to clear the issue out. 

/approve

tuminoid:
Needs rebase on top of #2277 after that has merged.

tuminoid (commented):
/lgtm

PR #2321, opened 2025-02-12
:seedling: pin osv-scanner image in verify-release.sh

tuminoid:
osv-scanner:latest image is built from unpublished v2 branch now, which does not support same command line args as v1. Pin it to v1.9.2 for now and bump the command line usage to v2 compatible when it is published.

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -153,7 +153,7 @@ if [[ -n "${CONTAINER_RUNTIME}" ]]; then
         "${CONTAINER_RUNTIME}" run --rm
         -v "${PWD}":/src -w /src
         --pull always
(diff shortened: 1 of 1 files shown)


tuminoid:

kashifest:
how about the other images which we are using `latest` like gcrane for example? We should pin them all IMO.

tuminoid:
> how about the other images which we are using `latest` like gcrane for example? We should pin them all IMO.

I can do that too when I rework the script when osv-scanner v2 is released. This local binary vs container image is bound to have conflicts regardless, so I need to think a better way or remove support of one of them.

lentzi90 (commented):
/lgtm

PR #2357, opened 2025-02-27
:book: Fix broken IPAM kustomization links in releasing.md

peppi-lotta:
**What this PR does / why we need it**: Change IPAM links to pass link check. These wiles no longer exist in main but are relevant for release branch.


tuminoid (commented):
/approve
/retitle :book: Fix broken IPAM kustomization links in releasing.md

PR #2364, opened 2025-03-04
:sparkles: Change IsProvisioned to include NodeReady

honza:
This allows us to give the Metal3Machine a `ProviderID` and mark it as ready and then still keep trying to find its Node.  Once the Node is found, we set the `KubernetesNodeReady` condition.

In some deployment scenarios, the creation of a Node depends on the InfraMachine being ready.  This change increases the granularity of what it means to be ready.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -163,7 +163,7 @@ func (m *MachineManager) UnsetFinalizer() {
 
 // IsProvisioned checks if the metal3machine is provisioned.
 func (m *MachineManager) IsProvisioned() bool {
-	if m.Metal3Machine.Spec.ProviderID != nil && m.Metal3Machine.Status.Ready {
+	if m.Metal3Machine.Spec.ProviderID != nil && m.Metal3Machine.Status.Ready && m.isKubernetesNodeReady() {
 		return true
 	}
 	return false
@@ -1316,6 +1316,7 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, providerIDOnM3M 
 	}
 
 	if matchingNodesCount == 1 {
+		m.SetConditionMetal3MachineToTrue(infrav1.KubernetesNodeReadyCondition)
 		return nil
 	}
 	nodes, countNodesWithLabel, err := m.getNodesWithLabel(ctx, nodeLabel, clientFactory)
@@ -1365,6 +1366,7 @@ func (m *MachineManager) SetNodeProviderID(ctx context.Context, providerIDOnM3M 
 			return errors.Wrap(err, "unable to update the target node with providerID")
 		}
 	}
+	m.SetConditionMetal3MachineToTrue(infrav1.KubernetesNodeReadyCondition)
 	m.Log.Info("ProviderID set on target node")
 	return nil
 }
@@ -1374,7 +1376,6 @@ func (m *MachineManager) SetProviderID(providerID string) {
 	m.Log.Info("ProviderID set on the Metal3Machine", "providerID", providerID)
 	m.Metal3Machine.Spec.ProviderID = &providerID
 	m.Metal3Machine.Status.Ready = true
-	m.SetConditionMetal3MachineToTrue(infrav1.KubernetesNodeReadyCondition)
 }
 
 // SetOwnerRef adds an ownerreference to this Metal3Machine.
@@ -1878,4 +1879,8 @@ func (m *MachineManager) duplicateProviderIDsExist(validNodes map[string][]strin
 		return errors.New(errMessage)
 	}
 	return nil
+}
+
+func (m *MachineManager) isKubernetesNodeReady() bool {
+	return conditions.IsTrue(m.Metal3Machine, infrav1.KubernetesNodeReadyCondition)
 }
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -40,6 +40,7 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	capierrors "sigs.k8s.io/cluster-api/errors"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
@@ -398,6 +399,9 @@ var _ = Describe("Metal3Machine manager", func() {
 				},
 				Status: infrav1.Metal3MachineStatus{
 					Ready: true,
+					Conditions: clusterv1.Conditions{
+						*conditions.TrueCondition(infrav1.KubernetesNodeReadyCondition),
+					},
 				},
 			},
 			ExpectTrue: true,


kashifest:
I am a bit lost about the use case, to my understanding CAPM3 setting the providerID on the node via bootstrapping is already an indication that the node is found and ready. So essentially now with this change wont we set the node ready condition to true even before the M3M status to ready?

baremetal/metal3machine_manager.go
```diff
@@ -163,7 +163,7 @@ func (m *MachineManager) UnsetFinalizer() {
 
 // IsProvisioned checks if the metal3machine is provisioned.
 func (m *MachineManager) IsProvisioned() bool {
-	if m.Metal3Machine.Spec.ProviderID != nil && m.Metal3Machine.Status.Ready {
+	if m.Metal3Machine.Spec.ProviderID != nil && m.Metal3Machine.Status.Ready && m.isKubernetesNodeReady() {
```
lentzi90, line 166: This doesn't make sense to me. The infra machine can be provisioned without the node being ready. Then you know you have a bootstrap/config issue, not an infra issue. If we say the M3M is provisioned only when the node is ready, how would we distinguish between these?

lentzi90 (commented):
I don't think we should be dealing with node ready conditions at the infra provider level at all. This could be very confusing at best, completely misleading at worst. We do not know for sure if the node is ready based on this check, we only know that it exists, i.e. it joined the cluster. But honestly, I think the node level is better handled at the CAPI/Machine level, as it already is.

baremetal/metal3machine_manager.go
honza: I agree that it would helpful to come up with a better name.

lentzi90:
Ok let me see if I got it right.

Current behavior:
- We say that the Metal3Machine is `provisioned` when it is `Ready` and has a `ProviderID`. At this point we start returning early in the reconciliation loop.
- If the "cloud provider" functionality of CAPM3 is used, we try to set the `ProviderID` on the Node also *before* marking the Metal3Machine as `Ready`

If I understand correctly, you are saying that there is a scenario where the Node *cannot* be created until the Metal3Machine is `Ready`, because there is an external process that needs to verify the Metal3Machine `ProviderID` or something like this before it can approve the kubelet CSR.

It looks like a solid use-case and something we should support. The question is then how it should be done. I don't think that changing what `Provisioned` means is the way. Instead I think we should refactor a bit so that we do these things separately:
1. set the `ProviderID` on the Metal3Machine
2. set the `ProviderID` on the Node, if cloud provider functionality is enabled, otherwise skip
3. mark the Metal3Machine as `Ready`

Essentially, I would suggest changing the code here instead: https://github.com/metal3-io/cluster-api-provider-metal3/blob/2ab2c8db19b086f44d100b4db92daf9bf4b8ef2a/controllers/metal3machine_controller.go#L271-L282
Can we hide the `SetNodeProviderID` behind a check for `CloudProviderEnabled`? It seems strange to me that we go all the way in and try to retrieve the Node before we do this check now.

I also think that CAPM3 should only set the NodeReady condition if the cloud provider functionality is enabled. It is the job of the cloud provider to do that and CAPM3 has no reason to deal with the Nodes if this isn't enabled.

kashifest:
> We need to decouple these two operations:
> 
> * Setting ready and ProviderID on the Metal3Machine
> * Setting ProviderID on the Node
> 
> During bootstrap initialization, the kubelet creates a CSR which needs to be approved before a Node can be created. This CSR can be approved in [one of two ways](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-tls-bootstrapping/#bootstrap-initialization):
> 
> * kube-controller-manager automatically approves the CSR
> * an outside process, possibly a person, approves the CSR using the Kubernetes API or via kubectl
> 
> It seems to me that we only deal with the first option at present.
> 
> In the second scenario, the presence of a Node may depend on the Metal3Machine being ready and having a ProviderID. We must be able to come back _later_ and set the ProviderID (after the Node is created). We can only effectively stop reconciling the Metal3Machine resource when a Node is ready (via the `IsProvisioned()` check). Considering `ready` and checking for a `ProviderID` is not enough.
> 
> A CAPI Machine goes from _provisioned_ to _running_ when a Node with a matching `ProviderID` is found. The CAPI Machine [controller treats a missing Node as a warning only](https://github.com/honza/cluster-api/blob/93db5b2d30ca19506e3d1994e5a8959ce271c4a2/internal/controllers/machine/machine_controller_noderef.go#L89). It expects that it might take a while to find a Node despite `ProviderID` being set.
> 

I think I get the use case now and it seems legit to me as well. Thanks for the explanation.

baremetal/metal3machine_manager.go
zaneb: This function is only called in one place, to see if the M3M is done enough that we can stop reconciling. It could just as easily be called `CanStopReconciling()`. (Alternatively, the `&& m.isKubernetesNodeReady()` could be moved to the site of the call instead of within the `IsProvisioned()` function, but that seems more prone to future coding errors to me.)

Traditionally in CAPI:

- Provisioned = Machine has provider ID
- Running = Machine has NodeRef

lentzi90: Well, I think we do need a way to determine if the M3M is provisioned, so it does make sense to have a function like this. However, the infra provider in general should not be dealing with nodes at all. It should reconcile InfraMachines and leave the node stuff to CAPI and the external cloud provider.

The confusing part for us is that CAPM3 can **optionally** do some of the external cloud providers job. This is my issue here. **IF** that functionality is enabled, **then** I am fine with checking the node, not otherwise. I realize that we already set this node ready condition even if the cloud provider functionality is disabled. That is a mistake IMO and we should change it. In any way, we should not make it worse here by including it in more places.

As @honza mentioned below, we need to decouple ProviderID on M3M and ProviderID on Node. I totally agree, but I think including the node ready condition here actually couples them more.

honza:
## Setting the ProviderID on the metal3machine

Where does this `ProviderID` come from?

The `ProviderID` can come from:

- The Node (when Cloud Provider is enabled, via `Spec.ProviderID`)
- The Node (when kubeadm is used, via label)
- CAPM3 controller (when the above are disabled)

When using Cloud Provider, we need to wait for the Node to be created and then we try to find it by `ProviderID` based on known patterns.  Once we find it, we can take its actual `ProviderID` and set it on the `Metal3Machine`.  Because the `ProviderID` now exists on both the `Metal3Machine` and the Node, we can set `ready=true` on the `Metal3Machine`.

When using kubeadm, we also wait for the Node to created and then we expect to find the `metal3.io/uuid` label on it.  We convert this label into a `ProviderID` and set it on both the Node and the `Metal3Machine`.  Here we can also set `ready=true` on the `Metal3Machine` because the `ProviderID` is set on the Node and the `Metal3Machine`.

When using neither, the CAPM3 controller generates the `ProviderID` based on the baremetalhost and sets it on the `Metal3Machine`.  The controller has to wait for the Node to be created.  But the Node won't be created until the infrastructure is ready.  There is nothing left for us to do.  So, we set the `ProviderID` on the `Metal3Machine` and `ready=true`.  This allows outside processes to approve the Node.  So before we can be done, we need to find this Node and set the `ProviderID` on it.  Before checking if the `Metal3Machine` is provisioned (and saying we are done reconciling), we need to check if we still need to set the `ProviderID` on the Node.  I think this can be accomplished by using this check:

``` go
if machineMgr.IsProvisioned() && machineMgr.HasNodeRef() {
```

If we don't have a `NodeRef` yet, we need to keep trying to set the `ProviderID` on the Node.  That's it.

The part that is still missing is: how do we find the Node that was externally approved?  That is addressed in my other PR.  I propose that we use hostnames.

lentzi90:
Oh I had missed the part about kubeadm and node label. So we take the label from the node and use that even when the cloud-provider functionality is disabled? This sounds backwards to me. Why don't we set the ProviderID on the M3M directly when we associate it with a BMH? :thinking: 

I thought the cloud-provider functionality was only for setting the ProviderID on the node and that we always set it on the M3M directly based on the BMH we select regardless.

honza:
> Why don't we set the ProviderID on the M3M directly when we associate it with a BMH? 

Sorry, I don't have enough context/history to know the reasons.

honza:
I'll take a stab at refactoring this.

zaneb:
> Why don't we set the ProviderID on the M3M directly when we associate it with a BMH?

Because traditionally for CAPI, seeing the ProviderID on the Machine was the trigger for the Machine to change state from `Provisioning` to `Provisioned`. (I'm talking back before there were separate Metal3Machines, but I assume this is still true.)

honza:
Superseded by #2365

PR #2380, opened 2025-03-06
🌱 E2E: Ensure cert-manager webhook is available

lentzi90:

The clusterctl-upgrade periodics that have been failing lately due to two issues. This is addressing the first of them. What happens is that the cert-manager webhook is not completely ready when we try to deploy ironic, even though we wait for the deployments to be `Available`. This is because it takes some time for the webhook to be registered, get a valid certificate and so on.
To fix this, we add a simple check that we can `Eventually` apply an Issuer and Certificate. When that succeeds we know the webhook is working.

Fixes #

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,9 @@ out
 # goland
 .idea
 
+# zed
+.zed*
+
 # Common editor / temporary files
 *~
 *.tmp
--- a/test/e2e/data/cert-manager-test.yaml
+++ b/test/e2e/data/cert-manager-test.yaml
@@ -1,0 +1,28 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: test
+spec:
+  selfSigned: {}
+---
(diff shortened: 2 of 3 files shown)


lentzi90:
The clusterctl-upgrade test is passing the webhook issue that we see in the periodics. I still expect it to fail due to another issue that we also see in the periodics. It will take more work to fix that, but I suggest we merge this for now to get rid of the first issue at least

lentzi90:
As expected and explained above. I suggest to still merge this to get rid of the first error.
/override metal3-e2e-clusterctl-upgrade-test-main

tuminoid (commented):
/approve

lentzi90:
/hold
Looks like this still doesn't solve it fully. It happened again in https://github.com/metal3-io/cluster-api-provider-metal3/pull/2382 even though it includes this commit. Rerunning the test here also to gather some data.

mquhuy:
/lgtm
Hopefully this will solve the issue

lentzi90:
Ok the code seems to work now. There could still be some flakiness if the webhook becomes available and then unhealthy again later for some reason, but I think we are fairly safe like this. (clusterctl-upgrade will still fail. I'll override it.)

PR #2384, opened 2025-03-11
:seedling: Change IPAM provider name to independent Metal3 IPAM in e2e test

mquhuy:
After the IPAM provider name was changed from `metal3ipam` to `metal3` in https://github.com/metal3-io/metal3-dev-env/pull/1509, we need to do the same thing in capm3 e2e.


lentzi90 (commented):
/approve
/hold
Let's make sure that the pivot succeedes

tuminoid (commented):
/lgtm

PR #2385, opened 2025-03-12
:seedling: fix deprecation notice paragraph style

tuminoid:
Per godoc, it must be another paragraph in same comment block. Fix the two instances where they're not. This results documentation generate properly.


lentzi90 (commented):
/approve
Thanks for fixing this

PR #2402, opened 2025-03-17
:bug: bump x/net, x/crypto and x/oauth2 go mods

tuminoid:
Bump mods to get rid of vulnerabilities in the versions, regardless of if they're called/uncalled.


Sunnatillo (approved):
/approve

PR #2403, opened 2025-03-17
:bug: bump x/net, x/crypto, x/oauth2

tuminoid:
Bump some modules to get rid of vulnerabilities in them, regardless of if they're called/uncalled.


Sunnatillo (commented):
/approve

PR #2500, opened 2025-04-16
:seedling: Enabled tagliatelle linter

### LINKED ISSUE #2448: Enable "tagliatelle" linter in golangci-lint
Enable "tagliatelle" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

erjavaskivuori:
**What this PR does / why we need it**: Enabled tagliatelle linter

Fixes #2448

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -61,7 +61,7 @@ linters:
   - rowserrcheck
   - staticcheck
   - stylecheck
-  #- tagliatelle
+  - tagliatelle
   - testifylint
   - thelper
   - tparallel
@@ -135,6 +135,11 @@ linters-settings:
     - unnecessaryDefer
     - whyNoLint
     - wrapperFunc
+  tagliatelle:
+    case:
+      rules:
+        json: goCamel
+      
 issues:
   exclude-dirs:
   - mock*
@@ -198,6 +203,9 @@ issues:
     - stylecheck
     text: "ST1016: methods on the same type should have the same receiver name"
     path: .*(api|types)\/.*\/conversion.*\.go$
+  - linters:
+    - tagliatelle
+    text: "DHCP|SLAAC"
   include:
   - EXC0002 # include "missing comments" issues from golangci-lint
   max-issues-per-linter: 0


.golangci.yaml
```diff
@@ -198,6 +198,9 @@ issues:
     - stylecheck
     text: "ST1016: methods on the same type should have the same receiver name"
     path: .*(api|types)\/.*\/conversion.*\.go$
+  - linters:
+    - tagliatelle
+    text: "IP|ID|DHCP|SLAAC"
```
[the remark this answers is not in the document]
erjavaskivuori: That is a good idea, I will look into it!

.golangci.yaml
```diff
@@ -135,6 +135,16 @@ linters-settings:
     - unnecessaryDefer
     - whyNoLint
     - wrapperFunc
+  tagliatelle:
+    case:
+      extended-rules:
+        json:
+          case: ''
```
lentzi90, line 142: We use camel case for the json tags so I think we should set it here
```suggestion
          case: 'camel'
```

.golangci.yaml
```diff
@@ -135,6 +135,16 @@ linters-settings:
     - unnecessaryDefer
     - whyNoLint
     - wrapperFunc
+  tagliatelle:
+    case:
```
lentzi90, line 139: We should probably add the "normal" rules here also
```suggestion
    case:
      rules:
        json: 'camel'
```

.golangci.yaml
erjavaskivuori: I thought that too and tried it but for some odd reason then the initialism-overrides didn't work anymore (locally at least) and errors like `json(camel): got 'ipv6DHCP' want 'ipv6Dhcp' (tagliatelle)` returned.

.golangci.yaml
erjavaskivuori: In the documentation it is said "Important: the `extended-rules` overrides `rules`" (https://golangci-lint.run/usage/linters/#tagliatelle) so I am not sure if they should be there if they are overridden

.golangci.yaml
lentzi90: Oh :open_mouth: 
Does the same happen if you have both camel case in both the rules and extended-rules?

.golangci.yaml
lentzi90: Well, we need something to override, some kind of base line. I think the way it is supposed to work is that we put the "default" that we want in the rules and then if it doesn't quite work for everything, we extend by overriding in the extended-rules.

.golangci.yaml
erjavaskivuori: Found out that that custom initialisms aren't supported: https://github.com/ldez/tagliatelle/issues/2. Too solve this I changed to use part of the initial solution + case: goCamel which includes some common initialisms

.golangci.yaml
```diff
@@ -135,6 +135,11 @@ linters-settings:
...
+      rules:
+        json: goCamel
```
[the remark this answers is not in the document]
erjavaskivuori: I will take a look.

PR #2507, opened 2025-04-22
:seedling: Bump CAPI 1.10.0 and CAPM3 1.10.0-beta.0 in FKAS

peppi-lotta:
**What this PR does / why we need it**: Bump CAPI 1.10.0-rc.0 and CAPM3 1.10.0-beta.0 in FKAS


hack/fake-apiserver/go.mod
```diff
@@ -4,12 +4,12 @@ go 1.23.7
...
 require (
 	github.com/metal3-io/baremetal-operator/apis v0.10.0-beta.0
-	github.com/metal3-io/cluster-api-provider-metal3/api v1.9.3
+	github.com/metal3-io/cluster-api-provider-metal3/api v1.10.0-beta.0
 	go.uber.org/zap v1.27.0
 	k8s.io/api v0.32.3
 	k8s.io/apimachinery v0.32.3
 	k8s.io/client-go v0.32.3
-	sigs.k8s.io/cluster-api v1.9.6
+	sigs.k8s.io/cluster-api v1.10.0-rc.0
```
tuminoid, line 12: RC1 is already out.

hack/fake-apiserver/go.mod
```diff
@@ -4,12 +4,12 @@ go 1.23.7
...
 	sigs.k8s.io/cluster-api/test v1.9.6
```
adilGhaffarDev, line 13: When we uplift this, it will cause issues. 
I don't remember, but there is some function that we are using from the CAPI e2e framework that is either gone or moved.
I would say let's not uplift capi in FKAS.

peppi-lotta: All right if these can't be bumped then I will change this PR.

But before I make that change I would like to understand this issue more. The `docker build ` can be run in `hack/fake-apiserver` with out any issue and at least VScode is not showing any problems with any functions. So based on this the bump seems ok. What test can I run to verify that these bumps will actually break something?

adilGhaffarDev: There was a syntax error because of removed functions in v1.10.0 capi/test, it seems like it's not there anymore

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm
Let's merge if tests pass.

tuminoid:
> /lgtm Let's merge if tests pass.

WHAT THE **** bot, the test is still running for another hour ...

PR #2508, opened 2025-04-23
:seedling: Enabled mnd linter

### LINKED ISSUE #2445: Enable "mnd" linter in golangci-lint
Enable "mnd" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

erjavaskivuori:
**What this PR does / why we need it**: Enabled mnd linter

Fixes #2445


main.go
```diff
@@ -218,6 +218,16 @@ func main() {
 func initFlags(fs *pflag.FlagSet) {
 	logs.AddFlags(fs, logs.SkipLoggingConfigurationFlags())
 	logsv1.AddFlags(logOptions, fs)
+	leaderElectionLeaseTimeout := 15
```
tuminoid, line 221: Should we make these durations already to not repeat time.Duration(...) everywhere?

main.go
```diff
@@ -355,7 +365,8 @@ func waitForAPIs(cfg *rest.Config) error {
 		err = discovery.ServerSupportsVersion(c, metal3GV)
 		if err != nil {
 			setupLog.Info(fmt.Sprintf("Waiting for API group %v to be available: %v", metal3GV, err))
-			time.Sleep(time.Second * 10)
+			APIGroupWaitTimeout := 10
```
tuminoid, line 368: Move this up with others maybe?

baremetal/metal3machine_manager.go
```diff
@@ -1836,7 +1837,8 @@ func (m *MachineManager) getMatchingNodesWithoutLabelCount(ctx context.Context,
 	err = m.duplicateProviderIDsExist(validNodes, providerIDLegacy, providerIDNew)
 	if err != nil {
 		// There are, at least, two nodes. Details are in err
-		return 2, err
+		nodes := 2
```
tuminoid, line 1840: This feels dumb IMO, can we just nolint this one?

tuminoid (commented):
Some nits.

/cc @lentzi90

baremetal/metal3machine_manager.go
lentzi90: I agree

baremetal/metal3machine_manager.go
```diff
@@ -1758,7 +1758,8 @@ func (m *MachineManager) getMachineSet(ctx context.Context) (*clusterv1.MachineS
 func (m *MachineManager) getBmhNameFromM3Machine() (string, error) {
 	annotationValue := m.Metal3Machine.ObjectMeta.GetAnnotations()[HostAnnotation]
 	valueParts := strings.Split(annotationValue, "/")
-	if (len(valueParts) < 2) || (valueParts[0] != m.Metal3Machine.GetNamespace()) {
+	minValueParts := 2
+	if (len(valueParts) < minValueParts) || (valueParts[0] != m.Metal3Machine.GetNamespace()) {
```
lentzi90, line 1762: I think this would also be better with nolint. To me at least it is harder to grasp the `minValueParts` comparison than just `len(valueParts) < 2`

lentzi90 (commented):
Thanks for this!
The verbosity levels are backwards compared to the Kubernetes [recommendations](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use) that we should try to follow. Tracing should be 5 and debug 4.
I wonder if it would be possible to set the constants in a central place also? It is not so nice to define them in every package.

main.go
erjavaskivuori: Did this, but as a general note, I noticed that inside function setting e.g.

`leaderElectionLeaseTimeout := 15 * time.Second`

will cause new error about magic number 15 which is why these were then also moved outside the function to const () block

baremetal/metal3machine_manager.go
erjavaskivuori: Oops missed the comment about verbosity so my latest push doesn't address it. I tried to look for some common ways to name verbosity levels but didn't come across Kubernetes recommendations. I was also thinking that declaring different levels on repository level instead of package would be more elegant solution. Do you have any recommendations where this should be done? And would it be meaningful to declare also the ones that are not yet used or only the ones that are (4 and 5)?

lentzi90: No worries!
I think we could define the levels in the baremetal package, for example in [utils.go](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/utils.go). Hopefully that package is already imported in most places where we log things and would not cause circular dependencies.
I think it is fine to only define 4 and 5 for now. It is easy to add more later :slightly_smiling_face:

main.go
```diff
@@ -67,6 +67,11 @@ const (
 	// out-of-service taint strategy (GA from 1.28).
 	minK8sMajorVersionOutOfServiceTaint   = 1
 	minK8sMinorVersionGAOutOfServiceTaint = 28
+	leaderElectionLeaseTimeout            = 15 * time.Second
+	leaderElectionRenewTimeout            = 10 * time.Second
+	leaderElectionRetryTimeout            = 2 * time.Second
+	defaultMinSyncPeriod                  = 10 * time.Minute
+	APIGroupWaitTimeout                   = 10 * time.Second
```
lentzi90, line 74: I think we should keep this lowercase to avoid exporting it. There is a nice style guide here that explains the different situations: https://google.github.io/styleguide/go/decisions#initialisms
```suggestion
	apiGroupWaitTimeout                   = 10 * time.Second
```

tuminoid (commented):
/lgtm

Good job!

PR #2493, opened 2025-04-14
:seedling: Bump BMO v0.10.0-beta.0 IPAM v1.10.0-beta.0 and Ginkgo v2.23.4

peppi-lotta:
**What this PR does / why we need it**: Bump BMO v0.10.0-beta.0 IPAM v1.10.0-beta.0 and Ginkgo v2.23.4

PR #2516, opened 2025-04-24
:seedling: Downgrade versions in FKAS that were bumbed in PR: 2493

peppi-lotta:
**What this PR does / why we need it**: BMO and IPAM bumps in CAPM3 went in fkas as well: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2493. Version miss match is causing `docker build .` to fail. Tried fixing the build issue by upgrading all required versions. PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2507 which did resolve the build issue. Scalability test is broken in the PR so let's try downgrading instead.


tuminoid:
We just bumped it to v1.10.0, should this be closed? @peppi-lotta

PR #2519, opened 2025-04-24
🐛 Handle NICs without IP addresses

Sunnatillo:
This PR improves how NICs without IP addresses are handled during MachineStatus population.

Skips adding NICs that have an ip address field to prevent invalid Metal3Machine status updates. Interfaces with no ip addresses are visible on bmh.

Otherwise,  patching Metal3Machine status will fail with following error:

`status.addresses[n].address: Invalid value: "": addresses[n].address in body should be at least 1 chars long`

Issue seen in [BML tests](https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-bml-integration-test-centos/318/)

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -1218,6 +1218,9 @@ func (m *MachineManager) nodeAddresses(host *bmov1alpha1.BareMetalHost) []cluste
 			Type:    clusterv1.MachineInternalIP,
 			Address: nic.IP,
 		}
+		if address.Address == "" {
+			continue
+		}
 		addrs = append(addrs, address)
 	}
 


baremetal/metal3machine_manager.go
```diff
@@ -1218,6 +1218,11 @@ func (m *MachineManager) nodeAddresses(host *bmov1alpha1.BareMetalHost) []cluste
 			Type:    clusterv1.MachineInternalIP,
 			Address: nic.IP,
 		}
+		if address.Address == "" {
+			// When interface does not have an ip address set status to NoIP
+			address.Address = "NoIP"
+			m.Log.V(4).Info("NIC does not have an IP address, setting ipaddress NoIP in status", "nicName", nic.Name)
+		}
```
lentzi90, line 1225: Wouldn't it make more sense to just skip the NIC without address?
```suggestion
		if address.Address == "" {
			// Skip if interface does not have an ip address
			continue
		}
```

kashifest: I am also of the opinion that we should skip this. I had a chat with @Rozzii and we agreed that we should perhaps still keep it to show in m3m that there’s a nic without an IP in BMH. But now as I sie it it doesnt bring any extra information like MAC which is already in BMH. So @Rozzii do we agree that we should skip listing NiCs with no IP here?

Rozzii: From the user's perspective I can imagine scenarios where they would like to verify e.g. installation of new hardware so they would like to see whether a NIC is there or not.

There is an other perspetive from Ironic and IPA as all the MACs are used and logged during the IPA lookup process so that would again cause a situation where in the Ironic and IPA logs there are a set of MACs while the inspection shows a different list.

All that said I would still prefer listing the NiC with an empty IP field.

kashifest: Agreed. But isnt it enough that user is able to see it on BMH. Does putting that on m3m bring any xtra value?

tuminoid: Setting non-IP string "NoIP" to address field is also a big nope IMO.

lentzi90: As far as I know, we are not showing anything other than the "addresses" in the M3M. There is nothing about MACs or NICs there so listing "NoIP" is not useful anyway IMO. Best case, it tells the user that there is "something" that could have an address but doesn't at the moment. Worst case, it is just confusing and breaking code that assumes these are real and proper addresses.

This is what we have in the M3M status:

```yaml
status:
  addresses:
  - address: 172.22.0.95
    type: InternalIP
  - address: fe80::cf2a:f20b:a6d2:9df8%enp1s0
    type: InternalIP
  - address: 192.168.111.21
    type: InternalIP
  - address: fe80::9bcc:260d:42be:26bc%enp2s0
    type: InternalIP
  - address: node-1
    type: Hostname
  - address: node-1
    type: InternalDNS
```

kashifest: > Setting non-IP string "NoIP" to address field is also a big nope IMO.

@tuminoid Agreed. This was my suggestion since while debugging we wanted to know if this is the case we are hitting. I think most of us are of the same opinion. So @Sunnatillo lets remove the NIC entry with IP from here. We can discuss more once @Rozzii is back from vacation if this is still needed.

Sunnatillo: We have NIC with no ip on bmh, imho,  we do not need on m3m.

baremetal/metal3machine_manager.go
```diff
@@ -1498,6 +1501,10 @@ func (m *MachineManager) AssociateM3Metadata(ctx context.Context) error {
 	if m.Metal3Machine.Spec.DataTemplate.Namespace == "" {
 		m.Metal3Machine.Spec.DataTemplate.Namespace = m.Metal3Machine.Namespace
 	}
+	// Attempt to fetch an existing Metal3DataClaim.
+	// If the claim exists, no further action is needed.
+	// If an error occurs and it's non-transient, return the error.
+	// If the error is transient (e.g., temporary API issue), proceed to create the claim.
```
kashifest, line 1507: Where is this change coming from?
/approve
/hold

Sunnatillo: ohh, this comments I added during debugging where code was not clear to me. I will remove them now.

PR #2529, opened 2025-04-29
:seedling: Bump BMO to v.0.10.1

peppi-lotta:
**What this PR does / why we need it**: Bump BMO to v.0.10.1

PR #2432, opened 2025-03-20
⚠️ Bump CAPI to v1.10.0-beta.0

adilGhaffarDev:
**What this PR does / why we need it**:
- Bumping capi to v1.10.0-beta.0
- Bumping Controller Runtime to 0.20.3
- Moving webhooks from `api/v1beta1` folder to `internal/webhooks/v1beta1`
- Moving webhooks to separate webhooks package.
- webhooks in `api/v1beta1` folder are still there but deprecated and they are not used anywhere. In next release we can remove these webhooks and also their tests.

We are implementing webhook changes based on https://github.com/kubernetes-sigs/kubebuilder/pull/4150/files. While simply changing the folder might be sufficient, moving the webhooks to a separate package provides the additional benefit of allowing us to add dependencies specific to webhooks without introducing them to the `api` package. This is something that CAPI is also doing.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2421

PR #2530, opened 2025-04-29
:rocket: Release v1.10.0

peppi-lotta:
**What this PR does / why we need it**: Release v1.10.0

### DIFF
--- a/releasenotes/v1.10.0.md
+++ b/releasenotes/v1.10.0.md
@@ -1,0 +1,113 @@
+<!-- markdownlint-disable no-inline-html line-length -->
+# Changes since v1.9.0
+
+## :warning: Breaking Changes
+
+- Bump CAPI to v1.10.1 (#2531, #2432)
+- Deprecate Metal3data templateReference (#2326)
+- Remove IPAM deployment logic (#1993)
+- Remove Metal3Machine owner reference from BMH (#1742)
+
+## :sparkles: New Features
+
+- Deprecate NoCloudProvider and add CloudProviderEnabled (#2108)
+
+## :bug: Bug Fixes
+
+- Handle NICs without IP addresses (#2519)
+- Revert "Remove Metal3Machine owner reference from BMH" (#2254)
+
+## :book: Documentation
+
+- Remove IPAM manifests related links and check in release process (#2355)
+- update OpenSSF CII Best Practices badge (#2304)
+- Updating docs for release 1.9 (#2232)
+- Improve e2e documentation (#2209)
+
+## :seedling: Others
+
+- Bump BMO to v0.10.1 (#2529)
+- Bump Kubernetes to v1.33.0 (#2523)
+- Bump IPAM 1.10.0 and BMO 0.10.0 (#2522)
+- Bump the kubernetes group to v0.32.4 (#2513)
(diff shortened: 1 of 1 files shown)


adilGhaffarDev:
Should we wait for this: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2529 to merge?

adilGhaffarDev:
/hold
For some reason, tests are still using v1.10.0-rc.1 of CAPI, we missed something. I will check.

adilGhaffarDev:
opened the PR: https://github.com/metal3-io/metal3-dev-env/pull/1525 
Not sure if it's the best way to solve this, but it works

tuminoid:
> https://github.com/metal3-io/metal3-dev-env/blob/8cd873825ebf1b7e06e3bf2b76690ff23c625959/lib/releases.sh#L7 This is not working properly; we need to fix it. It's returning v1.10.0-rc.1 even though the latest release v1.10.0. `sort -rV` in above function is not sorting according to semantic version, instead just doing simple string sort. I will fix this.

Yup, `-V` is not semantic version sort, just a version sort.

Currently v1.10.1 is out and it has solved this problem for us already, but let's merge the fix, as it removes this short window of failure on next cycle.

tuminoid:
> Should we wait for this: #2529 to merge?

It is merged now anyways.

releasenotes/v1.10.0.md
```diff
@@ -0,0 +1,112 @@
+<!-- markdownlint-disable no-inline-html line-length -->
+# Changes since v1.9.0
+
+## :warning: Breaking Changes
+
+- Bump CAPI to v1.10.0-beta.0 (#2432)
```
lentzi90, line 6: I think we should clean this up so that #2517 is listed here and #2432 is put as superseeded

tuminoid: This is the one PR that actually has changes. We could combine them "bump to capi v1.10.0" and list both here, but just adding the simple module bump from rc.1 to stable with no other changes would be very misleading IMO.

adilGhaffarDev:
> > Should we wait for this: #2529 to merge?
> 
> It is merged now anyways.

@peppi-lotta please update the release notes so it contains https://github.com/metal3-io/cluster-api-provider-metal3/pull/2529 too.

releasenotes/v1.10.0.md
lentzi90: My issue with this is that it looks like we are bumping to a beta release and then in the fine print we are saying that we actually took he proper release. I know the actual changes happened when we took the first pre-release, but it was still superseded by the later proper release. If we would have skipped the pre-release, then we would definitely list the proper release as breaking, right?

peppi-lotta:
Let's wait until https://github.com/metal3-io/cluster-api-provider-metal3/pull/2531 is merged then this should be fine to merge.

releasenotes/v1.10.0.md
```diff
@@ -0,0 +1,113 @@
+<!-- markdownlint-disable no-inline-html line-length -->
+# Changes since v1.9.0
+
+## :warning: Breaking Changes
+
+- Bump CAPI to v1.10.1 (#2531)(#2432)
```
tuminoid, line 6:
```suggestion
- Bump CAPI to v1.10.1 (#2531, #2432)
```

lentzi90 (commented):
/approve

PR #2534, opened 2025-04-30
:seedling: Change Kubernetes version in e2e tests

peppi-lotta:
**What this PR does / why we need it**: Change Kubernetes version in e2e tests
Envtest uplift to 1.33 is done in a separate PR https://github.com/metal3-io/cluster-api-provider-metal3/pull/2535 because the is an issue with it.

### DIFF
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -170,8 +170,8 @@ variables:
   # KUBERNETES_PATCH_FROM_VERSION and KUBERNETES_PATCH_TO_VERSION are used to
   # test upgrade scenarios where we only want to upgrade the patch version of
   # kubernetes.
-  KUBERNETES_PATCH_FROM_VERSION: "v1.31.0"
-  KUBERNETES_PATCH_TO_VERSION: "v1.31.2"
+  KUBERNETES_PATCH_FROM_VERSION: "v1.32.0"
+  KUBERNETES_PATCH_TO_VERSION: "v1.32.1"
   CONTROL_PLANE_MACHINE_COUNT: 3
   WORKER_MACHINE_COUNT: 1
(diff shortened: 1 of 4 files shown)


test/e2e/config/e2e_conf.yaml
```diff
@@ -170,8 +170,8 @@ variables:
   # KUBERNETES_PATCH_FROM_VERSION and KUBERNETES_PATCH_TO_VERSION are used to
   # test upgrade scenarios where we only want to upgrade the patch version of
   # kubernetes.
-  KUBERNETES_PATCH_FROM_VERSION: "v1.31.0"
-  KUBERNETES_PATCH_TO_VERSION: "v1.31.2"
+  KUBERNETES_PATCH_FROM_VERSION: "v1.31.2"
```
adilGhaffarDev, line 173:
```suggestion
  KUBERNETES_PATCH_FROM_VERSION: "v1.32.0"
```

lentzi90:
What's the deal with  metal3-e2e-1-29-1-30-upgrade-test-main? Should the name be updated?
I'm not sure what test to trigger so that it does the node reuse. We should test that since it makes use of these versions.

lentzi90:
metal3-ubuntu-e2e-feature-test-main-pivoting tested versions v1.32.0 and v1.32.1 as part of node reuse so that has been confirmed. I'll try to trigger the upgrade test now even though it has older versions in the name...
/test metal3-e2e-1-29-1-30-upgrade-test-main

lentzi90:
The k8s upgrade job is wrong. Fix here: https://github.com/metal3-io/project-infra/pull/1006
/hold
Until we have merged the fix and tested the correct k8s upgrade here.

lentzi90:
Confirmed that the correct versions are used in the upgrade job now.
/hold cancel

PR #2544, opened 2025-05-05
:seedling: Enable exhaustive linter

### LINKED ISSUE #2441: Enable "exhaustive" linter in golangci-lint
Enable "exhaustive" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

erjavaskivuori:
**What this PR does / why we need it**:  Enabled exhaustive linter

Fixes #2441

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -546,7 +546,7 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			return WithTransientError(errors.New(errMessage), 0*time.Second)
 		}
 
-		waiting := true
+		var waiting bool
 		switch host.Status.Provisioning.State {
 		case bmov1alpha1.StateRegistering,
 			bmov1alpha1.StateMatchProfile, bmov1alpha1.StateInspecting,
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
Can you add in the PR description why ignoring the linter is appropriate solution here, instead of  actually fixing it?

baremetal/metal3machine_manager.go
```diff
@@ -861,6 +864,10 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmov1alpha1.BareMetal
 			} else if !m.nodeReuseLabelExists(ctx, &host) {
 				switch host.Status.Provisioning.State {
```
tuminoid, line 865: We have a case now where all options are continue, which is same as removing the switch and just executing continue directly. Even the default option is continue.

baremetal/metal3machine_manager.go
```diff
@@ -558,6 +558,9 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 			// We have no control over provisioning, so just wait until the
 			// host is powered off.
 			waiting = host.Status.PoweredOn
+		case bmov1alpha1.StatePreparing, bmov1alpha1.StateProvisioning, bmov1alpha1.StateProvisioned,
+			bmov1alpha1.StateDeprovisioning, bmov1alpha1.StatePoweringOffBeforeDelete,
+			bmov1alpha1.StateDeleting:
```
tuminoid, line 563: Let's add a comment here at least that waiting = true for these, or add the literal waiting = true. Then L549 can be made `var waiting bool`. Let's also add default case, which defaults to true as well.

baremetal/metal3machine_manager.go
tuminoid: This is actually wrong, in Go the empty case is not falling thru. We added a comment to make this more explicit.

tuminoid (commented):
/approve
/cc @lentzi90

lentzi90 (commented):
/lgtm

PR #2548, opened 2025-05-06
:book: Update build badges and triggers in README.md

peppi-lotta:
**What this PR does / why we need it**: Update build badges and triggers in README.md


tuminoid (commented):
/retitle :book: Update build badges and triggers in README.md 
/approve
/lgtm

I'll merge this so we get CAPM3 tags fixed to enable e2e testing.

PR #2554, opened 2025-05-13
🌱 Fix ipam deploy in clusterctl tests

adilGhaffarDev:
Fix ipam deploy in clusterctl tests

Fixes #

### DIFF
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -107,15 +107,15 @@ providers:
 - name: metal3
   type: IPAMProvider
   versions:
-  - name: "v1.10.0"
-    value: "https://github.com/metal3-io/ip-address-manager/releases/download/v1.10.0/ipam-components.yaml"
+  - name: "{go://github.com/metal3-io/ip-address-manager@v1.10}"
+    value: "https://github.com/metal3-io/ip-address-manager/releases/download/{go://github.com/metal3-io/ip-address-manager@v1.10}/ipam-components.yaml"
     type: "url"
     contract: v1beta1
     replacements:
     - old: --metrics-addr=127.0.0.1:8080
       new: --metrics-addr=:8080
     files:
-    - sourcePath: "../data/infrastructure-metal3/main/metadata.yaml"
+    - sourcePath: "../data/ipam-metal3/v1.10/metadata.yaml"
       targetName: "metadata.yaml"
 
 - name: metal3
--- a/test/e2e/upgrade_clusterctl_test.go
+++ b/test/e2e/upgrade_clusterctl_test.go
@@ -43,8 +43,7 @@ var _ = Describe("When testing cluster upgrade from releases (v1.9=>current) [cl
 		imageURL, imageChecksum := EnsureImage(k8sVersion)
 		os.Setenv("IMAGE_RAW_CHECKSUM", imageChecksum)
 		os.Setenv("IMAGE_RAW_URL", imageURL)
-		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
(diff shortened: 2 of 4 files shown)


test/e2e/config/e2e_conf.yaml
```diff
@@ -107,15 +107,15 @@ providers:
...
+  - name: "{go://github.com/metal3-io/ip-address-manager@v1.10}"
+    value: "https://github.com/metal3-io/ip-address-manager/releases/download/{go://github.com/metal3-io/ip-address-manager@v1.10}/ipam-components.yaml"
     type: "url"
     contract: v1beta1
     replacements:
     - old: --metrics-addr=127.0.0.1:8080
       new: --metrics-addr=:8080
     files:
-    - sourcePath: "../data/infrastructure-metal3/main/metadata.yaml"
+    - sourcePath: "../data/infrastructure-metal3/v1.10/metadata.yaml"
```
peppi-lotta, line 118: IPAM is no longer a part of CAPM3 deployment but deployed as an independent IPAMProvider.  I'm thinking we should have a `/data/ipam-metal3/v1.10/` instead. 

This is fine as longs as IPAM and CAPM3 versions always have the same minor version but having `/data/ipam-metal3/v1.10/`  would be more correct IMO as IPAM release now has it own metadata.yaml as well.

adilGhaffarDev: I agree, I guess it makes it clear. Once the test is done, I will update the pr.

test/e2e/upgrade_clusterctl_test.go
```diff
@@ -44,7 +44,7 @@ var _ = Describe("When testing cluster upgrade from releases (v1.9=>current) [cl
 		os.Setenv("IMAGE_RAW_CHECKSUM", imageChecksum)
 		os.Setenv("IMAGE_RAW_URL", imageURL)
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
-		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
+		clusterctlLogFolder = filepath.Join(artifactFolder, bootstrapClusterProxy.GetName())
```
lentzi90, line 47: I think this is OK, but the comment above should be removed. We are not giving any special credentials to clusterctl that could be exposed in the logs.

test/e2e/upgrade_clusterctl_test.go
```diff
@@ -103,7 +103,7 @@ var _ = Describe("When testing cluster upgrade from releases (v1.8=>current) [cl
...
```
lentzi90, line 106: Same here

lentzi90 (commented):
Thanks for this @adilGhaffarDev ! I just have one minor thing below

test/e2e/upgrade_clusterctl_test.go
adilGhaffarDev: done

lentzi90 (commented):
/approve

test/e2e/data/infrastructure-metal3/v1.10/metadata.yaml
terror96, line 1: Sorry for a potentially stupid question, but is this file also needed because you replace in  test/e2e/config/e2e_conf.yaml the path from 'infrastructure-metal3' to 'ipam-metal3', and you below create an identical file in  'ipam-metal3'?

adilGhaffarDev: This is a totally valid question. We don't need it right now, but we will need it when we update the tests (which I will do once this PR is merged).
I will remove the file for now, thanks @terror96 for pointing it out.

tuminoid (commented):
/lgtm

PR #2555, opened 2025-05-14
✨ Add parameters blob for bond network interface definition

### LINKED ISSUE #1466: CAPM3 lacks of options for bond interfaces
**User Story**

As a [developer/user/operator] I would like to [high level description] for [reasons]
CAPM3 bond definition lacks of options for bond definition regarding what is possible with cloud-init.

**Detailed Description**

In Project Sylva we're maintaining a helm chart sylva-capi-cluster to create CAPI manifests definitions for clusters (not using clusterctl). We use Metal3 to provision BM hosts in our cluster and our network design recommends using bond interfaces. For some use cases we set the bond_mode to 802.3ad (LACP).

With some OSs like SuSE, the network backend (`wicked`) needs explicit parameters to properly configure a bond. This `wicked` backend is not able to build a proper bond without the parameter `miimon` set to a given value (even if the bonding kernel module has a default one).

Metal3 lacks of options in the bond definition [1] and there is some missing info for the network configuration sent to cloud-init. If we take a look to the cloud-init documentation, a bond definition has multiple parameters for bonding [2] [3] that could properly set a bond even on SuSE OSs with wicked.

The problem here is if we need some tuning for any type of bond, it's impossible via Metal3 to provide it to the BM node. So it would be nice to introduce a way to inject bond parameters and having a 1:1 compatibility to what parameters are exposed by cloud-init.

[1] for example here https://github.com/metal3-io/cluster-api-provider-metal3/blob/v1.6.0/baremetal/metal3data_manager.go#L1001-L1007
[2] https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v1.html#params-dictionary-of-key-value-bonding-parameter-pairs
[3] https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html#bonds

-

/kind feature

surabujin:
Cloud-init allows to define a wide list of parameters for network bond interfaces. But definition of `NetworkDataLinkBond` type do not allow to pass externded set of parameters via network-data. This change adds "params" blob/dict into `NetworkDataLinkBond` object that can be used to pass any set of bond parameters.

Fixes #1466


tuminoid:
/ok-to-test

Please sign-off the commit for DCO check, and also, I feel :bug: is not correct for this but rather :sparkles: as it is an addition.

tuminoid (commented):
/cc @lentzi90 @kashifest @adilGhaffarDev

api/v1beta1/metal3datatemplate_types.go
```diff
@@ -264,11 +265,17 @@ type NetworkDataLinkBond struct {
 	// used to render it.
 	MACAddress *NetworkLinkEthernetMac `json:"macAddress"`
 
+	// params blob passed without any validation/modifications into cloud-init config
+	Params NetworkDataLinkBondParams `json:"params,omitempty"`
```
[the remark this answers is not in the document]
surabujin: Ok. Lets it be "parameters".

surabujin:
@lentzi90 

>This looks quite ok. But from what I understood in https://github.com/metal3-io/cluster-api-provider-metal3/issues/1466, the goal was to first add only bond_miimon, not support every option there is. I would rather go that path actually.

Sorry, looks like I have confused you. My goal is not to fix issue #1466, but improve bonds interface configuration so I would be able to pass any possible bond configuration parameter. I have mention issue #1466 just because it will be "also" fixed by this improvement. The  ability to pass only `bond_miimon` value will not satisfy my current demands.

>Allowing a "blob" like this in the API can be powerful but it can also be quite dangerous and easy to break.

Static format have disadvantages too - we should implement all cloud-init config options and track/follow all future changes in this config format.

This is not "our" data format. We do not proceed this data in any way. Even more we mimic format of cloud-init config. So operator knows the target config format and can expect that he can pass any required config option.

It was surprise for me that metal3data represents only some subset of cloud-init config parameters. I thought that whole network_config represented as some data "blob" with "variable" substitutions.

baremetal/metal3data_manager.go
```diff
@@ -992,15 +992,33 @@ func renderNetworkLinks(networkLinks infrav1.NetworkDataLink,
...
 			"mtu":                   link.MTU,
 			"ethernet_mac_address":  macAddress,
 			"bond_mode":             link.BondMode,
 			"bond_xmit_hash_policy": link.BondXmitHashPolicy,
 			"bond_links":            link.BondLinks,
-		})
+		}
+
+		illegalParameters := []string{}
+		for opt, value := range link.Parameters {
```
Rozzii, line 1006: I think this would be ineffective if I would supply
"bond_mode" or "bond_xmit_hash_policy" as parameter names/keys as in this logic the "opt" is expected to **not** contain the "bond_" prefix but that expectation is not enforced anywhere AFAIK. So I would be able to overwrite the structured data still.

Do I see it correctly or did I miss something here?

Rozzii (changes requested):
I am okay with this feature in principle even though it is very cloud-init inspired but I recognize that even though we have plans to broaden the network configuration tool support we still only support cloud-init natively.

I have one question/confusion that I have commented please reflect on that + please rebase.

baremetal/metal3data_manager.go
surabujin: Didn't get. How you will be able to overwrite existing/structured fields? Despite presence of "bond_" prefix inside option name into parameters map key this code will add (one more if already exists) prefix "bond_" and will check is it collide with already produced fields. 

If option will have more than one "bond_" prefix, option will be ignored by cloud-init.

Rozzii: Oh okay, I missed that it saves the modified name e.g it would be "bond_bond_x" propagated. Then all is fine, thanks for explaining.

Rozzii (commented):
/lgtm

Rozzii (commented):
Let's try to get this in @adilGhaffarDev

PR #2557, opened 2025-05-15
:seedling: [WIP] Add kcp_remediation test

### LINKED ISSUE #1080: Reuse CAPI e2e test
**User Story**

As a developer, I would like to reuse CAPI (Cluster API)  e2e tests such as[ kcp_remediation test](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/kcp_remediations.go) in CAPM3 (Cluster API Provider Metal3) however CAPI randomly generates the namespaces hosting the cluster in the [BeforeEach](https://github.com/kubernetes-sigs/cluster-api/blob/916b187262c96441f53f21c171308e1d7e1016b5/test/e2e/kcp_remediations.go#L106) of the test, and in CAPM3, the cluster expect the BMH resources to be pre-created that namespace.  I want to find a solution that allows us to make BMH (BareMetalHost) resources available in the namespace that hosts the cluster.

**Detailed Description**

Currently, when reusing CAPI e2e test in CAPM3, we encounter difficulties because CAPI creates clusters in randomly generated namespaces. However, in CAPM3, we pre-create the BMH resources in a fixed namespace where we expect the cluster to be created. This misalignment prevents us from reusing CAPI e2e tests.

To overcome this challenge, we need to find a way to synchronize the creation of namespaces and BMH resources, ensuring that the required BMHs are available in the namespace hosting the cluster at the time of the creation of this cluster. This will enable us to reuse many CAPI tests in CAPM3 instead of reimplimenting them.

We have explored potential workarounds, the best option seems to include BMH resources in the cluster template in a new CI cluster template dedicated for testing only, but this approach is not be suitable for official cluster templates and could potentially confuse developers and users regarding the intended production usage of Metal3.

Additionally, creating BMHs for each test poses challenges in terms of time and resource consumption. We need to consider the trade-off between running tests in parallel, skipping inspection, or finding alternative solutions to maintain efficiency without compromising the integrity of tests.

Overall, we aim to find a solution that allows us to reuse CAPI tests in CAPM3, ensuring the availability of BMH resources in the namespace generated by CAPI tests when creating a cluster.

I believe addressing this issue will improve the testability  of CAPM3, reusing more e2e tests that are maintained by CAPI community. This enhancement will facilitate the adoption of new e2e test created on CAPI.

/kind feature

erjavaskivuori:
**What this PR does / why we need it**: Adopting KCP remediation test from CAPI https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/kcp_remediations.go

Fixes #1080

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,6 +189,8 @@ cluster-templates: $(KUSTOMIZE) ## Generate cluster templates
 	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-upgrade-workload > $(E2E_OUT_DIR)/cluster-template-upgrade-workload.yaml
 	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-centos-md-remediation > $(E2E_OUT_DIR)/cluster-template-centos-md-remediation.yaml
 	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu-md-remediation > $(E2E_OUT_DIR)/cluster-template-ubuntu-md-remediation.yaml
+	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-centos-kcp-remediation > $(E2E_OUT_DIR)/cluster-template-centos-kcp-remediation.yaml
+	$(KUSTOMIZE) build $(E2E_TEMPLATES_DIR)/cluster-template-ubuntu-kcp-remediation > $(E2E_OUT_DIR)/cluster-template-ubuntu-kcp-remediation.yaml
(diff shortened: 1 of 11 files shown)


erjavaskivuori:
The KCP remediation test has this condition: "Control plane machines having a pre-kubeadm command that queries for a well-known ConfigMap on the management cluster, holding up bootstrap until a signal is passed via the config map. NOTE: In order for this to work communications from workload cluster to management cluster must be enabled." This cannot be satisfied because the management cluster is a kind-cluster created in dev-env and in order to to communicate with that from the BMHs we would have to expose the management cluster publicly

PR #2568, opened 2025-05-22
:seedling: WIP Adopt k8s conformance test from CAPI

erjavaskivuori:
**What this PR does / why we need it**: Adopting k8s conformance test from CAPI

Fixes #


docs/e2e-test.md
```diff
@@ -81,12 +82,13 @@ sudo rm -rf /opt/metal3-dev-env/
...
 ## Included tests
 
-The e2e tests currently include three different sets:
+The e2e tests currently include five different sets:
 
 1. Pivoting based feature tests
-1. Remediation based feature tests
-1. clusterctl upgrade tests
-1. K8s upgrade tests
+2. Remediation based feature tests
```
tuminoid: These are 1. on purpose, they're auto-generated by markdown on rendering.

scripts/ci-e2e.sh
```diff
@@ -50,7 +50,7 @@ export USE_IRSO="${USE_IRSO:-false}"
 EOF
 
 case "${GINKGO_FOCUS:-}" in
-  clusterctl-upgrade|k8s-upgrade|basic|integration)
+  clusterctl-upgrade|k8s-upgrade|basic|integration|k8s-conformance)
```
tuminoid, line 53: Note: This is conflicting with https://github.com/metal3-io/cluster-api-provider-metal3/pull/2596, one of you need to rebase. I'd love #2596 get priority as it is fixing broken test.

tuminoid:
Also, since we don't have PR job in Prow for this, how has this been tested. Please provide links or instructions?

docs/e2e-test.md
erjavaskivuori: Oh well that's convenient, didn't know this! I will change it back

test/e2e/data/kubetest/conformance.yaml
```diff
@@ -0,0 +1,15 @@
+ginkgo.focus: \[Conformance\]
+ginkgo.skip: \[Serial\]
+disable-log-dump: true
+# ginkgo.progress flag is deprecated but its still used in
+# k8s versions <= v1.26, we have to keep it if we want to 
+# test these versions
+ginkgo.progress: true
```
lentzi90: I think we can safely skip this

scripts/ci-e2e.sh
erjavaskivuori: Is it? In #2596 the changes are in environment.sh and these are ci-e2e.sh

test/e2e/e2e_suite_test.go
```diff
@@ -300,3 +301,26 @@ func updateCalico(config *clusterctl.E2EConfig, calicoYaml, calicoInterface stri
 	err = os.WriteFile(calicoYaml, yamlOut, 0600)
 	Expect(err).ToNot(HaveOccurred(), "Cannot print out the update to the file")
 }
+
+// postNamespaceCreated is a hook function that can be called after creating
+// a namespace, it creates the needed bmhs in the namespace hosting the cluster.
+func postNamespaceCreated(clusterProxy framework.ClusterProxy, clusterNamespace string) {
```
lentzi90: nit: Perhaps better to name it after what it is doing?
```suggestion
func createBMHsInNamespace(clusterProxy framework.ClusterProxy, clusterNamespace string) {
```

scripts/ci-e2e.sh
tuminoid: Nevermind. It is two similar if/case in two different files being edited.

erjavaskivuori:
I created the PR job too for this one https://gerrit.nordix.org/c/infra/cicd/+/23763 but I guess it can't be used yet since it isn't reviewed. I run the test by starting the pivoting tests (since it had suitable configurations) and then stopped and replayed with correct GINKGO_FOCUS set. Here is the run https://jenkins.nordix.org/view/Metal3/job/metal3-ubuntu-e2e-feature-test-main-pivoting/36/.

scripts/environment.sh
```diff
@@ -85,6 +85,13 @@ if [[ ${GINKGO_FOCUS:-} == "features" && ${GINKGO_SKIP:-} == "pivoting remediati
   export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"2"}
 fi
 
+# K8s conformance test environment vars and config
+if [[ ${GINKGO_FOCUS:-} == "k8s-conformance" ]]; then
+  export NUM_NODES="6"
+  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"1"}
+  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"5"}
```
lentzi90, line 104: Do we need so many? If we do, it would be good to link to some source to explain it. E.g. conformance tests need X nodes to test functionality Y...

lentzi90 (commented):
Nice work!
I would suggest testing this by adding a test commit that makes the normal integration tests run the conformance test instead (hard code it to always run only the conformance test and ignore any ginkgo focus things). That way we can see how it goes without first going through the hassle of adding the job to prow and/or jenkins etc.

scripts/environment.sh
erjavaskivuori: Yes, it is because CAPI has it hard coded that there must be 5 workers https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/k8s_conformance.go#L104. The tests are then run in parallel. In my understanding that is also why the tests with Serial tag must be skipped because they can't be run in parallel. Should I add the comment on the code here?

lentzi90: Thanks! Yes please, put a comment there with link to the CAPI test :slightly_smiling_face:

erjavaskivuori:
The tests are still running but I noticed that there is some issue with BMHs not coming available:

```
15:26:40    INFO: Calling postNamespaceCreated for namespace k8s-conformance-8nb1iz
15:26:40    Listing BareMetalHosts
15:26:40    Name:     Status:        Consumer:    Online:  
15:26:40    node-0    registering                 false    
15:26:40    node-1    registering                 false    
15:26:40    node-2                                false    
15:26:40    node-3                                false    
15:26:40    node-4                                false    
15:26:40    node-5                                false    
15:26:40  
15:26:40    INFO: Waiting for 6 BMHs to be in available state
16:16:55    [FAILED] in [BeforeEach] - /home/metal3ci/tested_repo/test/e2e/common.go:478 @ 06/09/25 09:16:40.705
```

After the earlier test run I changed so that BMHs aren't applied in the dev-env (because then they would just be deleted first thing in the test because they have to be re-applied in the new namespace) and I am wondering if that might have caused this? Nothing else has changed so this is a bit odd

erjavaskivuori:
Well let's see what happens

/test metal3-centos-e2e-feature-test-main-pivoting metal3-ubuntu-e2e-feature-test-main-pivoting

erjavaskivuori:
Must have been Monday brain because these aren't even supposed to work on centos because docker is needed... On ubuntu the tests passed but failed on AfterEach. Have to figure out what to do for that.

erjavaskivuori:
There was just some mismatch with my configs and the pivoting pipeline but that is now figured out and the test passed /cc @lentzi90 @tuminoid

tuminoid:
/hold
Holding as it contains the test commit.

test/e2e/data/kubetest/conformance.yaml
```diff
@@ -0,0 +1,11 @@
+ginkgo.focus: \[Conformance\]
+ginkgo.skip: \[Serial\]
+disable-log-dump: true
+ginkgo.slow-spec-threshold: 120s
```
lentzi90, line 4: We get some warnings from this:
```
     --ginkgo.slow-spec-threshold is deprecated --slow-spec-threshold has been deprecated and will be removed in a future version of Ginkgo.  This feature has proved to be more noisy than useful.  You can use --poll-progress-after, instead, to get more actionable feedback about potentially slow specs and understand where they might be getting stuck.
```
Since it is directly from CAPI, which probably got it from k8s, I think we should leave it for now. But it would be a good follow up to check if we can do something about it and then also fix it for CAPI.

lentzi90 (commented):
Very good!
Please remove the test commit
/approve

test/e2e/data/kubetest/conformance.yaml
erjavaskivuori: I saw this too and I can look into this later.

tuminoid:

Not related, but let's run them anyways.

tuminoid:
/unhold
Test commit has been removed.

PR #2579, opened 2025-06-01
🌱 nilnil linter

SD-13:

Fixes #

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -77,6 +77,9 @@ const (
 var (
 	// Capm3FastTrack is the variable fetched from the CAPM3_FAST_TRACK environment variable.
 	Capm3FastTrack    = os.Getenv("CAPM3_FAST_TRACK")
+	ErrAnnotationsNotFound = errors.New("annotations not found")
+	ErrHostIsEmpty = errors.New("Host is empty")
+	ErrCredentialsSecretIsEmpty = errors.New("Credentials secret is empty")
 	errNotFound       *NotFoundError
 	associateBMHMutex sync.Mutex
 )
@@ -761,11 +764,12 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
 ) (*bmov1alpha1.BareMetalHost, error) {
 	annotations := m3Machine.ObjectMeta.GetAnnotations()
 	if annotations == nil {
-		return nil, nil
+		return nil, ErrAnnotationsNotFound
 	}
 	hostKey, ok := annotations[HostAnnotation]
+	var ErrHostKeyNotFound = errors.New("host key not found")
 	if !ok {
-		return nil, nil
+		return nil, ErrHostKeyNotFound
 	}
 	hostNamespace, hostName, err := cache.SplitMetaNamespaceKey(hostKey)
 	if err != nil {
@@ -779,9 +783,11 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
 		Namespace: hostNamespace,
 	}
 	err = cl.Get(ctx, key, &host)
+
+	var ErrAnnotatedHostNotFound = errors.New("Annotated host not found")
 	if apierrors.IsNotFound(err) {
 		mLog.Info("Annotated host not found", "host", hostKey)
-		return nil, nil
+		return nil, ErrAnnotatedHostNotFound
 	} else if err != nil {
 		return nil, err
 	}
@@ -1008,8 +1014,11 @@ func (m *MachineManager) nodeReuseLabelExists(_ context.Context, host *bmov1alph
 
 // getBMCSecret will return the BMCSecret associated with BMH.
 func (m *MachineManager) getBMCSecret(ctx context.Context, host *bmov1alpha1.BareMetalHost) (*corev1.Secret, error) {
-	if host == nil || host.Spec.BMC.CredentialsName == "" {
-		return nil, nil
+
+	if host == nil {
+		return nil, ErrHostIsEmpty
+	} else if host.Spec.BMC.CredentialsName == "" {
+		return nil, ErrCredentialsSecretIsEmpty
 	}
 	tmpBMCSecret := corev1.Secret{}
 	key := host.CredentialsKey()
--- a/baremetal/metal3remediation_manager.go
+++ b/baremetal/metal3remediation_manager.go
@@ -88,6 +88,7 @@ var outOfServiceTaint = &corev1.Taint{
 	Value:  "nodeshutdown",
 	Effect: corev1.TaintEffectNoExecute,
 }
+var ErrNodeIsNotFound = errors.New("Node is not found")
 
 // RemediationManager is responsible for performing remediation reconciliation.
 type RemediationManager struct {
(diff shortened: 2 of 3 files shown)


SD-13:
@tuminoid I am new to golang and still learning semantic errors. I would really appreciate if you can leave some early feedback on this. Thanks!

tuminoid:
> @tuminoid I am new to golang and still learning semantic errors. I would really appreciate if you can leave some early feedback on this. Thanks!

Sorry I missed your message.
/ok-to-test

This needs some time to review as returning errors instead of nil. nil might break the system as sometimes we have cases nil, nil is the expected output.  For example in BMO nilnil PR: https://github.com/metal3-io/baremetal-operator/pull/2514 we had these cases.

I don't have time to review it before my vacations, so I'll ping other guys to check it.
/cc @adilGhaffarDev @lentzi90

baremetal/metal3machine_manager.go
```diff
@@ -761,11 +764,12 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
 ) (*bmov1alpha1.BareMetalHost, error) {
 	annotations := m3Machine.ObjectMeta.GetAnnotations()
 	if annotations == nil {
-		return nil, nil
+		return nil, ErrAnnotationsNotFound
```
lentzi90: This is not really something we consider an error. We need to keep it as `nil` and tell the linter to ignore it instead.
```suggestion
		return nil, nil //nolint:nilnil
```

baremetal/metal3machine_manager.go
```diff
@@ -761,11 +764,12 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
...
 	}
 	hostKey, ok := annotations[HostAnnotation]
+	var ErrHostKeyNotFound = errors.New("host key not found")
 	if !ok {
-		return nil, nil
+		return nil, ErrHostKeyNotFound
```
lentzi90:
```suggestion
		return nil, nil //nolint:nilnil
```

baremetal/metal3machine_manager.go
```diff
@@ -779,9 +783,11 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
 		Namespace: hostNamespace,
 	}
 	err = cl.Get(ctx, key, &host)
+
+	var ErrAnnotatedHostNotFound = errors.New("Annotated host not found")
 	if apierrors.IsNotFound(err) {
 		mLog.Info("Annotated host not found", "host", hostKey)
-		return nil, nil
+		return nil, ErrAnnotatedHostNotFound
```
lentzi90: This is also deliberate. We ignore the error only if the BMH is not found. All other errors are returned.
```suggestion
		return nil, nil //nolint:nilnil
```

baremetal/metal3machine_manager.go
```diff
@@ -1008,8 +1014,11 @@ func (m *MachineManager) nodeReuseLabelExists(_ context.Context, host *bmov1alph
 
 // getBMCSecret will return the BMCSecret associated with BMH.
 func (m *MachineManager) getBMCSecret(ctx context.Context, host *bmov1alpha1.BareMetalHost) (*corev1.Secret, error) {
-	if host == nil || host.Spec.BMC.CredentialsName == "" {
-		return nil, nil
+
+	if host == nil {
+		return nil, ErrHostIsEmpty
+	} else if host.Spec.BMC.CredentialsName == "" {
+		return nil, ErrCredentialsSecretIsEmpty
```
lentzi90: This I think is fine, but please remove the extra newlines above.

baremetal/metal3remediation_manager.go
```diff
@@ -391,7 +392,7 @@ func (r *RemediationManager) GetNode(ctx context.Context, clusterClient v1.CoreV
 
 	node, err := clusterClient.Nodes().Get(ctx, capiMachine.Status.NodeRef.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		return nil, nil
+		return nil, ErrNodeIsNotFound
```
lentzi90:
```suggestion
		return nil, nil //nolint:nilnil
```

baremetal/utils.go
```diff
@@ -199,8 +203,9 @@ func fetchM3DataTemplate(ctx context.Context,
 	clusterName string,
 ) (*infrav1.Metal3DataTemplate, error) {
 	// If the user did not specify a Metal3DataTemplate, just keep going.
+	var ErrTemplateRefIsNil = errors.New("Template ref is nil")
 	if templateRef == nil {
-		return nil, nil
+		return nil, ErrTemplateRefIsNil
```
lentzi90:
```suggestion
		return nil, nil //nolint:nilnil
```

baremetal/utils.go
```diff
@@ -294,7 +300,7 @@ func getM3Machine(ctx context.Context, cl client.Client, mLog logr.Logger,
 				mLog.Info(errMessage)
 				return nil, WithTransientError(errors.New(errMessage), requeueAfter)
 			}
-			return nil, nil
+			return nil, ErrMetal3MachineNotFound
```
lentzi90:
```suggestion
			return nil, nil //nolint:nilnil
```

baremetal/utils.go
```diff
@@ -306,14 +312,14 @@ func getM3Machine(ctx context.Context, cl client.Client, mLog logr.Logger,
...
+		return nil, ErrMissingDataTemplateName
 	}
 	if tmpM3Machine.Spec.DataTemplate.Name != dataTemplate.Name {
-		return nil, nil
+		return nil, ErrDataTemplateNameMismatch
 	}
 	if tmpM3Machine.Spec.DataTemplate.Namespace != "" &&
 		tmpM3Machine.Spec.DataTemplate.Namespace != dataTemplate.Namespace {
-		return nil, nil
+		return nil, ErrDataTemplateNamespaceMismatch
```
lentzi90: These I am not so sure about. I think we could clean them up but that would be a bit bigger refactor.
For now I think the better way forward is to put `nil //nolint:nilnil` also here.

SD-13:
Hello @lentzi90 Thanks for reviewing my PR!
I addressed your comments, can you please take another look?
Thanks!

baremetal/metal3machine_manager.go
```diff
@@ -1008,8 +1008,10 @@ func (m *MachineManager) nodeReuseLabelExists(_ context.Context, host *bmov1alph
 
 // getBMCSecret will return the BMCSecret associated with BMH.
 func (m *MachineManager) getBMCSecret(ctx context.Context, host *bmov1alpha1.BareMetalHost) (*corev1.Secret, error) {
-	if host == nil || host.Spec.BMC.CredentialsName == "" {
-		return nil, nil
+	if host == nil {
+		return nil, errors.New("Host is empty")
+	} else if host.Spec.BMC.CredentialsName == "" {
+		return nil, errors.New("Credentials secret is empty")
```
lentzi90: Looks like this is still causing a unit test to fail. We will need to either change that unit test or revert and put nolint here as well

SD-13: I am trying to modify that unit test accordingly.

SD-13: I added nolint for now.

lentzi90 (commented):
/approve

PR #2592, opened 2025-06-05
✨ Add parameters blob for bond network interface definition (backport)

tmmorin:
This PR is a backport of #2555 (direct cherry-pick)


tuminoid:
> This PR is a backport of #2555 (direct cherry-pick)

Any reason to not use cherry-pick bot, if that is the case?

Also, how do we feel about backporting :sparkles: ? 
/cc @lentzi90 @Rozzii

tmmorin:
> Any reason to not use cherry-pick bot, if that is the case?

No good reason.

I actually wondered if there was a typical way of doing backports in metal3, searched a bit, and  finding only https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/CONTRIBUTING.md#backporting-a-patch which did not mention it, I did this PR manually.

Looking a bit closer now, I see that [another section](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/CONTRIBUTING.md#backporting) of the same document mentions the cherry-pick bot...

tmmorin:
I [just tried](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2555#issuecomment-2948601086) to use the bot, and he doesn't like me  ;)

> @tmmorin: only [metal3-io](https://github.com/orgs/metal3-io/people) org members may request cherry picks.

Can I become one or is this something more for maintainers and experienced contributors ?

tuminoid:
> > Any reason to not use cherry-pick bot, if that is the case?
> 
> No good reason.

No biggie, I see you now tried to initiate it later and it was not available to you.

> I actually wondered if there was a typical way of doing backports in metal3, searched a bit, and finding only https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/CONTRIBUTING.md#backporting-a-patch which did not mention it, I did this PR manually.
> 
> Looking a bit closer now, I see that [another section](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/CONTRIBUTING.md#backporting) of the same document mentions the cherry-pick bot...

You are correct that the contribution documentation is not up to par. We cleaned it half-way when creating community repo (see the ladder and roles etc docs), but the actual contributing.md is lacking. Since we host quite a few repos here, we should move it under the community, update it and point to it from repos, so each repo has consistent guidance.

I've created https://github.com/metal3-io/community/issues/28 to track this.

> I [just tried](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2555#issuecomment-2948601086) to use the bot, and he doesn't like me ;)
> 
> > @tmmorin: only [metal3-io](https://github.com/orgs/metal3-io/people) org members may request cherry picks.
> 
> Can I become one or is this something more for maintainers and experienced contributors ?

Membership requirements and process are detailed in contributor ladder document:
https://github.com/metal3-io/community/blob/main/CONTRIBUTOR-LADDER.md#organization-member

Rozzii:
I am not found of back-porting new functionality, this is a brand new thing so unless it solves a security issue or fixes an existing functionality I would not support this backport. This feature is equivalent to a change that would allow the use of e.g an existing Ironic feature etc...

tmmorin:
> I am not found of back-porting new functionality, this is a brand new thing so unless it solves a security issue or fixes an existing functionality I would not support this backport. 

Yes, this is an argument that I can very well understand.

I made this PR to open the discussion, thinking that the change may still qualifies as a backport given that it relates to a fully optional feature, and does not change any existing schema/behavior when the feature isn't used.  Could that in your view be a criteria to make an exception ?

If not, my question would be: in a context where we wouldn't be able to quickly move to CAPI 1.11 / CAPM3 1.11, and where we'd have to use CAPI 1.10, could we use capm3 1.11 to have this feature ?

> This feature is equivalent to a change that would allow the use of e.g an existing Ironic etc...

(sorry, I did not understand this part)

tuminoid:
> I made this PR to open the discussion, thinking that the change may still qualifies as a backport given that it relates to a fully optional feature, and does not change any existing schema/behavior when the feature isn't used. Could that in your view be a criteria to make an exception ?

Would you be able to join the community meeting next Wednesday 14.00 UTC and discuss it?

Rozzii:
Related to this : 
> > This feature is equivalent to a change that would allow the use of e.g an existing Ironic etc...
> 
> (sorry, I did not understand this part)

I mean that introducing an optional parameter for cloud-init is a new provisioning functionality, same as if someone would make an Ironic feature available in BMO. I just wanted to put this change in context. 

Confusion might be because I wanted to write  "existing Ironic feature" but the word "feature" was missing . I have fixed that comment.

lentzi90:
I don't think we should backport :sparkles: . If we consider the partially implemented API a bug, we may consider it, but I would much rather keep this for the next minor release. This also makes it easier to pin-point any bugs that may appear

Rozzii:
>  If we consider the partially implemented API a bug, we may consider it, but...

IMO partially propagating features of "lower level" services can't be considered as bug because then we would need to say that,  we are buggy until all the features of Ironic and all the cloud-init features and all the ignition features etc... are not propagated. So IMO "opening up" or propagating lower level functionality to capm3 level can't be viewed as bug, it is a feature and even a cloud-init specific feature AFAIK.

tmmorin:
It's certainly hard to argue that adding this ability would be a "bug fix" .  I think the best way to describe this is that it addresses a limitation that prevents using Metal3 in some contexts (the context that lead us to raise this problem is a context where not being able to set mii_mon would prevent the deployment, but maybe other people would see different cases). 

I'm absolutely fine if the answer is "let's not backport this" - you are the ones who decide where how you want to manage the risk/benefit balance for maintenance releases.

Rozzii:
> It's certainly hard to argue that adding this ability would be a "bug fix" . I think the best way to describe this is that it addresses a limitation that prevents using Metal3 in some contexts (the context that lead us to raise this problem is a context where not being able to set mii_mon would prevent the deployment, but maybe other people would see different cases).
> 
> I'm absolutely fine if the answer is "let's not backport this" - you are the ones who decide where how you want to manage the risk/benefit balance for maintenance releases.

User can re-parse the the network-data file, with the knowledge of which interface the user is  looking for and the user can use any linux networking tool  to extend the network configuration or to find the the network manager files generated by cloud-init and edit them.

It is true that the parameter option is much much much more user friendly, but because we support pre and post kubeadm scripts any network management logic can be run and the config drive also stays on the disk by default so the user  can at any time re-use both the netwok_data or the meta_data . I understand the issue but it is simply not enough IMO to facilitate a backport.

That said, I will present this topic on today's community meeting, let's hear the community's voice before I would cut the next patch release of this branch.

tmmorin:
> User can re-parse the the network-data file, with the knowledge of which interface the user is looking for and the user can use any linux networking tool to extend the network configuration or to find the the network manager files generated by cloud-init and edit them.

Yes, there are workarounds.

In our project (Sylva) a colleague developped a Kyverno policy to patch the network data secret on the fly (https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/1730/diffs#a5c541bc2f5a91ace381e05dc351532cc811ef48).

We'd just love our infra-as-code better if it's not infra-as-spaghetti-code ;)
 
> I understand the issue but it is simply not enough IMO to facilitate a backport.

Understood.

> That said, I will present this topic on today's community meeting, let's hear the community's voice before I would cut the next patch release of this branch.

If your policy had been "we accept backports for optional behavior identified as not risking regressions when the behavior is inactive" then this PR would have possibly qualified,  but since it is not, then this PR doesn't qualify.   I'm really not asking for making an exception to the rule here ;-)

Rozzii:
I will close this pull request then as we have came to a conclusion.
Thanks for the understanding @tmmorin and thanks everyone for the great discussion.
/close

PR #2596, opened 2025-06-09
🌱Add vars for feature tests and improve the script

smoshiur1237:
Add vars for pivoting, remediation and upgrade test. Also removing `if else `statement and replace it with `case` for better readability.

### DIFF
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -58,25 +58,35 @@ export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
 export CAPM3_LOCAL_IMAGE="${CAPM3PATH}"
 
 # Upgrade test environment vars and config
-if [[ ${GINKGO_FOCUS:-} == "clusterctl-upgrade" ]]; then
-  export NUM_NODES="5"
-fi
+case "${GINKGO_FOCUS:-}" in
+  # clusterctl upgrade var
+  clusterctl-upgrade)
+    export NUM_NODES="5"
+  ;;
 
-# Scalability test environment vars and config
-if [[ ${GINKGO_FOCUS:-} == "scalability" ]]; then
-  export NUM_NODES=${NUM_NODES:-"10"}
-  export BMH_BATCH_SIZE=${BMH_BATCH_SIZE:-"2"}
-  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"1"}
-  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"0"}
-  export KUBERNETES_VERSION_UPGRADE_FROM=${FROM_K8S_VERSION}
-fi
+  # Integration test environment vars and config
+  basic|integration)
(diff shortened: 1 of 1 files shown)


tuminoid:
Still hitting the same

```
[2025-06-09T09:09:03.834Z] • [FAILED] [3639.542 seconds]
[2025-06-09T09:09:03.834Z] Testing nodes remediation [remediation] [features] [It] Should create a cluster and run remediation based tests [remediation, features]
[2025-06-09T09:09:03.834Z] /home/metal3ci/tested_repo/test/e2e/remediation_based_feature_test.go:73
[2025-06-09T09:09:03.834Z] 
[2025-06-09T09:09:03.834Z]   [FAILED] Timed out after 1200.001s.
[2025-06-09T09:09:03.834Z]   Expected
[2025-06-09T09:09:03.834Z]       <[]string | len:3, cap:5>: ["node_1", "node_0", "node_2"]
[2025-06-09T09:09:03.834Z]   to contain elements
[2025-06-09T09:09:03.834Z]       <[]string | len:1, cap:1>: ["node_3"]
[2025-06-09T09:09:03.834Z]   the missing elements were
[2025-06-09T09:09:03.834Z]       <[]string | len:1, cap:1>: ["node_3"]
[2025-06-09T09:09:03.834Z]   In [It] at: /home/metal3ci/tested_repo/test/e2e/remediation.go:420 @ 06/09/25 09:02:15.629
```

scripts/environment.sh
```diff
@@ -58,25 +58,35 @@ export M3PATH=${M3PATH:-"${HOME}/go/src/github.com/metal3-io"}
...
+  ;;
+
+  # Scalability test environment vars and config
+  scalability)
+    export NUM_NODES=${NUM_NODES:-"10"}
+    export BMH_BATCH_SIZE=${BMH_BATCH_SIZE:-"2"}
+    export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"1"}
+    export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"0"}
+    export KUBERNETES_VERSION_UPGRADE_FROM=${FROM_K8S_VERSION}
+  ;;
```
tuminoid, line 88: Btw, should we add a default case here, that would exit if none of the focus cases are met ?

smoshiur1237: do you use any default condition in our e2e test. It seems all of them have specific vars. what are you suggesting? do I add case `*)` and `exit 0`?

smoshiur1237: It doesn't look like it is failing in this conditions

tuminoid: > do you use any default condition in our e2e test. It seems all of them have specific vars. what are you suggesting? do I add case `*)` and `exit 0`?

Yes, `*)` with error message and `exit 1`, if it safe to do (ie. there is no config that leaves it empty).

tuminoid: > It doesn't look like it is failing in this conditions

Not here, but all these branches we've now fixed these conditions couple times as we've missed some cases. We can add it in another PR to catch these.

PR #2607, opened 2025-06-13
:seedling: fix ephemeral worker/cp count

tuminoid:
Ephemeral has GINKGO_FOCUS == features, so it is falling thru here and runs with default 1+1 from common.sh.

Add an else to the features check, to have use the right amount of control planes nodes.

Move the k8s-conformance to the case as well.

### DIFF
--- a/scripts/environment.sh
+++ b/scripts/environment.sh
@@ -86,23 +86,37 @@ case "${GINKGO_FOCUS:-}" in
     export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"0"}
     export KUBERNETES_VERSION_UPGRADE_FROM=${FROM_K8S_VERSION}
   ;;
+
+  # IPReuse feature test environment vars and config
+  features)
+    if [[ "${GINKGO_SKIP:-}" == "pivoting remediation" ]]; then
+      export NUM_NODES="5"
+      export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"3"}
+      export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"2"}
+    else
+      export NUM_NODES=${NUM_NODES:-"4"}
+      export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"3"}
+      export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"1"}
+    fi
+  ;;
+
+  # K8s conformance test environment vars and config
+  # The test needs 6 nodes in total because the tests run in parallel on 5 worker machines:
+  # https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/k8s_conformance.go#L104
+  k8s-conformance)
+    export NUM_NODES="6"
+    export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"1"}
+    export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"5"}
+  ;;
+
+  *)
+    # unknown GINKGO_FOCUS, let's print out the crucial env and continue
+    echo "WARNING: unrecognized GINKGO_FOCUS='${GINKGO_FOCUS:-}'"
+  ;;
 esac
 
-# IPReuse feature test environment vars and config
-if [[ ${GINKGO_FOCUS:-} == "features" && ${GINKGO_SKIP:-} == "pivoting remediation" ]]; then
-  export NUM_NODES="5"
-  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"3"}
-  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"2"}
-fi
-
-# K8s conformance test environment vars and config
-# The test needs 6 nodes in total because the tests run in parallel on 5 worker machines:
-# https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/k8s_conformance.go#L104
-if [[ ${GINKGO_FOCUS:-} == "k8s-conformance" ]]; then
-  export NUM_NODES="6"
-  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"1"}
-  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"5"}
-fi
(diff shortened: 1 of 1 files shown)


tuminoid:
/test metal3-centos-e2e-feature-test-main-remediation

Hmm, we don't have ephemeral as PR job, nor k8s-conformance.

scripts/environment.sh
```diff
@@ -86,23 +86,29 @@ case "${GINKGO_FOCUS:-}" in
...
   ;;
-esac
 
-# IPReuse feature test environment vars and config
-if [[ ${GINKGO_FOCUS:-} == "features" && ${GINKGO_SKIP:-} == "pivoting remediation" ]]; then
-  export NUM_NODES="5"
-  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"3"}
-  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"2"}
-fi
+  # K8s conformance test environment vars and config
```
tuminoid: I'll rotate the cases to minimize the diff, if this ends up fixing the issue.

adilGhaffarDev:
> Hmm, we don't have ephemeral as PR job, nor k8s-conformance.

 pr is not merged for k8s-conformance. https://gerrit.nordix.org/c/infra/cicd/+/23763 
And I think because ephemeral is too big.

tuminoid:
> > Hmm, we don't have ephemeral as PR job, nor k8s-conformance.
> 
> pr is not merged for k8s-conformance. https://gerrit.nordix.org/c/infra/cicd/+/23763 

OK, let's check that.

>And I think because ephemeral is too big.

We have now so much resources, we should reconsider for cases like this. It wouldn't be run too often from PRs, but it is bad that we cannot verify changes related to it. I'll open a PR to enable them, let's continue discussion there.

tuminoid:
> > > Hmm, we don't have ephemeral as PR job, nor k8s-conformance.
> > 
> > pr is not merged for k8s-conformance. https://gerrit.nordix.org/c/infra/cicd/+/23763
> 
> OK, let's check that.

That is LGTM +2, I've pinged @lentzi90 to PTAL and submit.

tuminoid:
> > > > Hmm, we don't have ephemeral as PR job, nor k8s-conformance.
> > > 
> > > 
> > > pr is not merged for k8s-conformance. https://gerrit.nordix.org/c/infra/cicd/+/23763
> > 
> > 
> > OK, let's check that.
> 
> That is LGTM +2, I've pinged @lentzi90 to PTAL and submit.

Done. Now we need to add the Prow keywords. @erjavaskivuori can you add those?

scripts/environment.sh
```diff
@@ -86,23 +86,29 @@ case "${GINKGO_FOCUS:-}" in
...
-# K8s conformance test environment vars and config
-# The test needs 6 nodes in total because the tests run in parallel on 5 worker machines:
-# https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/k8s_conformance.go#L104
-if [[ ${GINKGO_FOCUS:-} == "k8s-conformance" ]]; then
-  export NUM_NODES="6"
-  export CONTROL_PLANE_MACHINE_COUNT=${CONTROL_PLANE_MACHINE_COUNT:-"1"}
-  export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"5"}
-fi
+  # IPReuse feature test environment vars and config
+  features)
```
Rozzii, line 91: I was confused for a minute that how come this is called feature while we have the different features in here https://github.com/metal3-io/cluster-api-provider-metal3/pull/2607/files#diff-f3736fd0d9f6c379e7622c6230d742bb019c1120223e1c3857bb1808d27d5413L75
But now I see this is ipreuse test. so how come this needs a bigger env if the pivot and remediation are skipped?

tuminoid: I just kept the existing numbers there and used the numbers @adilGhaffarDev said are correct for this. Note that these are also fallback variables, so they might be overridden somewhere else like JJB. Based on the remediation job output, these 3+1 numbers are correctly used now though.

Rozzii: yeah all makes sense, I am just confused by the 5 node requirement for any case  but Adil might know the reason.

tuminoid:
Remediation tests are now working as they should, but they will fail now again at the node shutdown/boot-up issue, which is the root cause of this 10 day failure streak. I will let the tests finish, and push the update.

tuminoid:
> Remediation tests are now working as they should, but they will fail now again at the node shutdown/boot-up issue, which is the root cause of this 10 day failure streak. I will let the tests finish, and push the update.

Hmm, actually remediation has its own case already so we don't actually know if this will help with ephemeral case. We need that Prow keyword. @adilGhaffarDev

scripts/environment.sh
```diff
@@ -86,23 +86,37 @@ case "${GINKGO_FOCUS:-}" in
...
+    export WORKER_MACHINE_COUNT=${WORKER_MACHINE_COUNT:-"5"}
+  ;;
+
+  *)
+    # unknown GINKGO_FOCUS, let's print out the crucial env and continue
+    echo "WARNING: unrecognized GINKGO_FOCUS='${GINKGO_FOCUS:-}'"
+  ;;
+esac
+
+echo -e "\nINFO: GINKGO_FOCUS='${GINKGO_FOCUS:-}' GINKGO_SKIP='${GINKGO_SKIP:-}'"
```
tuminoid, line 118: I want to leave this here, so we can later on actually get a hang of what focus and config combo we run at on a glance.

tuminoid:
/test metal3-ubuntu-e2e-feature-test-main-features

Remediation will fail to unrelated issues, so I don't bother running it.

scripts/environment.sh
adilGhaffarDev: 5 nodes are only for IP-reuse and clusterctl upgrade tests. 
Here is where we changed it for ip-reuse: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2201 
For clusterctl, we are using the CAPI test, which I believe requires 5 nodes.

For IP-reuse we can improve it, maybe try to create a similar scenario with fewer nodes where we can make sure IP-reuse is working. But I need to check that. I remember there were some issues.

adilGhaffarDev:
> Remediation will fail to unrelated issues, so I don't bother running it.

This will fix the cp and worker count issue, let's merge this we can check remediation in other pr.
/lgtm

lentzi90:
/hold
Did you want to wait for those tests?

tuminoid:
> /hold Did you want to wait for those tests?

Those should pass, so we can wait for them.

tuminoid:
> > /hold Did you want to wait for those tests?
> 
> Those should pass, so we can wait for them.

Can see though already from the logging of all these 4 jobs that the counts are correct for all these combinations, so if something fails unexpectedly, we can override it to get this merged as the failure is not related.

PR #2616, opened 2025-06-17
🌱 Fixing remediation tests

adilGhaffarDev:
**What this PR does / why we need it**:
Fixing remediation tests.
Always get BMH before changing, so we make sure it is not an outdated or stale object.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2618, opened 2025-06-18
:seedling: Fix deleting BMHs in md_remediation

erjavaskivuori:
**What this PR does / why we need it**: Improve deleting BMHs in BeforeEach. Before BMHs weren't deleted conditionally and test failed if there wasn't any BMHs applied yet. Removed unnecessary reapplying of BMHs after the test.

Fixes #

### DIFF
--- a/test/e2e/md_remediations_test.go
+++ b/test/e2e/md_remediations_test.go
@@ -2,16 +2,13 @@ package e2e
 
 import (
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 
-	bmov1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/utils/ptr"
 	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -19,6 +16,12 @@ const (
 )
 
 var _ = Describe("When testing MachineDeployment remediation [healthcheck] [remediation] [features]", Label("healthcheck", "remediation", "features"), func() {
+
+	var (
+		specName  = "metal3"
+		namespace = "metal3"
+	)
+
 	BeforeEach(func() {
 		osType := strings.ToLower(os.Getenv("OS"))
 		Expect(osType).ToNot(Equal(""))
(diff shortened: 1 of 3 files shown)


erjavaskivuori:
This fix is working as expected for md_remediation but the issue with other remediation test is fixed in #2616 and the remediation tests probably won't pass before those changes are merged.

tuminoid:
> This fix is working as expected for md_remediation but the issue with other remediation test is fixed in #2616 and the remediation tests probably won't pass before those changes are merged.

2616 is merged, please rebase.

erjavaskivuori:
The actual tests passed, maybe the centos tests can be overridden if this is merged before the issue is fixed.

test/e2e/md_remediations_test.go
```diff
@@ -46,25 +66,7 @@ var _ = Describe("When testing MachineDeployment remediation [healthcheck] [reme
...
-		ListBareMetalHosts(ctx, bootstrapClient, client.InNamespace(namespace))
-		WaitForNumBmhInState(ctx, bmov1alpha1.StateAvailable, WaitForNumInput{
-			Client:    bootstrapClient,
-			Options:   []client.ListOption{client.InNamespace(namespace)},
-			Replicas:  4,
-			Intervals: e2eConfig.GetIntervals(specName, "wait-bmh-available"),
-		})
-		ListBareMetalHosts(ctx, bootstrapClient, client.InNamespace(namespace))
+		By("Recreate BMHs used in CAPI namespace in metal3 namespace")
+		ApplyBmh(ctx, e2eConfig, bootstrapClusterProxy, namespace, specName)
```
adilGhaffarDev: If this test runs last, it ends up applying BMHs that are never used.
A better approach would be for each test that requires BMHs in the `metal3` namespace to ensure they exist by creating them in its `BeforeEach` block if they are missing.

adilGhaffarDev:
> The actual tests passed, maybe the centos tests can be overridden if this is merged before the issue is fixed.

I can override the tests

test/e2e/md_remediations_test.go
```diff
@@ -1,38 +1,58 @@
...
 		Expect(osType).ToNot(Equal(""))
 		validateGlobals(specName)
 		// We need to override clusterctl apply log folder to avoid getting our credentials exposed.
 		clusterctlLogFolder = filepath.Join(os.TempDir(), "target_cluster_logs", bootstrapClusterProxy.GetName())
 
-		Logf("Removing existing BMHs from source")
-		bmhData, err := os.ReadFile(filepath.Join(workDir, bmhCrsFile))
-		Expect(err).ToNot(HaveOccurred(), "BMH CRs file not found")
-		kubeConfigPath := bootstrapClusterProxy.GetKubeconfigPath()
-		err = KubectlDelete(ctx, kubeConfigPath, bmhData, "-n", "metal3")
```
lentzi90, line 33: We already had `ctx` here, where was it coming from, and why do we now need to define it at the top?
Ideally, I think we should set the context in one place globally for the whole test suite like they [do in CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/4852b676b3ff1adfcd9aa946d4bba883f0339a00/test/e2e/e2e_suite_test.go#L74).
(Not necessarily something that we should do as part of this PR though, if it turns out to require a lot of changes.)

test/e2e/md_remediations_test.go
```diff
@@ -45,26 +65,4 @@ var _ = Describe("When testing MachineDeployment remediation [healthcheck] [reme
 			Flavor:                ptr.To(osType + "-md-remediation"),
 		}
 	})
-	AfterEach(func() {
-		ListBareMetalHosts(ctx, bootstrapClusterProxy.GetClient(), client.InNamespace(namespace))
-		ListMetal3Machines(ctx, bootstrapClusterProxy.GetClient(), client.InNamespace(namespace))
-		ListMachines(ctx, bootstrapClusterProxy.GetClient(), client.InNamespace(namespace))
```
lentzi90, line 51: This part is still useful I think. It gives a little overview of the current state at the end of the test and can be useful to explain issues that may happen elsewhere.

test/e2e/md_remediations_test.go
erjavaskivuori: Yeah I was not quite sure what approach to take here. Previously it came from `pivoting_based_feature_test.go` which also seemed a weird place to get the context from and could cause weird bugs if it were to change (? not sure if that would ever happen though). Same was for the specName and namespace. I think these all are the same in most, if not all, tests so it would make sense to define them in some common place. It seems that many tests currently get those values from `pivoting_based_feature_test.go`.

I noticed that ctx, specName and namespace are defined in the function like this in `remediation_based_feature_test.go` and `upgrade_kubernetes_test.go` which is why I decided to do this.

test/e2e/md_remediations_test.go
erjavaskivuori: I was debating whether it is useful or not. It lists the BMHs and machines only in metal3 namespace and if there would be any things would have failed in BeforeEach when we are deleting the BMHs. So well in the case that fails we could get the list of remaining BMHs and machines but otherwise it will always be empty :thinking:

test/e2e/md_remediations_test.go
lentzi90: Good, then I'm fine to keep the `context.TODO()` for now. It would be a nice cleanup to fix this for all the tests later.

test/e2e/md_remediations_test.go
lentzi90: Ah I see! This was probably added when metal3 was the only namespace we used. Maybe we can change it to list them in all namespaces then? Or the test namespace (if that is available in the AfterEach).

erjavaskivuori: The namespace used in the the test is deleted in the test's cleanup in the AfterEach so I am not sure if there would be any namespaces where the machines would be found from but it might be reasonable to check the all namespaces anyhow. I actually tried that at some point and couldn't figure out how to do it :sweat_smile:  But I didn't spend much time on it so I could take another look

adilGhaffarDev: When we run CAPI tests, they clean up everything on the CAPI side, so we won't see any resources in CAPM3 during the AfterEach phase. We could enable skipCleanup and handle the cleanup ourselves in order to inspect and print BMHs, but I'm not sure it's worth it.

Ideally, CAPI would provide a hook just before the cleanup process where we could print BMHs. I'm not sure if such a hook exists.

erjavaskivuori: I took it off

lentzi90 (commented):
/lgtm

PR #2637, opened 2025-06-30
:book: update docs/releasing.md on dependabot configs

tuminoid:
Update docs/releasing.md on dependabot config changes while releasing new minors.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -151,6 +151,23 @@ release!
 
 Some post-release actions are needed if new minor or major branch was created.
 
+### Dependabot configuration
+
+In `main` branch, we need to update Dependabot configuration to allow updates
+to release branch dependencies and GitHub Workflows.
+
+If project dependencies or modules have not changed, we copy the previous release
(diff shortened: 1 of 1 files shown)


docs/releasing.md
```diff
@@ -151,6 +151,23 @@ release!
 
 Some post-release actions are needed if new minor or major branch was created.
 
+### Dependabot configuration
+
+In `main` branch, we need to update Dependabot configuration to allow updates
+to release branch dependencies and GitHub Workflows.
+
+If project dependencies or modules have not changed, we copy the previous release
```
Rozzii: I would use either "you"  or "one, the release team etc..." or just write it in passive instead of the "we". This "we" was introduced in the "Release artifacts" section and doesn't really fit into a technical document or to an instruction guide. The "we" is also not used outside of this and the "release artrifacts" section so it breaks the cohesion of the text quite a bit.

tuminoid: I'll reword it.

tuminoid: Done

tuminoid:

PR #2640, opened 2025-07-02
:seedling:  Change deprecated ginkgo functionality

erjavaskivuori:
**What this PR does / why we need it**: Remove deprecated --slow-spec-threshold. I tested the suggested --poll-progress-after but for me it seemed that the outputs might not be too useful for us as the time limit exceeds in many tests in normal situation, too. If some test fails it will be anyhow logged.

```
--ginkgo.slow-spec-threshold is deprecated --slow-spec-threshold has been deprecated
and will be removed in a future version of Ginkgo.  This feature has proved to be more
noisy than useful.  You can use --poll-progress-after, instead, to get more actionable 
feedback about potentially slow specs and understand where they might be getting stuck.
```

### DIFF
--- a/test/e2e/data/kubetest/conformance.yaml
+++ b/test/e2e/data/kubetest/conformance.yaml
@@ -1,7 +1,6 @@
 ginkgo.focus: \[Conformance\]
 ginkgo.skip: \[Serial\]
 disable-log-dump: true
-ginkgo.slow-spec-threshold: 120s
 ginkgo.flake-attempts: 3
 ginkgo.trace: true
 ginkgo.v: true


erjavaskivuori:
It was quite challenging to find any documentation about what the deprecated `slow-spec-threshold` would have outputted and it didn't output anything in our case. But it seems to me that we might not want to use the new `poll-progress-after` either. As the name says the `poll-progress-after` will start to poll the progress of a ginkgo process after the set time (in this case 120s) to see what is slowing down the process. Now that I have seen the functionality in action I think it doesn't make much sense to use it in this case. There is very little we can do for the speed of conformance tests, and the logs will now just be filled with this IMO unnecessary info.

The conformance.yaml is the same that CAPI is using since we are using the same spec that runs the kubetest. That is why the `slow-spec-threshold` was included in the first place but I'd say it or any alternative isn't needed.

Run where the --poll-progress-after was used: https://jenkins.nordix.org/view/Metal3/job/metal3-e2e-conformance-test-main/1/consoleFull

kashifest:
Please update the description as well.

PR #2646, opened 2025-07-11
🌱 Limit API listing to 200 at a time via pagination

fmuyassarov:
Control how many objects (`metal3data`,`metal3datatemplate`, `metal3machine`, `metal3machinetemplate`) should be retrieved per list call (200). This should in theory reduce the load on the API server's memory consumption, especially when there is a large number of objects with large amount of data on them.


Rozzii (commented):
/lgtm

schrej:
These list calls do not even reach the API server since controller-runtime returns cached values. In addition this MR does not implement pagination, it just limits the amount of objects requested. If there is more than 200 resources to consider with one of the remaining list calls, this might lead to unintended behavior.

This change was reverted for Metal3Machine host association, but the remaining cases are lingering bugs for large-scale setups as well.

PR #2667, opened 2025-07-28
[TEST] Try Cilium CNI

peppi-lotta:
**What this PR does / why we need it**: Try out Cilium CNI in e2e test


test/e2e/e2e_suite_test.go
```diff
@@ -309,6 +322,53 @@ func updateCalico(config *clusterctl.E2EConfig, calicoYaml, calicoInterface stri
...
+		},
+	}
+
+	manifest, err := generateTemplateFromHelmChart(ctx, opts, manifestOverwriteValues, e2eConfig)
+	Expect(err).ToNot(HaveOccurred(), "failed to generate template: %v", err)
+
+	// Replace ${BIN_PATH} with /opt/cni/bin. This is done to prevent
+	// framework.RegisterClusterResourceSetConfigMapTransformation from throwing
+	// an error due to unresolvable "envsubst" variable.
+	manifest = strings.ReplaceAll(manifest, "${BIN_PATH}", "/opt/cni/bin")
```
peppi-lotta, line 360: Expansion on this addition. `initContainers` has this env variable and command combo and it valid for the manifest.
```
        env:
        - name: CGROUP_ROOT
          value: /run/cilium/cgroupv2
        - name: BIN_PATH
          value: /opt/cni/bin
        command:
        - sh
        - -ec
        - |
          cp /usr/bin/cilium-mount /hostbin/cilium-mount;
          nsenter --cgroup=/hostproc/1/ns/cgroup --mount=/hostproc/1/ns/mnt "${BIN_PATH}/cilium-mount" $CGROUP_ROOT;
          rm /hostbin/cilium-mount
```

But when the `framework.RegisterClusterResourceSetConfigMapTransformation` gets called it runs environment variable substitution and gives error for ${BIN_PATH}. This is why I substitute the value before the function is called. 

https://github.com/kubernetes-sigs/cluster-api/blob/605c98767234e04d2441ee943311640399380561/test/framework/clusterctl/repository.go#L67C1-L73C4

peppi-lotta:
/hold 
Trying some test values so don't want to merge yet :)

Rozzii:
Is it OK to review now @peppi-lotta or are you planning additional changes?

tuminoid:
This is one of the constantly rebuilt PRs. @peppi-lotta can you work on it so we can either merge or close? Thanks.

peppi-lotta:
/hold cancel 

This is ready to go in

test/e2e/config/e2e_conf.yaml
```diff
@@ -274,8 +274,9 @@ variables:
   UPGRADED_BMO_IMAGE_TAG: "${UPGRADED_BMO_IMAGE_TAG:-main}"
 
   PROVIDER_ID_FORMAT: "metal3://{{ ds.meta_data.providerid }}"
-  # Pin Calico version
+  CNI_PROVIDER: "${CNI_PROVIDER:-calico}"
   CALICO_VERSION: "${CALICO_VERSION:-v3.30.3}"
+  CILIUM_VERSION: "${CILIUM_VERSION:-v1.18.0}"
```
lentzi90, line 279: I think I commented before on these. I do not see the benefit of the environment variable syntax here. The e2e framework is already taking values from environment variables as a way of overriding what is in the e2e config, so we could simply put `v1.18.0` here with the same result. We have it for many other things though so I am not going to block on that. It is just something we could clean up at some point.

lentzi90 (commented):
Looks good!
For the future, I think we should consider keeping things simple by just storing the manifests directly in the repo. Now we have ~200 lines of go code for downloading, rendering and modifying manifests from a helm chart. We never see the actual yaml that gets applied so it is hard to tell where the issue is if something goes wrong.
We could instead manually render the manifest and commit it (like we [do in CAPO](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/test/e2e/data/cni/calico.yaml)). It is not like we are going to change the CIDR that is used in these tests very often, so that could be hard coded.

/approve

PR #2679, opened 2025-08-01
:rocket: Release v1.11.0-alpha.0

kashifest:


releasenotes/v1.11.0-alpha.0.md
```diff
@@ -0,0 +1,107 @@
...
+
+## :sparkles: New Features
+
+- Add parameters blob for bond network interface definition (#2555)
+- Refactor setting ProviderID (#2365)
+
+## :bug: Bug Fixes
+
+- Fix trying to patch Metal3Data if it is deleted (#2666)
+- Bump github.com/go-viper/mapstructure/v2 from 2.2.1 to 2.3.0 in /test (#2636)
```
lentzi90: I'm not sure why this is marked as 🐛 but I can see that @tuminoid explicitly did it. Should we change it to 🌱 ?

adilGhaffarDev: also its only in test, I am also leaning towards seedling.

kashifest: ok I can change it, I thought we do the PR list sanitising while doing proper releases (minor/patch) cause it is also missing the superseeding block

adilGhaffarDev: its fine to merge it like this for alpha.

kashifest: Done

adilGhaffarDev:

releasenotes/v1.11.0-alpha.0.md
```diff
@@ -0,0 +1,107 @@
...
+- Bump github.com/go-logr/logr from 1.4.2 to 1.4.3 (#2575)
+- Bump CAPI to v1.10.2 (#2562)
+- Bump github.com/docker/docker from 28.1.1+incompatible to 28.2.0+incompatible in /test (#2576)
+- Updating main branch clusterctl tests (#2560)
+- Bump golang to v1.24 (#2569)
+- Apply BMH from CAPM3 during test and skip BMH creation from dev-env (#2505)
+- Fix ipam deploy in clusterctl tests (#2554)
+- Enable exhaustive linter (#2544)
+- Fix typo in bmo deployment overlay kustomization file name (#2556)
+- Bump CAPI to v1.10.1 (#2537)
```
Rozzii, line 91: Should we use superseded in alpha or not, I used it but I don't remember actually if we have a policy for that or not.

Rozzii: Okay we have discussed offline that according to existing code the block is not generated for Beta and RC so it should not be generated for Alpha either because it is even less impactful but the exclusion of the superseded block of the alpha release is/was missing from the repos.

We can revisit this topic later.

PR #2686, opened 2025-08-05
:seedling: [WIP] Parameterize K8s version in e2e tests

erjavaskivuori:
**What this PR does / why we need it**: Parameterize K8s version in e2e test to make it easier to test with different versions. Related to the new pipeline built for testing with pre-release Kubernetes versions (see  https://github.com/metal3-io/project-infra/pull/1060).


smoshiur1237 (commented):
/approve

test/e2e/config/e2e_conf.yaml
```diff
@@ -168,7 +168,7 @@ providers:
 variables:
   CNI: "/tmp/calico.yaml"
   KUBETEST_CONFIGURATION: "./data/kubetest/conformance.yaml"
-  KUBERNETES_VERSION: "v1.33.0"
+  KUBERNETES_VERSION: "${KUBERNETES_VERSION:-v1.33.0}"
```
tuminoid, line 171: I don't think this works. `envsubst` is not bash, it does simple string replace. It now tries to look for variable named `KUBERNETES_VERSION:-v1.33.0`  which it cannot find, so it does nothing.

```
$ export KUBERNETES_VERSION=yippee
$ envsubst < foo
kind: foo
values:
- name: bar
  value: ${KUBERNETES_VERSION:-default}
```

and even if it did, it would print out `value: foo:-default`  which is not what you want.

tuminoid (commented):
/lgtm

This should be fine as we're using envsubst-go to expand the variables, and not the normal envsubst.

PR #2694, opened 2025-08-11
🌱 Bump OSV scanner action to fix SARIF upload

lentzi90:

This version includes a fix for the issue with uploading the SARIF file (it is missing results). (https://github.com/google/osv-scanner/pull/2064 Fix SARIF v3 output to include results.)

This commit also makes the workflow run on PRs that touch the workflow file.

Fixes #

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -152,11 +152,11 @@ if [[ -n "${CONTAINER_RUNTIME}" ]]; then
         "${CONTAINER_RUNTIME}" run --rm
         -v "${PWD}":"/src:ro,z"
         -w /src
-        ghcr.io/google/osv-scanner:v2.1.0@sha256:9a1ba57d2a1506c9e9d0dfbeaf46346507e829745b70d47d77e12c38e66de8d7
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
/approve

Let's do the same fix for IPAM/BMO/IRSO as well where we have 2.1.0 bumped.

hack/verify-release.sh
```diff
@@ -218,7 +218,7 @@ check_tools()
         case "${tool}" in
             osv-scanner)
                 version=$("${OSVSCANNER_CMD[@]}" -v | grep version | cut -f3 -d" ")
-                min_version="2.1.0"
+                min_version="2.2.0"
```
lentzi90, line 221: Perhaps not strictly needed since the issue was only abut uploading the results, but I included it anyway

tuminoid: I was thinking the same, but as 2.1.0 has issues producing results, it is not bad idea to have 2.2.0 as minimum.

tuminoid:
GH action change.

PR #2696, opened 2025-08-12
:seedling: Uplift Ubuntu to 24.04

erjavaskivuori:
**What this PR does / why we need it**: Uplift Ubuntu to 24.04


tuminoid:
Hmm, it failed like Centos jobs failed at stream connection at fetch logs?

/retest

PR #2697, opened 2025-08-12
🌱 Fix osv-scanners Slack reporting

tuminoid:
osv-scanner scan returns 1 (error), if there are vulnerabilities. Script sections of GH actions run by default with set -e, which means has_vulnerabilities is never set, leading to Slack reporting not being activated.

### DIFF
--- a/.github/workflows/osv-scanner-scan.yml
+++ b/.github/workflows/osv-scanner-scan.yml
@@ -35,10 +35,12 @@ jobs:
     - name: Run OSV Scanner
       id: osv-scan
       run: |
+        # osv-scanner returns 1 if there is vulnerability, but script runs with set -e by default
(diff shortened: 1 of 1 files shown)


tuminoid:
/cc @lentzi90 @kashifest 
/hold

Slack notification is now triggered correctly, but no message is seen. Slack secret configuration to be verified on org-level for this repo.

lentzi90 (commented):
/lgtm

kashifest:
Test succeeded here https://github.com/metal3-io/cluster-api-provider-metal3/pull/2698, the issue was that Slack secret configuration only works at metal3 org level, for this PR the branch is from Nordix org and thus it wont have the access to the secret.

kashifest:

PR #2718, opened 2025-08-27
🌱 Add IPA_BASEURI in ironic config to fix IPA download from cache

smoshiur1237:
We have seen flaky behavior during e2e test where IPA downloader was having longer time to download image and which was causing failure to have available ironic deployment. This will fix the slowness in IPA image download process.

### DIFF
--- a/test/e2e/data/ironic-deployment/overlays/pr-test/ironic_bmo_configmap.env
+++ b/test/e2e/data/ironic-deployment/overlays/pr-test/ironic_bmo_configmap.env
@@ -10,3 +10,4 @@ DEPLOY_KERNEL_URL=http://172.22.0.2:6180/images/ironic-python-agent.kernel
 DEPLOY_RAMDISK_URL=http://172.22.0.2:6180/images/ironic-python-agent.initramfs
 IRONIC_ENDPOINT=http://172.22.0.2:6385/v1/
 USE_IRONIC_INSPECTOR=false
+IPA_BASEURI=https://artifactory.nordix.org/artifactory/openstack-remote-cache/ironic-python-agent/dib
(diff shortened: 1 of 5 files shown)


smoshiur1237:
/hold
Its open for your review. I need to do multiple test in centos job to confirm the fix

adilGhaffarDev:
It did not work, let's set it in the configmap in the ironic overlays. 
https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/data/ironic-deployment/overlays/pr-test/ironic_bmo_configmap.env 
Let's set it in all of them

smoshiur1237:
> It did not work, let's set it in the configmap in the ironic overlays. https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/test/e2e/data/ironic-deployment/overlays/pr-test/ironic_bmo_configmap.env Let's set it in all of them

IPA_BASEURI has the exported value in all over the test except the clean function. It seems like the variable is set initially from dev-env and later when we export in capm3 e2e, it is not taking the exported.

I am going to change it in config files, but I also think, if we fix it from dev-env, can also solve the issue

smoshiur1237:
After adding it to the config file, the fix is working at least in Ubuntu. Waiting for centos test to pass. We can take it in. 
https://jenkins.nordix.org/view/Metal3/job/metal3-ubuntu-e2e-integration-test-main/1582/
/unhold

dtantsur:
This fixes the tests but what about production deployments? Do we have any reason not to switch to this cache everywhere?

smoshiur1237:
> This fixes the tests but what about production deployments? Do we have any reason not to switch to this cache everywhere?

I will open an issue to switch it to cache and address it separately. For now I was focusing on our CI to work properly.

smoshiur1237:
Current fix will solve the CAPM3 e2e test and at this moment it is working quickly from nordix cache

Sunnatillo:
I am still not convinced that why we should add it here. If we need to change, we should change it in metal3-dev-env. Also, if we add in camp3 it adds up to number of variables used in camp3 e2e tests.

smoshiur1237:
Metal3-dev-env is not setting the IPA_BASEURI variable(it just get the default var which points to the opendev tarball), it depends on Project-infra for this variable in the test.. When running the capm3 e2e test it is getting the default value from dev-env, but we wants to download the IPA image from our Nordix cache. So It is not changing any function in CAPM3, but changing the IPA_BASEURI variable for CAPM3 e2e test.

PR #2737, opened 2025-09-02
:seedling: Enable testing with CAPI nightly builds

erjavaskivuori:
**What this PR does / why we need it**: Enable testing main branch against CAPI's nightly builds which are built from its main branch.

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -49,6 +49,18 @@ export FORCE_REPO_UPDATE="false"
 export USE_IRSO="${USE_IRSO:-false}"
 EOF
 
+# Always set DATE variable for nightly builds because it is needed to form
+# the URL for CAPI nightly build components in e2e_conf.yaml even if not used.
+DATE=$(date '+%Y%m%d' -d '1 day ago')
+export DATE
+
+# If CAPI_NIGHTLY_BUILD is true, it means that the tests are run against the
+# nightly build of CAPI components which are built from CAPI's main branch.
+if [[ "${CAPI_NIGHTLY_BUILD:-false}" == "true" ]]; then
+  export CAPIRELEASE="v1.12.99"
+  echo 'export CAPI_NIGHTLY_BUILD="true"' >>"${M3_DEV_ENV_PATH}/config_${USER}.sh"
+fi
+
 case "${GINKGO_FOCUS:-}" in
   clusterctl-upgrade|k8s-upgrade|basic|integration|remediation|k8s-conformance|capi-md-tests)
     # if running basic, integration, k8s upgrade, clusterctl-upgrade, remediation, k8s conformance or capi-md tests, skip apply bmhs in dev-env
--- a/docs/e2e-test.md
+++ b/docs/e2e-test.md
@@ -71,6 +71,17 @@ export GINKGO_SKIP=healthcheck remediation
 make test-e2e
 ```
 
+### Testing against CAPI nightly builds
+
+Cluster API publishes nightly versions of the project components’ manifests from
+the main branch. If you want to run tests against the nightly builds you can set:
+
+```sh
+export CAPI_NIGHTLY_BUILD=true
+```
+
+The used build is from the previous day.
+
 ## Cleanup
 
 After a finished test, cleanup is performed automatically. If the test setup
--- a/test/e2e/config/e2e_conf.yaml
+++ b/test/e2e/config/e2e_conf.yaml
@@ -17,6 +17,15 @@ providers:
 - name: cluster-api
   type: CoreProvider
   versions:
+  - name: "v1.12.99"
+    value: "https://storage.googleapis.com/k8s-staging-cluster-api/components/nightly_main_${DATE}/core-components.yaml"
+    type: "url"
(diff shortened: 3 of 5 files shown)


erjavaskivuori:
/hold

https://github.com/metal3-io/metal3-dev-env/pull/1565 has to be merged first.

scripts/ci-e2e.sh
```diff
@@ -52,11 +53,20 @@ export USE_IRSO="${USE_IRSO:-false}"
...
 
 case "${GINKGO_FOCUS:-}" in
-  clusterctl-upgrade|k8s-upgrade|basic|integration|remediation|k8s-conformance|capi-md-tests)
-    # if running basic, integration, k8s upgrade, clusterctl-upgrade, remediation, k8s conformance or capi-md tests, skip apply bmhs in dev-env
+  k8s-upgrade|basic|remediation|k8s-conformance|capi-md-tests)
+    # if running basic, k8s upgrade, remediation, k8s conformance or capi-md tests, skip apply bmhs in dev-env
     echo 'export SKIP_APPLY_BMH="true"' >>"${M3_DEV_ENV_PATH}/config_${USER}.sh"
   ;;
 
+  clusterctl-upgrade|integration)
```
Sunnatillo: I am still not sure if we need to run integration tests against CAPI  nightly builds. We run integration tests on PRs. IMHO it would be better to run CI related tests against CAPI stable branch/release. 
We can include any other test to run against CAPI nightly builds.

adilGhaffarDev: I agree, I think separate pipeline just for nightly build tests is better solution, but even with separate pipeline lets only run these 2 tests.

erjavaskivuori: Okay. I have thought about solution for this. I would propose adding `CAPI_NIGHTLY_BUILD` (boolean) variable on the [capm3-e2e-tests.pipeline](https://github.com/metal3-io/project-infra/blob/main/jenkins/jobs/capm3-e2e-tests.pipeline) environment variables. It would be defaulted to false. 

I would prefer using the same job template for the jobs running against CAPI nightly builds (to avoid copy-paste) so I would only add the variable on the existing job template. And then I would add new condition `capi_nightly_build` true/false for the job on the `cicd/jjb/metal3/projects.yml`. Something like this:

```
  - capm3_periodic_e2e_integration_tests_prow:
      capi_nightly_build:
      - True:
          capm3_target_branch:
          - main:
              ...
      - False:
          capm3_target_branch:
          - main:
              ...
          - release-1-10:
              ...
```

I think this way we can minimize needed changes and unnecessary copy-paste. Do we want this to be available from PRs or only periodical?

erjavaskivuori:
Now this only enables the usage but does not affect any tests.

erjavaskivuori:
> /hold This will affect clusterctl upgrade tests. They go from the specified start version to the latest it can find. Since we add v1.12.99 in the e2e config, that will be what is used unless we specify something specific.
> 
> It is controlled here:
> 
> https://github.com/metal3-io/cluster-api-provider-metal3/blob/75c2e50d45a97d292004032efc98b56aa0964a81/test/e2e/upgrade_clusterctl_test.go#L149-L152
> 
> 
> CAPI has an example for how to set specific versions in their upgrade test: https://github.com/kubernetes-sigs/cluster-api/blob/eda1321915ad07644b9ed8427e7657492e7f6882/test/e2e/clusterctl_upgrade_test.go#L89-L99
> Do we want to upgrade CAPI to latest nightly or should we keep it as it was and go to latest stable release? I don't have a strong opinion on this. We upgrade CAPM3 to latest main, so from that perspective it could be nice to also upgrade CAPI to nightly.

@lentzi90 Thank you for pointing this out, I was not aware of this. We have discussed elsewhere that if we run tests against CAPI nightly builds, we can have separate jobs for that and keep the ones we have unaffected. So I will look into how to set the specific versions.

lentzi90 (commented):
Now it looks correct to me!
/lgtm
Please keep the hold until we have verified that clusterctl-upgrade does the right thing

adilGhaffarDev:
LGTM 

In a separate PR, I suggest extending this to cover nightly builds for clusterctl tests too. Specifically, we could add another Describe block and attach an additional label, for example:
`Describe("When testing cluster upgrade from releases (v1.11 => capi-nightly-build)", Label("clusterctl-upgrade", "nightly-build"))`

The extra label ensures these tests are not triggered alongside the existing clusterctl tests. Instead, they would run under a dedicated trigger and daily job.

adilGhaffarDev:
> Now it looks correct to me! /lgtm Please keep the hold until we have verified that clusterctl-upgrade does the right thing

clusterctl seems to be working fine: 

`01:53:04.504    INFO: clusterctl upgrade apply --config /home/metal3ci/tested_repo/_artifacts/repository/clusterctl-config.yaml --kubeconfig /tmp/e2e-kubeconfig1111028545 --wait-providers --core cluster-api:v1.11.1 --bootstrap kubeadm:v1.11.1 --control-plane kubeadm:v1.11.1 --infrastructure metal3:v1.11.0 --ipam metal3:v1.11.0 `

PR #2752, opened 2025-09-04
🐛 Delete child resources before machine

adilGhaffarDev:
Remove postUpgradeWaitDeletingResources for testing

Cherry-pick of https://github.com/metal3-io/cluster-api-provider-metal3/pull/2717 
Fixes #

### DIFF
--- a/controllers/metal3machine_controller.go
+++ b/controllers/metal3machine_controller.go
@@ -292,17 +292,18 @@ func (r *Metal3MachineReconciler) reconcileDelete(ctx context.Context,
 
 	errType := capierrors.DeleteMachineError
 
+	// dissociate metadata if any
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
/approve

Rozzii:
FYI for future reference.

I would still say that the bug is severe enough that fixing it worth the trouble that we might cause for some folks who might be doing some automation that might have been adopted to this. But functionally it does not make any sense to not delete these child resources as new resources will be created anyhow.

I think this passes the standards that we have set, for backports it breaks our design and we do not provide a new feature we are just implementing one that should have been present all along and it does degrade the cluster over time so this is not just some optional feature that useful to have for a few users but it is a problem source.

PR #2755, opened 2025-09-04
✨ Retrieve gateway from CAPI IPAM IP claims

sathieu:

Gateway was missing from #769.

Fixes #


Rozzii:
/cc @peppi-lotta 
You are more familiar with this topic, could you please take a look?

peppi-lotta:
@sathieu Please document more clearly what you are doing and why. You are adding deprecation notices to quite many values. Please note that CAPM3 is not actually using CAPI's IPAddress and IPAddress resources even though they are supported. CAPM3 is mostly using Metal3-IPAM's IPAddress and IPClaim resources and these for example don't have field called poolRef. 

I'll run the test against these changes and see what happens. I'll review more thoroughly after the test have run. I don't see any harm in changing the naming but I think this needs a wider community discussion.

sathieu:
@peppi-lotta I've removed the deprecations and focus only on the feature I need: ability to retrieve gateway from IP claims. Please re-run the tests.

sathieu:
@peppi-lotta it looks like your /test command is not working. Maybe `/test metal3-ubuntu-e2e-feature-test-main-features`

peppi-lotta:
The test are running. They are e2e tests that can take a few hours to complete.

sathieu:
Hello @peppi-lotta @Rozzii v1.11.0 is now released. Can this PR be merged?

api/v1beta1/metal3datatemplate_types.go
```diff
@@ -360,6 +360,9 @@ type NetworkGatewayv4 struct {
 	// FromIPPool is the name of the IPPool to fetch the gateway from
 	// +optional
 	FromIPPool *string `json:"fromIPPool,omitempty"`
+
+	// FromPoolRef is a reference to a IP pool to fetch the gateway from
+	FromPoolRef *corev1.TypedLocalObjectReference `json:"fromPoolRef,omitempty"`
 }
```
adilGhaffarDev, line 366: Please explain what this is improving or fixing?

Right now user can set FromIPPool like below:
`fromIPPool: baremetalv4-pool`
And code take cares of converting it to IPPool of metal3 ipam.

But with FromPoolRef it will be like this:
```
fromPoolRef:
 - name: baremetalv4-pool
    kind: IPPool
    apiGroup: ipam.metal3.io
```
With FromPoolRef we are allowing user to change the kind and apiGroup, is this what we want?

sathieu: @adilGhaffarDev Yes. This is to support standard CAPI IPAM, for example [cluster-api-ipam-provider-in-cluster](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster).

For example:

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: Metal3DataTemplate
metadata:
  name: foo
spec:
  networkData:
    networks:
      ipv4:
      - id: "Baremetal"
        fromPoolRef:
          name: pool-1
          apiGroup: ipam.cluster.x-k8s.io
          kind: InClusterIPPool
        routes:
        - network: "0.0.0.0"
          netmask: 0
          gateway:
            fromPoolRef:  # << here and vvv here
              name: pool-1
              apiGroup: ipam.cluster.x-k8s.io
              kind: InClusterIPPool
```

sathieu: @adilGhaffarDev See also https://github.com/metal3-io/ip-address-manager/issues/971 for related topic.

adilGhaffarDev: Okay, but right now the code only expects either CAPI IPAM or Metal3 IPAM. We don’t have any validation hook in place to prevent users from specifying something outside of these (e.g., another kind or API group).

sathieu: @adilGhaffarDev The code expect any non-metal3 IPAM pool to be a CAPI IPAM pool. This is the correct behavior because such pool can have any GVK.

The behavior when using a wronk GVK would be the same as using a pool without installing the IPAM provider, or using a wrong name (i.e requeue).

adilGhaffarDev:
@sathieu what is this PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2810 doing? looks same to me.

api/v1beta1/metal3datatemplate_types.go
adilGhaffarDev: ok, Please update the description of PR in detail what its adding and why its adding it.

adilGhaffarDev:
> @sathieu what is this PR: #2810 doing? looks same to me.

If this PR is introducing the issue and PR: #2810 is solving it. Please add the fix here because it is not merged yet, dont open separate PR.

sathieu:
@adilGhaffarDev Correct (I pushed the wrong commit). I fixed #2810 to be as said in its title.

api/v1beta1/metal3datatemplate_types.go
sathieu: @adilGhaffarDev PR description updated

sathieu:
> If this PR is introducing the issue and PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2810 is solving it. Please add the fix here because it is not merged yet, dont open separate PR.

#2810 is fixing an issue for retrieving ip address from a CAPI IPAM Pool, while #2755 is about adding gateway support from CAPI IPAM pool. Sorry for pushing wrong commit to #2810 (now corrected).

adilGhaffarDev (approved):
/approve
@peppi-lotta can give final lgtm.

peppi-lotta:
@sathieu Have you tested these changes manually somehow? Can you outline steps you have taken to test?

sathieu:
@peppi-lotta I have tested from current main (75c2e50d45) with PRs #2755 #2763 and #2810 applied. I also upgraded CAPI to v1.11.1 (I was in 1.10.x). I manually updated the CRDs

I can confirm that:

- ✅ the Pool not in cache problem is fixed (#2810 )
- ✅ gateway can now come from IPAM (#2755)
- ✅  controller starts without IPClaim CRD (#2763)

kashifest:
lgtm, thanks for adding this.

sathieu:
@Rozzii Would this be in 1.11.x?

PR #2763, opened 2025-09-09
✨ Conditionally watch M3 IPClaims and CAPI IPAddressClaims

### LINKED ISSUE #2654: CAPM3 1.10.1 fails to start when
`clusterctl init --core cluster-api:v1.9.6 --bootstrap - --control-plane - --infrastructure metal3:v1.10.1`

CAPM3 Manager Pod keeps complaining until it fails:

```
main.go:218] "problem running manager" err="failed to wait for metal3data caches to sync kind source: *v1alpha1.IPClaim: timed out waiting for cache to be synced for Kind *v1alpha1.IPClaim" logger="setup"
```

To boot normally, as v1.9.4

When running v1.10.1, no IPAM crds will get created
```
kubectl get crd | grep -i ipam.metal3
```

However when running v1.9.4, they will.

```
kubectl get crd | grep -i ipam.metal3
ipaddresses.ipam.metal3.io                                   2025-07-23T16:19:39Z
ipclaims.ipam.metal3.io                                      2025-07-23T16:19:39Z
ippools.ipam.metal3.io                                       2025-07-23T16:19:39Z
```

I understand this might come from the decoupling of the IPAM component, however shouldn't this probably be included in the advertised manifest, or have it as optional component? In our case we do not use IPAM, but either way would work.

As a workaround at the moment we are forced to deploy IPAM CRDs

**Environment:**

- Cluster-api version: 1.9.6
- Cluster-api-provider-metal3 version: 1.10.1
- Environment (metal3-dev-env or other): kind
- Kubernetes version: (use `kubectl version`): v1.31.2

/kind bug

sathieu:

- only watch for IPClaims if the CRD is available (use direct API calls as cache is not ready yet)
- also watch for CAPI IPAM IPAddressClaims when CRD is available (direct)
- also fixes an AddToScheme typo

Fixes #2654.

Fixes #2654.

### DIFF
--- a/controllers/metal3data_controller.go
+++ b/controllers/metal3data_controller.go
@@ -25,9 +25,11 @@ import (
 	ipamv1 "github.com/metal3-io/ip-address-manager/api/v1alpha1"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	capipamv1beta1 "sigs.k8s.io/cluster-api/api/ipam/v1beta1"
 	"sigs.k8s.io/cluster-api/controllers/clustercache"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/annotations"
@@ -46,6 +48,7 @@ const (
 // Metal3DataReconciler reconciles a Metal3Data object.
 type Metal3DataReconciler struct {
 	Client           client.Client
+	ClientReader     client.Reader
 	ClusterCache     clustercache.ClusterCache
 	ManagerFactory   baremetal.ManagerFactoryInterface
 	Log              logr.Logger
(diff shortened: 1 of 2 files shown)


sathieu:
I've improved the code to automatically disable the watch. Ready for review...

controllers/metal3data_controller.go
```diff
@@ -168,13 +168,25 @@ func (r *Metal3DataReconciler) reconcileDelete(ctx context.Context,
...
-		Watches(
-			&ipamv1.IPClaim{},
-			handler.EnqueueRequestsFromMapFunc(r.Metal3IPClaimToMetal3Data),
-		).
+		WithOptions(options)
+	// Test for IPClaim existence
+	ipClaimList := &ipamv1.IPClaimList{}
+	err := r.Client.List(ctx, ipClaimList, client.Limit(1))
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
```
adilGhaffarDev:
```suggestion
		if meta.IsNoMatchError(err) {
```
I would suggest using `IsNoMatchError` from `k8s.io/apimachinery/pkg/api/meta` to be more precise.

sathieu: Good idea. Done. back to you @adilGhaffarDev

Rozzii:
Cherry Picking to 1.11 and 1.10 should be fine too.

peppi-lotta:
/hold
I want to understand why we need to watch IPClaims in CAPM3. And if we are not watching the Metal3-IPAMProvider IPCLaims should we watching the for CAPI's IPAddressClaim types? 

I agree that this fixes the errors related to deploying with out Metal3-IPAM but I'm not sure if this is to full solution yet.

sathieu:
@peppi-lotta Yes, ipclaims are retrieved in metal3data manager:

https://github.com/metal3-io/cluster-api-provider-metal3/blob/5d5f20894dd7fdbe54fb3317cdf744dd099cc9f0/baremetal/metal3data_manager.go#L243
https://github.com/metal3-io/cluster-api-provider-metal3/blob/5d5f20894dd7fdbe54fb3317cdf744dd099cc9f0/baremetal/metal3data_manager.go#L277

 Network data needs to be updated when an ip claim is created or changed.

Maybe a conditional watch is also needed for CAPI IPClaims? Should I add it in this PR?

sathieu:
I have improved (and renamed) this PR. Review welcome :pray:

tuminoid:
/retitle ✨ Conditionally watch M3 IPClaims and CAPI IPAddressClaims

A typofix

peppi-lotta:
I think there should be a check that atleast one of the supported ippool CRDs is present. If no ip pool crd is present these should be some warning and pod creations should fail. With these changes all capm3 pods become ready with no complaint but missing IPAM will later cause issues because CAPM3 is dependent on an IPAM. I think the `capm3-controller-manager` should not become running if no valid IPAM is deployed.

sathieu:
@peppi-lotta I'm not sure cap-m3 depends on an IPAM. One can use [Pre-Provisioning NetworkData](https://book.metal3.io/bmo/advanced_instance_customization#pre-provisioning-networkdata), and not set networkData in Metal3DataTemplate (this will use Pre-Provisioning NetworkData, [ref](https://github.com/metal3-io/baremetal-operator/blob/6f75021d4384e1f4b7daaa6e543ff2ca0a3749d5/internal/controller/metal3.io/host_config_data.go#L73)).

sathieu:
@peppi-lotta Have you read [my above comment](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2763#issuecomment-3355046893)?

I can change the code depending on your response.

peppi-lotta:
I think you’re right, setting IPAddressesFromPool in the Metal3DataTemplate is optional. So, if no pools are defined in the template, IPAM shouldn’t be required. 
/lgtm

sathieu:
/assign @Sunnatillo 

please approve

(Sorry if this is not the correct command,

controllers/metal3data_controller.go
```diff
@@ -168,23 +171,81 @@ func (r *Metal3DataReconciler) reconcileDelete(ctx context.Context,
...
+
+	// Test for CAPI IPAddressClaim existence
+	ipAddressClaimList := &capipamv1beta1.IPAddressClaimList{}
+	err := r.ClientReader.List(ctx, ipAddressClaimList, client.Limit(1))
+	if err != nil {
+		if !meta.IsNoMatchError(err) {
+			return err
+		}
+		ctrl.LoggerFrom(ctx).Info("IPAddressClaim CRD not found, skipping watch")
+	} else {
```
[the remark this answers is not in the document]
adilGhaffarDev: This is fine as it is.

PR #2764, opened 2025-09-10
WIP: ✨ Add FailureDomain Support

Sunnatillo:

Fixes #


controllers/metal3machine_controller.go
```diff
@@ -125,6 +125,12 @@ func (r *Metal3MachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	machineLog = machineLog.WithValues("machine", capiMachine.Name)
 
+	// Set Failuredomain from machine to metal3Machine
+	if capm3Machine.Spec.FailureDomain != capiMachine.Spec.FailureDomain {
+		capm3Machine.Spec.FailureDomain = capiMachine.Spec.FailureDomain
+		return ctrl.Result{}, nil
```
lentzi90: I don't think we should return here. Just continue and it will do the patch in the end.

```suggestion
```

api/v1beta1/metal3cluster_types.go
```diff
@@ -66,6 +66,25 @@ type Metal3ClusterSpec struct {
...
+	// +optional
+	FailureDomains FailureDomains `json:"failureDomains,omitempty"`
+}
+
+// FailureDomains is a slice of FailureDomainSpecs.
+type FailureDomains map[string]FailureDomainSpec
+
+// FailureDomainSpec is the Schema for Cluster API failure domains.
+// It allows controllers to understand how many failure domains a cluster can optionally span across.
+type FailureDomainSpec struct {
```
lentzi90: There is a struct for this already upstream. Can we reuse it instead of creating our own?
https://github.com/kubernetes-sigs/cluster-api/blob/4eb07516c7c3db623d798d08f7407eb44fc0c08d/api/core/v1beta2/cluster_types.go#L1614

lentzi90 (commented):
Looks quite good! This will be a nice improvement.
A couple of comments below.

api/v1beta1/metal3cluster_types.go
Sunnatillo: Initially I tried using that, but then realized cluster-api looks for type FailureDomains map[string]FailureDomainSpec type from InfraCluster. CAPI wants all FD should be inside one map.
Other CAPI providers implemented the same way.

controllers/metal3machine_controller.go
Sunnatillo: done

api/v1beta1/metal3cluster_types.go
lentzi90: Sorry I don't understand. Can you show how others are doing it?
I was just thinking we could do something like this:

```go
type FailureDomains map[string]clusterv1.FailureDomain
```

adilGhaffarDev: I agree with @lentzi90 we shouldn’t define our own struct. I checked the vSphere implementation, and they’re using the CAPI structs instead of defining their own.
ref: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/430d770b780e442b523c6a8d1747d3036614df32/apis/v1beta1/vspherecluster_types.go#L172

adilGhaffarDev:
Are there any tests we can add? Or maybe we could update an existing e2e test YAMLs just to verify that this works as expected.

adilGhaffarDev:
Please open a documentation PR as well. We don’t need to explain how FailureDomains work,  just mention that they’re now supported in CAPM3 and refer to the CAPI documentation for details.

api/v1beta1/metal3cluster_types.go
Sunnatillo: I have tested with failureDomain struct of CAPI. In CAPI  v1beta2 failureDomains type is map and in v1beta2 it is an array. 
When CAPI copies metal3Cluster status, it checks contract version of metal3Cluster and according to that it will expect the type. We are still in v1beta1 so it expects map. Here is an error when I tried to use CAPI failureDomains struct:

`pod/capi-controller-manager-548bcf65b-cvfg7/manager] E1021 13:01:34.903106 1 controller.go:353] "Reconciler error" err="failed to retrieve status.failureDomains from infrastructure provider for Cluster \"test1\" in namespace \"metal3\": failed to get .status.failureDomains from object: .status.failureDomains accessor error: [map[controlPlane:true name:rack-1] map[controlPlane:false name:rack-2]] is of the type []interface {}, expected map[string]interface{}" controller="cluster" controllerGroup="cluster.x-k8s.io" controllerKind="Cluster" Cluster="metal3/test1" namespace="metal3" name="test1" reconcileID="3d531abd-fc68-4d97-80b3-d3b9bf12834b"`

When we create v1beta2, then we should create failureDomains type with CAPI struct

Sunnatillo:
> Please open a documentation PR as well. We don’t need to explain how FailureDomains work, just mention that they’re now supported in CAPM3 and refer to the CAPI documentation for details.

Documentation PR added. https://github.com/metal3-io/metal3-docs/pull/588

api/v1beta1/metal3cluster_types.go
Sunnatillo: Here CAPI retrives failureDomains from infraCluster:
https://github.com/sbueringer/cluster-api/blob/3e9e41670af093243ce1b6ab8c1cd6be26ca5d5d/internal/controllers/cluster/cluster_controller_phases.go#L235

Sunnatillo:
> Are there any tests we can add? Or maybe we could update an existing e2e test YAMLs just to verify that this works as expected.

Unit tests are added

Sunnatillo:
@adilGhaffarDev @lentzi90.

I have added documentation PR, unit tests and logic when no hosts available in Failure Domain it will choose any other available host. 
Also checked if this works correctly in manual deployment. Everything works fine. PTAL again. Thanks

api/v1beta1/metal3cluster_types.go
lentzi90: I think you are misunderstanding what we are saying. The map/array can stay, just the content (this struct) can be reused.

Up on line 76, put this instead:

```go
type FailureDomains map[string]clusterv1.FailureDomains
```

It is still `map[string]`, but now mapping to the upstream struct.

Sunnatillo: I have tried.When adding failureDomain field to Metal3Cluster object it accepts it, but it does not show up in the specs. Failure doimain is not showing up after adding it to metal3Cluster object. I have tried by editing, by changing yaml and applying, Did not help. 
 I could not figure out why. Maybe we use v1beta2 CAPI contract. I do not know how this is relevant.

adilGhaffarDev: I don't understand why we need v1beta2? 
CAPI FailureDomains struct is v1beta1. Check the following code
```
	FailureDomains clusterv1beta1.FailureDomains `json:"failureDomains,omitempty"`
```
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/430d770b780e442b523c6a8d1747d3036614df32/apis/v1beta1/vspherecluster_types.go#L172

Sunnatillo: I make it work with v1beta1 struct of capi. I guess it was the issue with my environment where it did not show the FailureDomains after accepting it. 

This is tested and working.

baremetal/metal3machine_manager_test.go
```diff
@@ -849,6 +879,43 @@ var _ = Describe("Metal3Machine manager", func() {
...
+				M3Machine:        m3mconfig6,
+				ExpectedHostName: hostWithFailureDomainLabel.Name,
+			}),
+			Entry("Choose available host, when hosts in FailureDomain not available", testCaseChooseHost{
+				Machine:          newMachine(machineName, infrastructureRef6),
+				Hosts:            &bmov1alpha1.BareMetalHostList{Items: []bmov1alpha1.BareMetalHost{*availableHost, hostWithOtherConsRef, hostWithNodeReuseLabelSetToCP}},
+				M3Machine:        m3mconfig6,
+				ExpectedHostName: availableHost.Name,
+			}),
+			Entry("Choose a host in Failure Domain, when NodeReuse is set", testCaseChooseHost{
```
lentzi90, line 894: What about a test for NodeReuse and host NOT in Failure Domain? It is an important corner case. We should prefer NodeReuse, and I think the logic is correct already, but we should test it also.

controllers/metal3cluster_controller.go
```diff
@@ -111,6 +112,12 @@ func (r *Metal3ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}()
 
+	// Set FailureDomains to status if it is not already set
+	if !equality.Semantic.DeepEqual(metal3Cluster.Spec.FailureDomains, metal3Cluster.Status.FailureDomains) {
+		metal3Cluster.Status.FailureDomains = metal3Cluster.Spec.FailureDomains
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil
```
lentzi90: Sorry I missed this before. I think we should not return here. We can continue and avoid an extra requeue.

Sunnatillo: removed now

baremetal/metal3machine_manager_test.go
Sunnatillo: Thanks. I added test case for this as well.

PR #2800, opened 2025-09-15
:sparkles: Bump IPAM to v1.11.0

Rozzii:
This commit:
 - Updates IPAM version to v1.11.0


kashifest:
why ✨ ? shouldn't it be 🌱 ?

adilGhaffarDev:
> why ✨ ? shouldn't it be 🌱 ?

yes should be seedling

PR #2804, opened 2025-09-17
🌱 Update clusterctl upgrade tests on main branch

adilGhaffarDev:
**What this PR does / why we need it**:
 Update clusterctl upgrade tests on main branch

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2805, opened 2025-09-17
🌱 Update infra metadata in tests

adilGhaffarDev:
Update infra metadata in tests
follow-up of https://github.com/metal3-io/cluster-api-provider-metal3/pull/2804

Fixes #


adilGhaffarDev:
cc @Rozzii @kashifest

test/e2e/config/e2e_conf.yaml
```diff
@@ -35,15 +35,6 @@ providers:
       new: --metrics-addr=:8080
     files:
     - sourcePath: "../data/shared/capi/v1.10/metadata.yaml"
-  - name: "{go://sigs.k8s.io/cluster-api@v1.9}"
```
Rozzii, line 38: AFAIK we 1.10 and 1.11 are supported but 1.9 stay tested.
I don't think we need to delete it.

test/e2e/config/e2e_conf.yaml
```diff
@@ -241,13 +203,10 @@ variables:
   IMAGE_CHECKSUM_TYPE: "sha256"
   IMAGE_USERNAME: "metal3"
   NODE_DRAIN_TIMEOUT: 0
-  IRONIC_RELEASE_26.0: "data/ironic-deployment/overlays/release-26.0"
-  IRONIC_RELEASE_27.0: "data/ironic-deployment/overlays/release-27.0"
```
Rozzii, line 248: AFAIK 1.9 should stay and ironic-image 27 too

test/e2e/config/e2e_conf.yaml
adilGhaffarDev: this is not used anywhere in the code, for 1.9 we have separate pipeline.
When it comes to upgrade, we are testing 1.9->1.11 and 1.10->1.11 upgrade in release-1.11 branch.
On main branch we will test 1.10->main and 1.11->main , this is what this pr is updating.

test/e2e/config/e2e_conf.yaml
adilGhaffarDev: same goes here. Not used in any test on main branch.

test/e2e/config/e2e_conf.yaml
Rozzii: I see, then let's leave it because we might need to actually use it in the future. I think stakeholders will do 1.9 -> 1.10 -> 1.11 so we have to discuss what we are missing .

adilGhaffarDev:
cc @kashifest 

test/e2e/upgrade_clusterctl_test.go
```diff
@@ -55,8 +55,8 @@ var _ = Describe("When testing cluster upgrade from releases (v1.11=>current)",
 	ironicFromRelease := "31.0"
 	bmoToRelease := "latest"
 	ironicToRelease := "latest"
-	capiStableRelease, err := capi_e2e.GetStableReleaseOfMinor(ctx, minorVersion)
-	Expect(err).ToNot(HaveOccurred(), "Failed to get stable version for CAPI minor release : %s", minorVersion)
+	// This will upgrade from v1.11.0 to latest patch release of v1.11.x
+	capiStableRelease := "1.11.0"
```
kashifest, line 59: so whats happening here, why are we hardcoding it ?

lentzi90: There was some change in the test upstream so that it fails if there is no upgrade to do. I.e. if we take the latest v1.11.x, there is no upgrade for CAPI and the test fails.
If we instead hard code to v1.11.0, then we will be able to upgrade to v1.11.1 and it passes :slightly_smiling_face:

lentzi90: Context: https://kubernetes.slack.com/archives/C8TSNPY4T/p1757671489165469
Related PR: https://github.com/kubernetes-sigs/cluster-api/pull/12753

PR #2819, opened 2025-09-23
🌱 Change interface names for predicatable ones for Centos

Sunnatillo:
We are changing interface names to predictable ones, because we enable predictable interface names for our node images here:
https://github.com/metal3-io/project-infra/pull/1053

This PR also uplifts kubernetes versions which makes easier to swap new node images with predictable interface names enabled.


tuminoid (commented):
Does the project-infra change break 1.9 without this? If not, 1.9 is tested-only and I'd prefer leaving this out unless necessary.

Sunnatillo:
> Does the project-infra change break 1.9 without this? If not, 1.9 is tested-only and I'd prefer leaving this out unless necessary.

Valid point.

Sunnatillo:
we do not need it. release 1.9 is only tested in CI. Branch is out of support
/close

PR #2820, opened 2025-09-24
:seedling: Add logging to releaseAddressFromM3Pool function

peppi-lotta:
**What this PR does / why we need it**: Add logging to releaseAddressFromM3Pool function to help debugging


Rozzii:
Cherry picking the log improvement  help debugging a suspected functional bug in 1.10.x and 1.11.x as well as in main.

PR #2829, opened 2025-09-29
:seedling: add specific time to dependabot config

tuminoid:
Previously Dependabot executed the updates for all branches at the same time (00:00 UTC). If there were enough of the PRs, the Prow would choke, despite it has scaling on as infra provider was sometimes slow to create and provision another Prow worker.

By pacing the updated 0.5hr apart per branch (we already have repos on separate weekdays), we avoid pod scheduling failures and make merging the updates smoother process.

Also, fix the job so all CAPM3 bumps are on Tuesday.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
   schedule:
     interval: "monthly"
     day: "tuesday"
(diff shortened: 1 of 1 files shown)


tuminoid:
GH action change.

PR #2832, opened 2025-09-30
🌱 Use build cache mounts and tidy up the dockerfile

fmuyassarov:
1. Use build cache mounts so that dependencies are cached between Docker builds, reducing the need to redownload them every time the image is built.
2. Tidy up the Dockerfile a bit (remove the unused `ARCH` argument and etc)

Fixes #
/hold

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ COPY api/go.mod api/go.sum api/
 COPY test/go.mod test/go.sum test/
 # Cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+RUN --mount=type=cache,target=/go/pkg/mod/ go mod download
 
 # Copy the sources
 COPY main.go main.go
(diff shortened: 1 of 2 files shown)


fmuyassarov:
@tuminoid can you please confirm that there is no CI or anything like that depends on the ARCH value? At, least if I look at the quay.io, I see that images are published without the architecture suffix.

Dockerfile
```diff
@@ -41,25 +41,23 @@ COPY controllers/ controllers/
...
-ARG ARCH
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} \
+RUN CGO_ENABLED=0 GOOS=linux \
     go build -a -ldflags "${LDFLAGS}" \
     -o manager .
 
 # Copy the controller-manager into a thin image
 FROM $BASE_IMAGE
+# Use uid of nonroot user (65532) because kubernetes expects numeric user when applying pod security policies
+USER 65532:65532
```
tuminoid, line 51: Make sure adding this here doesn't break Tilt.

tuminoid:
> @tuminoid can you please confirm that there is no CI or anything like that depends on the ARCH value? At, least if I look at the quay.io, I see that images are published without the architecture suffix.

AFAIK we don't use the makefile based docker build anywhere, especially in a form it has ARCH in it. 

We can also clean up the makefile and remove docker-push and docker-push-all targets, we're not manually pushing anything for long time.

Dockerfile
fmuyassarov: I wish I can test it, but unfortunately currently Tilt is broken already.

tuminoid: Oh, any idea what broke it? It was working "some" time ago

PR #2830, opened 2025-09-29
🌱 cmd: strip out symbol table & DWARF debugging info

fmuyassarov:
**What this PR does / why we need it**:
DWARF is mostly needed when we are using Delve. DWARF was originally created for C language and the need for it in go and other debugging features is very small in general. As such, switch to using `-ldflags="-s -w"` by default and have an option to opt in for debugging info for those who would really need it. As a real example, we can gain ~30% (i.e. 26MB) image size reduction for capm3 controller and faster build process.

**Before**
Size profiler comparison: 

```
$ bloaty -d sections manager
    FILE SIZE        VM SIZE
 --------------  --------------
  32.1%  25.6Mi  46.4%  25.6Mi    .text
  22.6%  18.0Mi  32.7%  18.0Mi    .gopclntab
  12.2%  9.70Mi  17.6%  9.70Mi    .rodata
   7.9%  6.28Mi   0.0%      64    .debug_info
   7.4%  5.90Mi   0.0%      64    .strtab
   5.7%  4.56Mi   0.0%      64    .debug_loc
   4.4%  3.47Mi   0.0%      64    .debug_line
   2.8%  2.26Mi   0.0%      64    .symtab
   1.8%  1.45Mi   0.0%      64    .debug_ranges
   1.3%  1.06Mi   1.9%  1.06Mi    .noptrdata
   1.1%   903Ki   0.0%      64    .debug_frame
   0.4%   340Ki   0.6%   340Ki    .data
   0.0%       0   0.4%   221Ki    .bss
   0.1%  80.9Ki   0.1%  80.9Ki    .typelink
   0.0%       0   0.1%  61.3Ki    .noptrbss
   0.0%  35.5Ki   0.1%  35.5Ki    .itablink
   0.0%  8.70Ki   0.0%  8.70Ki    .go.buildinfo
   0.0%  7.41Ki   0.0%       0    [Unmapped]
   0.0%  2.55Ki   0.0%  2.55Ki    [LOAD #2 [RX]]
   0.0%  1.02Ki   0.0%     720    [7 Others]
   0.0%     407   0.0%      64    .debug_abbrev
 100.0%  79.7Mi 100.0%  55.1Mi    TOTAL
```

**After**
```
$ bloaty -d sections manager
    FILE SIZE        VM SIZE
 --------------  --------------
  46.7%  25.6Mi  46.4%  25.6Mi    .text
  32.9%  18.0Mi  32.7%  18.0Mi    .gopclntab
[...]

PR #2833, opened 2025-09-30
🌱 Dockerfile: strip out symbol table by default for public images

fmuyassarov:
This is a follow up of https://github.com/metal3-io/cluster-api-provider-metal3/pull/2830, so that our public images in quay.io also benefit from this change. Since GitHub workflow in project-infra doesn't run the Makefile target but instead uses the Dockerfile directly, fix the defaulting of `-s -w` within the Dockerfile here. 
 
Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -549,7 +549,6 @@ generate-examples-clusterclass: $(KUSTOMIZE) clean-examples ## Generate examples
 .PHONY: docker-build
 docker-build: ## Build the docker image for controller-manager
 	docker build --network=host --pull \
-	--build-arg LDFLAGS="-s -w -extldflags=-static" \
 	--build-arg ARCH=$(ARCH) . -t $(CONTROLLER_IMG)-$(ARCH):$(TAG)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -549,7 +549,6 @@ generate-examples-clusterclass: $(KUSTOMIZE) clean-examples ## Generate examples
 .PHONY: docker-build
 docker-build: ## Build the docker image for controller-manager
 	docker build --network=host --pull \
-	--build-arg LDFLAGS="-s -w -extldflags=-static" \
```
kashifest, line 552: whats the harm if we still keep it here?

fmuyassarov: there no harm nor a benefit since it is equal to the default value https://github.com/Nordix/cluster-api-provider-metal3/blob/8a000d2d027bdbec6b065131acd98e4fd9c257c5/Dockerfile#L26

kashifest: Ah right, so in that sense a duplicate

tuminoid (commented):
/lgtm

PR #2826, opened 2025-09-26
:seedling: Add prep fkas directory -step for fkas-image-build action 

peppi-lotta:
**What this PR does / why we need it**: Prep fkas directory for build in github action
I've been able to run this the added steps successfully locally and build steps should work as always. I don't know if there is any way to run this workflow before merging. These steps are needed because in this pr, https://github.com/metal3-io/cluster-api-provider-metal3/pull/2814, I switch the fkas builds are using from static to main.

PR #2846, opened 2025-10-08
:seedling: Pass fkas artifact to reusable image building workflow

peppi-lotta:
**What this PR does / why we need it**: A docker image building requires that all the element needed for building are in the same directory or sub directory where the Dockerfile is located. This is why I created a [PR ](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2826)that pres FKAS for image building by moving the required CAPM3 directory into the FKAS direcory before building. Because we are using a reusable workflow for image building the directory prep has to be done in a separate job. It seems changes are not retained accross jobs even when they are run in the same workflow so I'm suggesting to use the upload-archive and download-archive actions to pass the pre prepaped FKAS folder from one job to another. Related PR in project-infra can be found here: 

Fixes #


tuminoid:

PR #2849, opened 2025-10-09
🌱 Switch to modern constructs via modernizer

fmuyassarov:
Switch to modern constructs via modernize analyzer.

Fixes #

### DIFF
--- a/baremetal/utils.go
+++ b/baremetal/utils.go
@@ -19,6 +19,7 @@ package baremetal
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
@@ -56,12 +57,7 @@ var (
 
 // Contains returns true if a list contains a string.
 func Contains(list []string, strToSearch string) bool {
-	for _, item := range list {
-		if item == strToSearch {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(list, strToSearch)
 }
 
 // NotFoundError represents that an object was not found.
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"path/filepath"
+	"slices"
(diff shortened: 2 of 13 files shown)


fmuyassarov:
> We have some more `interface{}` in `test/e2e` and `hack/fake-apiserver`, can you change those as well?

Yes, done.

tuminoid:
> > We have some more `interface{}` in `test/e2e` and `hack/fake-apiserver`, can you change those as well?
> 
> Yes, done.

That fixed only FKAS, `test/e2e` has 3 files with matches still.

fmuyassarov:
> > > We have some more `interface{}` in `test/e2e` and `hack/fake-apiserver`, can you change those as well?
> > 
> > 
> > Yes, done.
> 
> That fixed only FKAS, `test/e2e` has 3 files with matches still.

Should be fixed now. 

baremetal/utils.go
```diff
@@ -56,12 +57,7 @@ var (
 
 // Contains returns true if a list contains a string.
 func Contains(list []string, strToSearch string) bool {
-	for _, item := range list {
-		if item == strToSearch {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(list, strToSearch)
```
lentzi90, line 60: Why not replace any call to `Contains` with `slices.Contains`?

controllers/suite_test.go
```diff
@@ -187,12 +188,7 @@ func bmcOwnerRef() *metav1.OwnerReference {
 }
 
 func contains(haystack []string, needle string) bool {
-	for _, straw := range haystack {
-		if straw == needle {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(haystack, needle)
```
lentzi90, line 191: Here it would also be nice to get rid of `contains` and just call `slices.Contains` directly.

PR #2850, opened 2025-10-09
🌱manager: clean up the manager manifest

fmuyassarov:
Clean up the manager manifest

Fixes #

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -29,11 +29,6 @@ spec:
         image: controller:latest
         imagePullPolicy: IfNotPresent
         name: manager
-        env:
-        - name: POD_NAMESPACE
-          valueFrom:
-            fieldRef:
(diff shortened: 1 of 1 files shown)


fmuyassarov:
> /lgtm
> 
> Only BMO uses POD_NAMESPACE for something. Similar cleanup can be done in IPAM.

it is already there https://github.com/metal3-io/ip-address-manager/pull/1170

fmuyassarov:
> /approve Should we also get rid of the configmap for fast-track? It is a bit silly to create a configmap just for one single environment variable that we could set directly with `env` instead of `envFrom`.
> 
> https://github.com/metal3-io/cluster-api-provider-metal3/blob/4db81420857b81deb1d7ab5aae11cf011fa94cd8/config/manager/kustomization.yaml#L7-L13

@lentzi90, +1

PR #2852, opened 2025-10-09
🌱 Remove k8s repo from kubeadm config

Sunnatillo:
This PR updates removes k8s repo from kubeadm config for centos

Fixes #

### DIFF
--- a/test/e2e/data/infrastructure-metal3/main/bases/centos-kubeadm-config/centos-kubeadm-config.yaml
+++ b/test/e2e/data/infrastructure-metal3/main/bases/centos-kubeadm-config/centos-kubeadm-config.yaml
@@ -92,17 +92,6 @@ spec:
       owner: root:root
       permissions: '0600'
     - content: |
-        [kubernetes]
-        name=Kubernetes
-        baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
-        enabled=1
-        gpgcheck=1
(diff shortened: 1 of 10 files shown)


Sunnatillo:
> What do we use these repos for? I don't see any installation of packages in the preKubeadmCommands that would make use of them. Can we really have a hard coded minor version of k8s or do we need to template the value?

I do not know what we use this for. But it really annoying live debugging which breaks yum install. Hard coding is happening because of the new format of repo link.

lentzi90:
Same comment as in https://github.com/metal3-io/metal3-dev-env/pull/1579
It sounds like we do not need these. They are already broken but the tests still pass, so I say we should remove them.
If they are needed (for debugging for example) then we should set the k8s version from a variable. I know these repos are specific per minor version, but it makes zero sense to have hard-coded v1.34 on the node if we are running a v1.32 test.

Sunnatillo:
> Same comment as in [metal3-io/metal3-dev-env#1579](https://github.com/metal3-io/metal3-dev-env/pull/1579) It sounds like we do not need these. They are already broken but the tests still pass, so I say we should remove them. If they are needed (for debugging for example) then we should set the k8s version from a variable. I know these repos are specific per minor version, but it makes zero sense to have hard-coded v1.34 on the node if we are running a v1.32 test.

I agree, I am removing them now.

Sunnatillo:

Ci issue, retriggering

PR #2853, opened 2025-10-09
🌱 Add SKIP_NODE_IMAGE_PREPULL var in ci-e2e.sh 

kashifest:
This var informs dev-env not to prepull node image for e2e tests from dev-env. e2e test should take care of correct node image download from within the test

### DIFF
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -39,6 +39,8 @@ M3_DEV_ENV_PATH="${M3_DEV_ENV_PATH:-${WORKING_DIR}/metal3-dev-env}"
 clone_repo "${M3_DEV_ENV_REPO}" "${M3_DEV_ENV_BRANCH}" "${M3_DEV_ENV_PATH}"
 
 # Config devenv
+# SKIP_NODE_IMAGE_PREPULL is set to true to avoid dev-env downloading the node 
+# image
 cat <<-EOF >"${M3_DEV_ENV_PATH}/config_${USER}.sh"
 export CAPI_VERSION="v1beta2"
 export CAPM3_VERSION=${CAPM3_VERSION:-"v1beta1"}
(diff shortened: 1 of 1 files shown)


kashifest:
/hold
It has a test commit

scripts/ci-e2e.sh
```diff
@@ -65,19 +68,22 @@ case "${GINKGO_FOCUS:-}" in
   clusterctl-upgrade|k8s-upgrade|basic|integration|remediation|k8s-conformance|capi-md-tests)
     # if running basic, integration, k8s upgrade, clusterctl-upgrade, remediation, k8s conformance or capi-md tests, skip apply bmhs in dev-env
     echo 'export SKIP_APPLY_BMH="true"' >>"${M3_DEV_ENV_PATH}/config_${USER}.sh"
+    echo 'export SKIP_NODE_IMAGE_PREPULL="true"' >>"${M3_DEV_ENV_PATH}/config_${USER}.sh"
```
kashifest, line 71: I am a bit unsure why we are doing like this, but I followed `SKIP_APPLY_BMH` here. @adilGhaffarDev any idea? Shouldn't exporting it to true at the top should be enough?

tuminoid: Values in config_xxx.sh are persistent across various make targets etc as the config is sourced consistently even if the env gets lost/not passed between scripts. Mixing makefiles and bash scripts to makefiles etc tends to screw these up.

adilGhaffarDev: Not sure, probably just `export` might also work, but this is the recommended way of doing it. https://github.com/metal3-io/metal3-dev-env/blob/main/vars.md#environment-variables

kashifest:
Here's the passed tests :

https://jenkins.nordix.org/view/Metal3/job/metal3-centos-e2e-integration-test-main/1767/consoleFull
https://jenkins.nordix.org/view/Metal3/job/metal3-ubuntu-e2e-integration-test-main/1743/consoleFull

kashifest:
Removed the test commit now. 
/hold cancel 

scripts/ci-e2e.sh
smoshiur1237: I think `config_user.sh` is used during BMH creation from CAPM3 side. For us user is same but incase if user shell is different it will not take the export in consideration. May be that is the cause to generalize and add it in the file

PR #2867, opened 2025-10-14
🌱Fix CAPI md and conformance tests

smoshiur1237:
CAPI md and conformance tests are failing and we were unable to find image in BMH . This change will fix the the issue.


smoshiur1237:
/unhold
It is passing both CAPI Md and Conformance test

PR #2868, opened 2025-10-14
:bug: Fix for 1.10->1.11 upgrade issue

Rozzii:
This commit:
 - Implements a fallback Metal3Data retrieval process as part of the Metal3DataTemplate controller's Metal3DataClaim and Metal3Data deletion process -- The new process allows the controller to retrieve the Metal3Data based on the Metal3DataClaim's render information that supposed to point directly to the Metal3Data
 - Adds related unit test
 - Extends the existing "UpdateDatas" unit test table evaluation process to take into consideration that when a claim is deleted the related data has to be deleted too

Original issue:

If the templateReference was used before upgrade to 1.11 and Metal3Data was generated based on the `templateReference name + Metal3DataClaim index` instead of the `template name + data claim index` then after the upgrade during the first new KCP or Metal3MachineDeployment rollout CAPM3 was unable to find the old Metal3Data because after upgrade to 1.11 CAPM3 was only looking for the Metal3Data based on `template name + data claim index` only while old Metal3Data  was named based on `templateReference + index`.

In CAPM3 technically Metal3Data is allowed to get orphaned in case CAPM3 can't find them based on the information in the Metal3DataClaim and the related Metal3DataTemplate so the Metal3Data with names based on the deprecated templateRefference got orphaned and then they were holding onto IPClaims.

The presence of the orphaned Metal3Data in environments where IP pre-allocation was used resulted in the old Metal3Data holding onto the pre allocated IP claims thus the during the rollout the newly created Machies got stuck as they were unable to re-claim the IPs.


Rozzii:
/cc @lentzi90 @adilGhaffarDev @Sunnatillo @smoshiur1237 

CC-ing the release team and asking them to please include this change in 1.11.1 release.

CCing a few other folks to have as many eyes on this as possible.

/cherrt-pick release-1.11

Rozzii:
/cc @honza 
FYI

baremetal/metal3datatemplate_manager.go
```diff
@@ -179,8 +179,9 @@ func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (bool, bool, erro
 				continue
 			}
 		}
-
+		m.Log.Info("Initiate updating data of claim", "Metal3DataClaim", dataClaim.Name, "Metal3DataTemplate", m.DataTemplate.Name)
```
lentzi90: Should we maybe use some other log level for this? I am not sure how spammy it would be, but it does feel closer to "debug" than "info" for me.

baremetal/metal3datatemplate_manager.go
```diff
@@ -179,8 +179,9 @@ func (m *DataTemplateManager) UpdateDatas(ctx context.Context) (bool, bool, erro
...
 		indexes, err = m.updateData(ctx, &dataClaim, indexes)
+		m.Log.Info("Success updating data of claim", "Metal3DataClaim", dataClaim.Name, "Metal3DataTemplate", m.DataTemplate.Name)
```
lentzi90: This one also feels closer to "debug" level.

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
 	return indexes, nil
 }
 
+// Handles errors generated by Metal3Data retrieval.
+// First return value is true of Metal3Data was retrieved successfully.
```
lentzi90: nit

```suggestion
// First return value is true if Metal3Data was retrieved successfully.
```

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
 	return indexes, nil
 }
 
+// Handles errors generated by Metal3Data retrieval.
```
lentzi90: Let's make it a proper go doc.

```suggestion
// handleDeleteDataError handles errors generated by Metal3Data retrieval.
```

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
...
+		m.Log.Error(err, persistentErrMsg, "Metal3DataClaim", dataClaimName, "Metal3DataTemplate", m.DataTemplate.Name, "Metal3Data", dataName)
+		return false, persistentErrMsg
+	} else if err == nil {
+		m.Log.Info("Metal3Data found!", "Metal3DataClaim", dataClaimName, "Metal3DataTemplate", m.DataTemplate.Name, "Metal3Data", dataName)
+		return true, ""
+	}
+	return false, ""
+}
+
+// Utility to retrieve Metal3Data based on name and save the data to specified [object.
```
lentzi90: nit
```suggestion
// retrieveData is a utility function to retrieve Metal3Data based on name and save the data to specified object.
```

baremetal/metal3datatemplate_manager_test.go
```diff
@@ -270,19 +268,44 @@ var _ = Describe("Metal3DataTemplate manager", func() {
...
+
+			for _, data := range dataObjects.Items {
+				if data.DeletionTimestamp == nil || data.DeletionTimestamp.IsZero() {
+					correctDeletion := false
+					parentClaim := data.Spec.Claim.Name
+					Expect(parentClaim).NotTo(BeNil())
+					for _, claim := range dataClaimObjects.Items {
+						if parentClaim == claim.ObjectMeta.Name {
+							if claim.DeletionTimestamp != nil {
+								Expect(claim.DeletionTimestamp.IsZero()).To(BeTrue())
```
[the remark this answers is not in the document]
adilGhaffarDev: @Rozzii can you check this one?

baremetal/metal3datatemplate_manager.go
```diff
@@ -212,6 +213,7 @@ func (m *DataTemplateManager) updateData(ctx context.Context,
 			return indexes, err
 		}
 	} else {
+		m.Log.Info("Attempting to delete data of calim", "Metal3DataClaim", dataClaim.Name)
```
lentzi90: There are plenty of log lines in `deleteData`. Is this really needed? If it is, maybe a different level would be good?

baremetal/metal3datatemplate_manager.go
adilGhaffarDev: let's move this log line after the `if err != nil {`
`Success` is only when there is no error.

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
...
 	if ok {
-		// Try to get the Metal3Data. if it succeeds, delete it
-		tmpM3Data := &infrav1.Metal3Data{}
-		key := client.ObjectKey{
-			Name:      m.DataTemplate.Name + "-" + strconv.Itoa(dataClaimIndex),
-			Namespace: m.DataTemplate.Namespace,
+		// Try to get the Metal3Data, if it succeeds, delete it
+		dataName = m.DataTemplate.Name + "-" + strconv.Itoa(dataClaimIndex)
+		err := m.retrieveData(ctx, dataName, tmpM3Data)
+		m3DataFound, persistentErrMsg = m.handleDeleteDataError(err, "template name and claim index", dataClaim.Name, dataName)
```
lentzi90: The name `handleDeleteDataError` is a bit confusing for me. It sounds like it handles errors *after* trying to delete and failing. Should it rather be `handleRetrieveDataError`? I understand that it is in context of "going to delete it" though.

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
 	return indexes, nil
 }
 
+// Handles errors generated by Metal3Data retrieval.
+// First return value is true of Metal3Data was retrieved successfully.
+// Second return value contains an error message if an unknown error was encountered.
+func (m *DataTemplateManager) handleDeleteDataError(err error, filter string, dataClaimName string, dataName string) (bool, string) {
+	if err != nil && apierrors.IsNotFound(err) {
+		m.Log.Error(err, "Transient error", "Metal3Data was not found based on filter:", filter, "Metal3DataClaim", dataClaimName, "Metal3DataTemplate", m.DataTemplate.Name, "Metal3Data", dataName)
```
lentzi90: Why is it transient error? Is this for cases when it was already deleted, so we are just checking to make sure?

baremetal/metal3machine_manager.go
```diff
@@ -1801,7 +1803,8 @@ func (m *MachineManager) DissociateM3Metadata(ctx context.Context) error {
 		m.Log.Info("Unable to remove finalizers from Metal3DataClaim", "Metal3DataClaim", metal3DataClaim.Name)
 		return err
 	}
-
+	m.Log.Info("Finalizers successfully removed.", "Metal3DataClaim", metal3DataClaim.Name)
+	m.Log.Info("Initiate deletion of related claim.", "Metal3DataClaim", metal3DataClaim.Name)
```
lentzi90: These could probably be at a different level to avoid spam

lentzi90 (commented):
The logic looks good, but I have concerns about the log messages.
I understand they are useful for debugging. But we should probably use different levels for them. Especially Error logs should be carefully considered since they will always be shown, no matter the verbosity level.

baremetal/metal3datatemplate_manager.go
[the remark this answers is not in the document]
Rozzii: Done.

baremetal/metal3datatemplate_manager_test.go
```diff
@@ -270,19 +268,44 @@ var _ = Describe("Metal3DataTemplate manager", func() {
...
+			dataObjects := infrav1.Metal3DataList{}
+			opts = &client.ListOptions{}
+			err = fakeClient.List(context.TODO(), &dataObjects, opts)
+			Expect(err).NotTo(HaveOccurred())
+
+			for _, data := range dataObjects.Items {
+				if data.DeletionTimestamp == nil || data.DeletionTimestamp.IsZero() {
+					correctDeletion := false
+					parentClaim := data.Spec.Claim.Name
+					Expect(parentClaim).NotTo(BeNil())
```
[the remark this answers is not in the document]
Rozzii: Done.

baremetal/metal3datatemplate_manager_test.go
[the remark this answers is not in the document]
Rozzii: Done. my goal was different compared to the suggestion but I have simplified it.

baremetal/metal3datatemplate_manager.go
Rozzii: Done.

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
...
+func (m *DataTemplateManager) retrieveData(ctx context.Context, dataName string, tmpM3Data *infrav1.Metal3Data) error {
+	key := client.ObjectKey{
+		Name:      dataName,
+		Namespace: m.DataTemplate.Namespace,
+	}
+	return m.client.Get(ctx, key, tmpM3Data)
+}
+
 // deleteData deletes the Metal3DataClaim and marks the Metal3Data for deletion.
 func (m *DataTemplateManager) deleteData(ctx context.Context,
```
adilGhaffarDev: lets rename this to deleteMetal3DataClaim

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +343,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
 	return indexes, nil
 }
 
+// Handles errors generated by Metal3Data retrieval.
+// First return value is true of Metal3Data was retrieved successfully.
+// Second return value contains an error message if an unknown error was encountered.
+func (m *DataTemplateManager) handleDeleteDataError(err error, filter string, dataClaimName string, dataName string) (bool, string) {
```
adilGhaffarDev: Probably a better name is `handleRetrieveMetal3DataError`? Because the error we are handling here is the one that we get when we are trying to get the data.

baremetal/metal3datatemplate_manager.go
Rozzii: It is the "not found" error and there are 2 situations, when the object simply just can't be found or when the retrieval process fails for some other reason. I can rename it to "Metal3Data NOT FOUND" .
When the object is not found we technically allow that to slide but in case there is some other error then we change the status of the claim. It doesn't make much sense to me either because claim will be deleted anyhow but that was the previous logic o I jut aligned the new code to it.

Rozzii:
> The logic looks good, but I have concerns about the log messages. I understand they are useful for debugging. But we should probably use different levels for them. Especially Error logs should be carefully considered since they will always be shown, no matter the verbosity level.

I will check what I can put behind V(5) log level.

baremetal/metal3datatemplate_manager_test.go
```diff
@@ -270,19 +268,42 @@ var _ = Describe("Metal3DataTemplate manager", func() {
...
+			for _, data := range dataObjects.Items {
+				if data.DeletionTimestamp == nil || data.DeletionTimestamp.IsZero() {
+					correctDeletion := false
+					parentClaim := data.Spec.Claim.Name
+					Expect(parentClaim).NotTo(BeEmpty())
+					for _, claim := range dataClaimObjects.Items {
+						if parentClaim == claim.ObjectMeta.Name {
+							Expect(claim.DeletionTimestamp.IsZero()).To(BeTrue())
+							Expect(claim.Status.RenderedData).NotTo(BeNil())
+							correctDeletion = true
```
adilGhaffarDev, line 301: Can you explain a bit what `correctDeletion` means here?
Is this checking that all `claims` and `data` that were supposed to be deleted are indeed deleted by UpdateDatas, and that there are no remaining `claims` or `data` with a `DeletionTimestamp`?

Rozzii: Goal is to check if the a data has no deletion timestamp then the parent claim should have no deletion timestamp either.
Basically checks that claims and data deletion process is in sync or not. And also that the renderData reference ha to be still present on the claim.

baremetal/metal3datatemplate_manager.go
```diff
@@ -341,40 +349,87 @@ func (m *DataTemplateManager) createData(ctx context.Context,
...
 		if err != nil && !apierrors.IsNotFound(err) {
-			dataClaim.Status.ErrorMessage = ptr.To("Failed to get associated Metal3Data object")
+			m.Log.Error(errors.New("Unable to remove finalizer from Metal3Data"), tmpM3Data.Name)
+			return indexes, err
+		}
+		// Delete the Metal3Data
+		m.Log.V(EXTRAINFO).Info("Deleting associated Metal3Data", "Metal3DataClaim", dataClaim.Name, "Metal3DataTemplate", m.DataTemplate.Name, "Metal3Data", dataName)
+		err = deleteObject(ctx, m.client, tmpM3Data)
+		if err != nil && !apierrors.IsNotFound(err) {
+			dataClaim.Status.ErrorMessage = ptr.To("Failed to delete associated Metal3Data object")
```
lentzi90, line 424: nit, let's leave it like this for now since this was the logic already before this PR.
I think we should not really set an error message on the claim just because the data that we are trying to delete cannot be found. Sounds more like "mission accomplished" to me 😂

lentzi90 (commented):
/approve

baremetal/metal3datatemplate_manager.go
Rozzii: It sets status only if it can't execute the operation, but not when it simply can't find the object. So we are handling here the "unknown" errors,  but that said I also agree with you in the sense that IMO setting anything in the status makes no sense here as we shortly after this error handling delete the claim anyways...

PR #2869, opened 2025-10-14
:seedling: Migrate to golangci-lint v2

### LINKED ISSUE #2815: Migrate golangci-lint to v2
Ref: https://github.com/metal3-io/baremetal-operator/pull/2528

Migrate golangci-lint to v2.

/area ci
/help
/triage accepted

armagankaratosun:

Hello 👋  This PR migrates the golangci-lint tooling to v2.1.0 with the cli command:

``` bash
$ golangci-lint migrate
```

In addition, I migrated the Github workflow `.github/workflows/golangci-lint.yml` and the helper script `hack/ensure-golangci-lint.sh`  to version 2.1.0.

Now 
1. `make lint`: uses `--fast-only` (matching the new golangci-lint v2 semantics with the v1) 
2.  `make lint-full`: dropped the obsolete `--fast=false` to run the complete set without extra flags.

In addition, make lint-full resulted 50 new issues.

```bash
* nilnesserr: 1
* staticcheck: 49
```
And the `nilnesserr` was convenient to fix right away, so I did it in a separate commit. Since the others are mainly related to styling, and might be out of scope of this PR and require their own issue, I left them as-is. Please let me know if you want me to work on those, too. 

Fixes #2815

### DIFF
--- a/baremetal/metal3cluster_manager.go
+++ b/baremetal/metal3cluster_manager.go
@@ -202,7 +202,7 @@ func (s *ClusterManager) listDescendants(ctx context.Context) (clusterv1.Machine
 		}),
 	}
 
-	if s.client.List(ctx, &machines, listOptions...) != nil {
+	if err := s.client.List(ctx, &machines, listOptions...); err != nil {
 		errMsg := fmt.Sprintf("failed to list metal3machines for cluster %s/%s", cluster.Namespace, cluster.Name)
 		return machines, errors.Wrapf(err, "%s", errMsg)
 	}


Makefile
```diff
@@ -417,20 +417,20 @@ $(ENVSUBST_BIN): $(ENVSUBST) ## Build envsubst from tools folder.
...
+	$(GOLANGCI_LINT) run -v --fast-only $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=15m
+	cd $(APIS_DIR) && $(GOLANGCI_LINT) run -v --fast-only $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=15m
+	cd $(TEST_DIR) && $(GOLANGCI_LINT) run -v --fast-only $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=15m
+	cd $(FAKE_APISERVER_DIR) && $(GOLANGCI_LINT) run -v --fast-only $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=15m
 
 .PHONY: lint-fix
 lint-fix: $(GOLANGCI_LINT) ## Lint the codebase and run auto-fixers if supported by the linter
 	GOLANGCI_LINT_EXTRA_ARGS=--fix $(MAKE) lint
 
 lint-full: $(GOLANGCI_LINT) ## Run slower linters to detect possible issues
```
kashifest, line 429: Thanks a lot for the PR. I am thinking whether we should run this now in `test:` phony since this is catching more issues. `test: ` is the phony which runs on the PR and checks the linting as well. @lentzi90 @adilGhaffarDev any thoughts?

lentzi90: I think it is fine to leave like this. It is good to be fast for local checks. There may anyway be differences compared to CI so I don't think it is worth it to run the slower linters.
Note that we do not use the make targets in CI. We use the [golangci-lint action](https://github.com/metal3-io/cluster-api-provider-metal3/blob/d9a39c2b109824d1c2b27771de89fb227b5b4f38/.github/workflows/golangci-lint.yml#L32-L37) directly.

kashifest: Ah right! forgot about the workflow, then we are good

armagankaratosun:
> Thank you for this, it looks solid! Could you put the comments back in the .golangci.yaml file? They are quite helpful for understanding why some of the settings are there, or even just grouping them. For example, the import aliases had comments grouping them together and some of the exclusions had comments explaining why or what they do.

Oh sorry, my bad 😅  I didn't realize the `golangci-lint migrate`  threw away the comments in the old file. Fixed it now.

lentzi90 (commented):
Code and config looks good now. Thanks!
Could you squash the commits and add the necessary exceptions to make the tests pass?
Probably 2 commits is fine, one for migration and one for lint fix and new exceptions.

armagankaratosun:
> Code and config looks good now. Thanks! Could you squash the commits and add the necessary exceptions to make the tests pass? Probably 2 commits is fine, one for migration and one for lint fix and new exceptions.

Done. Thanks for all the help 👍

armagankaratosun:
> There is something strange going on with the changes after the squash. I don't understand it. I can see changes from commits that are already on main, which does not make sense to me. Why would it show a diff when this is already on main? 🤯 For example this commit [279eb1c](https://github.com/metal3-io/cluster-api-provider-metal3/commit/279eb1c60f87894659bfae6e2ebeda7032039c7d)
> 
> Could you maybe try a rebase on latest main?

Well, I really don't know what happened, but I think I fixed it now. Sorry for the unintentional mess.

lentzi90 (commented):
/approve

PR #2870, opened 2025-10-15
🌱 Add image pull for md remediation test

smoshiur1237:
We have seen failure in BMH creation during e2e tests because of missing image. md-remediation test usually do the test along with remediation test. In that scenario, it will pass the test. But if user wants to test it separately, I will fail because of the missing image.  This will help user to run this test separately.


smoshiur1237:
/hold
till it pass all the tests

Sunnatillo (commented):
/approve

smoshiur1237:
Tests are passing .
/unhold

PR #2874, opened 2025-10-15
🌱  Remove checking m3Data from Associate function

Sunnatillo:
From what we understand this part of the code is never reached. Because  when we reprovision for upgrade(node reuse), metal3Machine name will change and consequently the name of the DataClaim and the rest changes.

Fixes #


lentzi90 (commented):
/approve
I would say this is 🌱 . We are basically removing dead code so there is no change really.

PR #2881, opened 2025-10-20
🐛 Fix hardcoded restconfig user-agent value

Arakos:
We now correctly use the controllerName variable instead of the hardcoded literal string 'controllerName ' as the restconfigs useragent value.

Main benefit: the mergepatch fieldowner name is now correctly set to the actual operators name instead of literally 'controllerName' as was before.


adilGhaffarDev (approved):
/approve

PR #2915, opened 2025-11-05
🌱Add Kubernetes N+3 upgrade test in e2e

smoshiur1237:
PR to have upgrade scenario to do control plane upgrade to three consecutive minor versions and we will keep the worker in the initial Kubernetes version.
```
Main branch k8s-upgrade N+3 tests:

- `v1.31` => `v1.32`

- `v1.32` => `v1.33`

- `v1.33` => `v1.34`
```

Documentation is also updated accordingly

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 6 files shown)


adilGhaffarDev:
Can you also document this test in https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/e2e-test.md

adilGhaffarDev:
also this https://gerrit.nordix.org/c/infra/cicd/+/24559/1/jjb/metal3/job_capm3_periodic_e2e_k8s_upgrade_n3_tests_prow.yml  is only adding periodic, can also add pr test so we can test it before merging.

kashifest:
Sorry my memory is fading a bit on this, why did we decide to go with a separate e2e test for this instead of adding it to the existing k8s upgrade test?

adilGhaffarDev:
> Sorry my memory is fading a bit on this, why did we decide to go with a separate e2e test for this instead of adding it to the existing k8s upgrade test?

only difference I see is that in existing k8s upgrade test we are upgrading workers too but here we are only bumping CP machines. Perhaps they can be combined, first bump all cp nodes 3 times, check if everything is working fine than do worker nodes.

kashifest:
> Perhaps they can be combined, first bump all cp nodes 3 times, check if everything is working fine than do worker nodes.

Yes that was my point as well but Moshiur pointed out in an offline discussion that 
- it would increase the test time significantly
- if the test fails, investigating issues might be difficult since there are other tests also running there
- this would run in a lesser frequency i.e. once a week

I think I am ok with the new test, if we think maintainability is better here.

smoshiur1237:
> Can you also document this test in https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/docs/e2e-test.md

I planned to do it once all the steps are doing well in our CI run

smoshiur1237:
I missed the PROW config addition for PR jobs: https://github.com/metal3-io/project-infra/pull/1186 
This needs to get in first

test/e2e/common.go
```diff
@@ -1227,3 +1227,105 @@ func IsMetal3DataCountEqualToMachineCount(ctx context.Context, c client.Client,
...
+			{Key: "node-role.kubernetes.io/master", Effect: corev1.TaintEffectNoSchedule}}
+	)
+	// Upgrade process starts here
+	// Download node image
+	By("Download image")
+	imageURL, imageChecksum := EnsureImage(k8sToVersion)
+
+	By("Create new KCP Metal3MachineTemplate with upgraded image to boot")
+	m3MachineTemplateName := clusterName + "-controlplane"
+	newM3MachineTemplateName := clusterName + "-new-controlplane"
```
adilGhaffarDev: newM3MachineTemplateName needs to be changed for each upgrade. Maybe add k8s version in the name.

test/e2e/common.go
```diff
@@ -1227,3 +1227,105 @@ func IsMetal3DataCountEqualToMachineCount(ctx context.Context, c client.Client,
...
+		controlplaneTaints = []corev1.Taint{{Key: "node-role.kubernetes.io/control-plane", Effect: corev1.TaintEffectNoSchedule},
+			{Key: "node-role.kubernetes.io/master", Effect: corev1.TaintEffectNoSchedule}}
+	)
+	// Upgrade process starts here
+	// Download node image
+	By("Download image")
+	imageURL, imageChecksum := EnsureImage(k8sToVersion)
+
+	By("Create new KCP Metal3MachineTemplate with upgraded image to boot")
+	m3MachineTemplateName := clusterName + k8sFromVersion + "-controlplane"
```
adilGhaffarDev: for the first upgrade we should not add `k8sFromVersion`
I would suggest dropping `k8sFromVersion` for all upgrades.

test/e2e/common.go
```diff
@@ -1227,3 +1227,105 @@ func IsMetal3DataCountEqualToMachineCount(ctx context.Context, c client.Client,
...
+	By("Update maxSurge field in KubeadmControlPlane back to default value(1)")
+	kcpObj = framework.GetKubeadmControlPlaneByCluster(ctx, framework.GetKubeadmControlPlaneByClusterInput{
+		Lister:      clusterClient,
+		ClusterName: clusterName,
+		Namespace:   namespace,
+	})
+	helper, err = v1beta1patch.NewHelper(kcpObj, clusterClient)
+	Expect(err).NotTo(HaveOccurred())
+	kcpObj.Spec.Rollout.Strategy.RollingUpdate.MaxSurge.IntVal = 1
+	for range 3 {
```
adilGhaffarDev, line 1395: i would suggest using `Eventually` here.

adilGhaffarDev: /approve
i would not block this pr for that since we have same code in other places. This refactoring can be done is separate PR where we can fix other places too.

adilGhaffarDev: please add documentation. I will review again

test/e2e/upgrade_kubernetes_n3_test.go
```diff
@@ -0,0 +1,153 @@
...
+			E2EConfig:             e2eConfig,
+			BootstrapClusterProxy: bootstrapClusterProxy,
+			TargetCluster:         targetCluster,
+			SpecName:              specName,
+			ClusterName:           clusterName,
+			Namespace:             namespace,
+			K8sFromVersion:        upgradedK8sVersion2,
+			K8sToVersion:          upgradedK8sVersion3,
+		}
+	})
```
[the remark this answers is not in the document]
adilGhaffarDev: i agree this is better.

adilGhaffarDev (approved):
looking fine to me. Can we run copilot again just to check if everything is good now.

tuminoid:
@smoshiur1237 update PR description

Sunnatillo (commented):
/approve

Good job @smoshiur1237. 

I will lgtm when you address commnets

smoshiur1237 (commented):
Addressed the final comments. Please review. If somethings comes up for minor changes, will do it in separate PR.
/test metal3-ubuntu-e2e-kubernetes-n3-upgrade-main

PR #2918, opened 2025-11-05
🐛 Fix host selection being unreliable with more than 200 hosts in one namespace

### LINKED ISSUE #2916: Machine Association is unreliable with more than 200 BareMetalHosts in a single namespace

We are running a pretty large bare metal based setup that uses metal3. Our architecture separates sites into namespaces on our management clusters. This also includes BareMetalHost resources. For large sites, the amount of BareMetalHosts easily exceeds 200.

We've observed the association process to take a long amount of time (several minutes), which in some cases ends up producing `NotReady` Nodes in clusters (due to a combination of long deploy times and MachineHealthChecks).

**Root Cause**

When provisioning a Metal3Machine, metal3 uses a [`chooseHost`](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3machine_manager.go#L767) function to find a suitable host for the machine. This function uses the following snippet to fetch hosts:

```go
hosts := bmov1alpha1.BareMetalHostList{}
opts := &client.ListOptions{
	Namespace: m.Metal3Machine.Namespace,
	Limit:     DefaultListLimit,
}
err := m.client.List(ctx, &hosts, opts)
if err != nil {
	return nil, nil, err
}
```

`DefaultListLimit` is set to 200 (which is why I mentioned that number before).
This can lead to issues when there is more than 200 hosts, but the first 200 that are fetched here are all in use. The RNG gods might decide to mix in a few available hosts during next reconciliation, but there is no guarantee.

This issue is worsened by the lack of a label selector in the list call. The reason is explained in the next few lines.

```go
	// Using the label selector on ListOptions above doesn't seem to work.
	// I think it's because we have a local cache of all BareMetalHosts.
	labelSelector := labels.NewSelector()
```

The label selector is then applied to the list manually. The comment is outdated - label selectors work just fine with controller-runtime's client nowerdays.

**Proposed Solution**

1. Pass the label selector to the `List` call to only fetch valid BMH candidates for the Machine
2. Implement pagination support (or remove the limit) to go through all hosts if necessary.
3. Use a `fields.Selector` (and a c-r index) to filter out in-use BMHs

**Environment:**

- Cluster-api-provider-metal3 version: main

/kind bug

schrej:
Host association is currently unreliable when more than 200 hosts reside in the same namespace. This PR fixes this issue by applying the host selector directly on the List call and dropping the request limit of 200.

I've also moved the host selector creation into a function for better readability.

An alternative would be to use pagination to fetch the hosts in batches. Since controller-runtime is caching resources, the hosts are in memory already anyway, so fetching all of them at once only consumes a little extra memory, but does not strain the API server. Therefore this is the easier solution.

Further optimization can be achieved using a field selector to only list available hosts.

Tests still pass. Without the list limit, I'm not sure if it makes sense to add new tests to make sure it works with a large amount of hosts.

Fixes #2916

**Checklist:**

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -765,50 +765,19 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
 // associated with the metal3 machine. It searches all hosts in case one already has an
 // association with this metal3 machine.
 func (m *MachineManager) chooseHost(ctx context.Context) (*bmov1alpha1.BareMetalHost, *v1beta1patch.Helper, error) {
-	// get list of BMH.
-	hosts := bmov1alpha1.BareMetalHostList{}
-	// without this ListOption, all namespaces would be including in the listing.
-	opts := &client.ListOptions{
-		Namespace: m.Metal3Machine.Namespace,
-		Limit:     DefaultListLimit,
-	}
-
-	err := m.client.List(ctx, &hosts, opts)
+	labelSelector, err := hostLabelSelectorForMachine(m.Metal3Machine, m.Log)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	// Using the label selector on ListOptions above doesn't seem to work.
-	// I think it's because we have a local cache of all BareMetalHosts.
-	labelSelector := labels.NewSelector()
-	var reqs labels.Requirements
-	var r *labels.Requirement
-
-	for labelKey, labelVal := range m.Metal3Machine.Spec.HostSelector.MatchLabels {
-		m.Log.Info("Adding requirement to match label",
-			"label key", labelKey,
-			"label value", labelVal)
-		r, err = labels.NewRequirement(labelKey, selection.Equals, []string{labelVal})
-		if err != nil {
-			m.Log.Error(err, "Failed to create MatchLabel requirement, not choosing host")
-			return nil, nil, err
-		}
-		reqs = append(reqs, *r)
+	hosts := bmov1alpha1.BareMetalHostList{}
+	err = m.client.List(ctx, &hosts,
+		client.InNamespace(m.Metal3Machine.Namespace),
+		client.MatchingLabelsSelector{Selector: labelSelector},
+	)
+	if err != nil {
+		return nil, nil, err
 	}
-	for _, req := range m.Metal3Machine.Spec.HostSelector.MatchExpressions {
-		m.Log.Info("Adding requirement to match label",
-			"label key", req.Key,
-			"label operator", req.Operator,
-			"label value", req.Values)
-		lowercaseOperator := selection.Operator(strings.ToLower(string(req.Operator)))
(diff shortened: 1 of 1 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -843,26 +812,22 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmov1alpha1.BareMetal
 			}
 		}
 
-		if labelSelector.Matches(labels.Set(host.ObjectMeta.Labels)) {
```
Sunnatillo, line 846: Why are you removing` labelSelector.Matches(labels.Set(host.ObjectMeta.Labels))` condition? Could you elaborate, please?

schrej: It's now passed to the `client.List` call and the list is already filtered when we iterate. See line 776.

Sunnatillo: Great, you are moving filtration already in a list.

tuminoid:
/cc @lentzi90 @adilGhaffarDev @fmuyassarov 

IMO we should use pagination here, WDYT?

lentzi90 (commented):
Thank you, this is a very nice cleanup!
/approve
Pagination would be nice to add also as @tuminoid mentioned, but I think that can be done separately. We did not have it before so we are not regressing here in any way.

adilGhaffarDev (approved):
/lgtm

schrej:
How far away is 1.12? We can wait a bit, otherwise a cherry-pick to 1.11 would be nice.

Rozzii:
/cherry-pick release-1.10 release-1.11
It is a bugfix, there is no issue with back-porting .

Rozzii:
> @Rozzii: #2918 failed to apply on top of branch "release-1.10":
> 
> ```
> Applying: use label selector on list call during host selection
> Using index info to reconstruct a base tree...
> M	baremetal/metal3machine_manager.go
> Falling back to patching base and 3-way merge...
> Auto-merging baremetal/metal3machine_manager.go
> CONFLICT (content): Merge conflict in baremetal/metal3machine_manager.go
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> hint: When you have resolved this problem, run "git am --continue".
> hint: If you prefer to skip this patch, run "git am --skip" instead.
> hint: To restore the original branch and stop patching, run "git am --abort".
> hint: Disable this message with "git config advice.mergeConflict false"
> Patch failed at 0001 use label selector on list call during host selection
> ```

Please take a look and create a back-porting PR for 1.10 too.

schrej:
> Please take a look and create a back-porting PR for 1.10 too.

This bug [was introduced with 1.11](https://github.com/metal3-io/cluster-api-provider-metal3/pull/2646) (I didn't realize it before), so backporting to 1.10 is not necessary. We might want to revert the other cases where this was changed, or properly implement pagination instead of just limiting the requests.

Sunnatillo:
/retitle 🐛 Fix host selection being unreliable with more than 200 hosts in one namespace

PR #2921, opened 2025-11-06
🌱 Bump CAPI to v1.12.0-beta-0

Sunnatillo:

Uplifts CAPI to v1.12.0-beta-0

Fixes: #2922 

**Checklist:**

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -926,7 +926,7 @@ func consumerRefMatches(consumer *corev1.ObjectReference, m3machine *infrav1.Met
 	if consumer.Namespace != m3machine.Namespace {
 		return false
 	}
-	if consumer.Kind != "Metal3Machine" {
+	if consumer.Kind != metal3MachineKind {
 		return false
 	}
 
@@ -1097,7 +1097,7 @@ func (m *MachineManager) setHostSpec(_ context.Context, host *bmov1alpha1.BareMe
 // Metal3Machine.
 func (m *MachineManager) setHostConsumerRef(_ context.Context, host *bmov1alpha1.BareMetalHost) error {
 	host.Spec.ConsumerRef = &corev1.ObjectReference{
-		Kind:       "Metal3Machine",
+		Kind:       metal3MachineKind,
 		Name:       m.Metal3Machine.Name,
 		Namespace:  m.Metal3Machine.Namespace,
 		APIVersion: m.Metal3Machine.APIVersion,
--- a/baremetal/metal3machine_manager_test.go
+++ b/baremetal/metal3machine_manager_test.go
@@ -106,7 +106,7 @@ func consumerRef() *corev1.ObjectReference {
 	return &corev1.ObjectReference{
 		Name:       metal3machineName,
 		Namespace:  namespaceName,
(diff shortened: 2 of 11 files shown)


baremetal/metal3machine_manager.go
```diff
@@ -783,7 +788,12 @@ func (m *MachineManager) chooseHost(ctx context.Context) (*bmov1alpha1.BareMetal
 	availableHostsWithNodeReuse := []*bmov1alpha1.BareMetalHost{}
 
 	for i, host := range hosts.Items {
-		if host.Spec.ConsumerRef != nil && consumerRefMatches(host.Spec.ConsumerRef, m.Metal3Machine) {
+		ok, err1 := consumerRefMatches(host.Spec.ConsumerRef, m.Metal3Machine)
```
adilGhaffarDev: is some linter forcing not to use `err` here?
 lets use `err` is possible.

baremetal/metal3machine_manager_test.go
```diff
@@ -106,7 +106,7 @@ func consumerRef() *corev1.ObjectReference {
 	return &corev1.ObjectReference{
 		Name:       metal3machineName,
 		Namespace:  namespaceName,
-		Kind:       "M3Machine",
+		Kind:       "Metal3Machine",
```
adilGhaffarDev: "Metal3Machine" string is used so many times in this file, would be nice have it in const at top of the file.

adilGhaffarDev (approved):
/approve
minor nits otherwise fine.
I am ok with merging it how it is right now.

baremetal/metal3machine_manager.go
lentzi90: Could we just return false instead if parsing fails? If we cannot parse it, it definitely does not match. BMO should take care of validating the field so I think we can just say that it does not match.

Sunnatillo: yes. I removed checking parsing errors.

baremetal/metal3machine_manager_test.go
Sunnatillo: yes, I made unified variable for metal3Machine Kind. Please check

baremetal/metal3machine_manager.go
```diff
@@ -917,16 +917,23 @@ func hostLabelSelectorForMachine(machine *infrav1.Metal3Machine, log logr.Logger
...
 	}
-	if consumer.GroupVersionKind().Group != m3machine.GroupVersionKind().Group {
+
+	// Parse and compare Group from consumer APIVersion
+	gv, _ := schema.ParseGroupVersion(consumer.APIVersion)
+	//lint:ignore S1008 readability
+	if gv.Group != infrav1.GroupVersion.Group {
 		return false
 	}
 	return true
```
lentzi90: Let's fix the linting issue and check the error. Otherwise everyone will wonder why we ignore the error.
```suggestion
	gv, err := schema.ParseGroupVersion(consumer.APIVersion)
	if err != nil {
		// If we cannot parse it, it certainly does not match.
		return false
	}
	return gv.Group == infrav1.GroupVersion.Group
```

Sunnatillo: done

PR #2938, opened 2025-11-14
🌱  Bump Crypto to v0.43.0 and Go to 1.24 to fix vulnerability

Sunnatillo:

  Uplifts Crypto to v0.43.0 and Go to 1.24 to fix vulnerability

Fixes #

**Checklist:**

### DIFF
--- a/api/v1beta1/metal3machine_types_test.go
+++ b/api/v1beta1/metal3machine_types_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1beta1
 
 import (
-	"context"
 	"testing"
 
 	"github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


Sunnatillo:
> This is a minor go version bump. Can we really do it on a release branch? I think we usually avoid that.

Bump Crypto to v0.43.0 is a vulnerability fix. And it specifically requires v1.24.9 version. I was hesitant at first. 
I am not sure. Should we uplift or not?

Rozzii:
> > This is a minor go version bump. Can we really do it on a release branch? I think we usually avoid that.
> 
> Bump Crypto to v0.43.0 is a vulnerability fix. And it specifically requires v1.24.9 version. I was hesitant at first, because we use v1.24.9 in Dockerfile. I am not sure. Should we uplift or not?

Also the go.mod's go version bump is quite big for a patch release so we might need some discussion about it with the whole release team and maintainers.

Rozzii:
> > > This is a minor go version bump. Can we really do it on a release branch? I think we usually avoid that.
> > 
> > 
> > Bump Crypto to v0.43.0 is a vulnerability fix. And it specifically requires v1.24.9 version. I was hesitant at first, because we use v1.24.9 in Dockerfile. I am not sure. Should we uplift or not?
> 
> Also the go.mod's go version bump is quite big for a patch release so we might need some discussion about it with the whole release team and maintainers.

Let's wait until monday, My assesment is that although in my eyes fixing the CVE would justify the go.mod GO version bump, the CVE description tells me that we are not affected even do this is a 7,5 CVE as we are not using the effected lib to facilitate SSH connections. So at least IMO we don't need to hurry this let's wait with the patch release until Monday let's the actual security expert take a look.
vuln: https://github.com/advisories/GHSA-56w8-48fp-6mgv

Rozzii:
@adilGhaffarDev  noticed CVE is only affecting our e2e test code not he actual application so 
/lgtm cancel
We can skip this bump.

api/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/metal3-io/cluster-api-provider-metal3/api
 
-go 1.23.7
+go 1.24.9
```
tuminoid, line 3: We shouldn't do this in release branches

tuminoid (commented):
Not really vulnerable, thus we don't bump the go.mod versioning to avoid breaking downstream users.

/close

PR #2959, opened 2025-11-21
:bug: Remove invalid kustomizeconfig from config/webhook

mdbooth:

This change has no effect on the output of this kustomization because
the removed configuration was redundant. However, it fixes a bug which
can be triggered when using this kustomization as a base for another
kustomization.

kustomizeconfig contained 3 directives:

* nameReference
* namespace
* varReference

varReference is redundant because vars are no longer used in this
kustomize: they are deprecated and usage was removed some time ago.

nameReference is redundant because the specified configuration is
already in kustomize's defaults. However, nameReference is the important
transformation here.

namespace is incorrect. It directs the namespace transformer to update
webhooks/clientConfig/service/namespace. However, this is not the
intended function of the namespace transformer: it should only set the
namespace directly on objects and allow references to be updated
automatically by nameReference. Configuring it to update a reference
directly leaves kustomize with inconsistent internal state. Depending on
execution order this can cause a subsequent transformation to fail to
update the reference when it makes further changes to the Service
object.

This issue is common amongst multiple providers, likely because it was present in an early version of the kubebuilder templates.

This PR is based on https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/5982

To confirm that the removed configuration was redundant:
* Execute `make release-manifests`
* Copy `out/infrastructure-components.yaml` to /tmp
* Apply/revert this change
* Execute `make release-manifests` again
* diff the outputs and confirm that they are identical


tuminoid:
Please fix the DCO.

PR #3078, opened 2026-01-13
🌱 Document CAPM3 pod placement and add optional affinity example

### LINKED ISSUE #940: CAPM3 (capm3-system namespace pods) should not be running among workloads
As discussed in metal3-sec group related to https://github.com/metal3-io/baremetal-operator/pull/1241 CAPM3/IPAM (capm3-system namespace) should not be running among user workloads, but with k8s controlplane or infra nodes.

CAPM3/IPAM (capm3-system) is running on controlplane or in infra nodes, not among regular workload. These should be sensible default value, and documentation should be available to explain the options.

**Environment:**

- CAPM3 version: any/latest
- Environment (metal3-dev-env or other): any

/kind bug
/kind documentation
/assign @lentzi90

vishalanarase:

Clarifies and improves how CAPM3 controller pods can be scheduled away from regular
workloads without enforcing environment-specific defaults. The PR keeps the default
deployment aligned with upstream CAPI (tolerations only, no node affinity), and
documents how operators can optionally harden placement (for example to control-plane
or infra nodes) using their preferred deployment tooling

Changes:
- Add a new optional example patch
  `examples/provider-components/manager_node_affinity_patch.yaml` that shows how to:
  - Require scheduling on control-plane and/or infra nodes
  - Prefer infra nodes first, then control-plane nodes
- Add a “Pod Placement Configuration” section to `docs/getting-started.md` that:
  - Describes the default behavior (tolerations only, no affinity)
  - Explains how and why to harden placement if desired
  - Links to CAPI Operator and `clusterctl` documentation for provider customization
  - Points to `examples/provider-components/` as a starting point for kustomize-based
    workflows
- Update `examples/provider-components/manager_tolerations_patch.yaml` comments to
  reference the new affinity example and clarify usage.
  

Fixes #940

**Checklist:**

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -216,6 +216,49 @@ WORKERS_KUBEADM_EXTRA_CONFIG="
 "
 ```
 
+## Pod Placement Configuration
+
+By default, CAPM3 controller pods are configured with **tolerations** so they
+can run on control-plane nodes (which are typically tainted), but there is **no
+default node affinity**. This means pods may still be scheduled on regular
+worker nodes unless additional scheduling constraints are applied by the admin.
+
+### Default Configuration
+
+The default deployment includes:
+
+- **Tolerations**: Allow pods to run on control-plane nodes (which typically
+   have taints)
+   - `node-role.kubernetes.io/master:NoSchedule`
+   - `node-role.kubernetes.io/control-plane:NoSchedule`
+
+- **Node Affinity**: Not set by default. Pods can run on any schedulable node
+   (control-plane or worker), depending on the cluster’s taints and labels.
+
+### Customizing Pod Placement
+
+If you want to ensure that CAPM3 pods do **not** run alongside regular
+workloads, you can:
+
+- Add node affinity and/or additional tolerations via your preferred deployment
+   mechanism (CAPI Operator, `clusterctl`, or kustomize), and
+- Use dedicated infrastructure nodes if your environment provides them.
+
+For more details on how to customize provider manifests, see the upstream
+Cluster API documentation:
+
+- **CAPI Operator provider configuration**:
+   [Provider spec configuration docs](https://cluster-api-operator.sigs.k8s.io/topics/configuration/provider-spec-configuration#provider-spec)
+- **`clusterctl` config overrides**:
+   [`clusterctl` configuration overrides](https://cluster-api.sigs.k8s.io/clusterctl/configuration#overrides-layer)
+- **`clusterctl generate provider`**:
+   [`clusterctl generate provider` command reference](https://cluster-api.sigs.k8s.io/clusterctl/commands/generate-provider)
+
+For kustomize-based workflows, you can use the CAPM3 example patches in
+[`examples/provider-components/`](https://github.com/metal3-io/cluster-api-provider-metal3/tree/main/examples/provider-components)
+as a starting point. The `manager_node_affinity_patch.yaml` example shows how
+to require control-plane or infra nodes and prefer infra nodes when available.
+
 ## Pivoting or updating Ironic
 
 Before running the `move` command of Clusterctl, elements such as Ironic if
--- a/examples/provider-components/manager_node_affinity_patch.yaml
+++ b/examples/provider-components/manager_node_affinity_patch.yaml
@@ -1,0 +1,41 @@
+---
+# This patch adds node affinity to require CAPM3 controller pods to run only on
+# control-plane or infra nodes instead of regular worker nodes.
+# Within those eligible nodes, it prefers dedicated infra nodes first, then
+# control-plane nodes.
+#
+# To use this patch, apply it after deploying CAPM3:
+# kubectl patch deployment capm3-controller-manager -n capm3-system --patch-file examples/provider-components/manager_node_affinity_patch.yaml
+#
+# Or use it with kustomize by adding it as a patch in your kustomization.yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: capm3-controller-manager
+  namespace: capm3-system
+spec:
+  template:
+    spec:
+      affinity:
+        nodeAffinity:
+          # Require pods to run on control-plane or infra nodes
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
+            - matchExpressions:
+              - key: node-role.kubernetes.io/infra
+                operator: Exists
+          # Prefer dedicated infra nodes first, then control-plane nodes
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/infra
+                operator: Exists
+          - weight: 50
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
--- a/examples/provider-components/manager_tolerations_patch.yaml
+++ b/examples/provider-components/manager_tolerations_patch.yaml
@@ -1,4 +1,9 @@
 ---
+# This patch adds tolerations to allow CAPM3 controller pods to run on
+# control-plane nodes (which typically have taints).
+#
+# Note: The default deployment already includes tolerations for control-plane nodes.
+# This patch is provided for reference or if you need to add additional tolerations.
 apiVersion: apps/v1
 kind: Deployment
 metadata:


config/manager/manager.yaml
```diff
@@ -67,3 +67,32 @@ spec:
...
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
+          - weight: 50
+            preference:
+              matchExpressions:
+              - key: node-role.kubernetes.io/infra
+                operator: Exists
```
tuminoid: Missing EOL

vishalanarase: Added

examples/provider-components/manager_node_affinity_patch.yaml
```diff
@@ -0,0 +1,49 @@
+---
+# This patch adds node affinity to ensure CAPM3 controller pods prefer
+# control-plane or infra nodes over regular worker nodes.
+# This is useful when you want to ensure CAPM3 pods run on dedicated
+# infrastructure nodes rather than worker nodes.
```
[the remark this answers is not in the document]
vishalanarase: Done

docs/getting-started.md
```diff
@@ -216,6 +216,84 @@ WORKERS_KUBEADM_EXTRA_CONFIG="
...
+   kubectl label node <node-name> node-role.kubernetes.io/infra=""
+   ```
+
+2. Taint the infrastructure nodes to prevent regular workloads:
+   ```bash
+   kubectl taint nodes <node-name> node-role.kubernetes.io/infra:NoSchedule
+   ```
+
+3. Apply the tolerations patch (already included in default deployment) to allow
+   CAPM3 pods to run on these tainted nodes.
```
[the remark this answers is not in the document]
vishalanarase: Done

docs/getting-started.md
```diff
@@ -216,6 +216,84 @@ WORKERS_KUBEADM_EXTRA_CONFIG="
...
+  taints)
+  - `node-role.kubernetes.io/master:NoSchedule`
+  - `node-role.kubernetes.io/control-plane:NoSchedule`
+
+- **Node Affinity**: Require pods to run on control-plane or infrastructure nodes
+  - **Required**: Pods will only schedule on nodes with one of these labels:
+    - `node-role.kubernetes.io/control-plane`
+    - `node-role.kubernetes.io/master` (for older clusters)
+    - `node-role.kubernetes.io/infra` (for dedicated infra nodes)
+  - **Preferred**: Among eligible nodes, prefers control-plane nodes first, then
```
lentzi90: I think it would make more sense to prefer dedicated infra nodes if they exist, to avoid risk of interfering with the control-plane.

examples/provider-components/manager_tolerations_patch.yaml
```diff
@@ -1,4 +1,15 @@
...
+# control-plane nodes (which typically have taints).
+#
+# Note: The default deployment already includes tolerations for control-plane nodes.
+# This patch is provided for reference or if you need to add additional tolerations.
+#
+# For node affinity configuration to prefer/require control-plane or infra nodes,
+# see manager_node_affinity_patch.yaml
+#
+# To use this patch, apply it after deploying CAPM3:
+# kubectl patch deployment capm3-controller-manager -n capm3-system --patch-file examples/provider-components/manager_tolerations_patch.yaml
```
lentzi90: I think it is better to avoid mentioning the other patch. This mention risks getting outdated.
I would also avoid suggesting specific deployment strategies, especially those involving post-deploy patches.

```suggestion
```

docs/getting-started.md
```diff
@@ -216,6 +216,84 @@ WORKERS_KUBEADM_EXTRA_CONFIG="
...
+    infra nodes
+
+This ensures CAPM3 pods do not run on regular worker nodes and only schedule on
+control-plane or infrastructure nodes. The tolerations ensure pods can run on
+tainted control-plane nodes.
+
+### Customizing Pod Placement
```
lentzi90, line 238: These three sections below are too generic (i.e. they are not metal3 specific). I think it would be better to link to Kubernetes or CAPI docs instead of trying to duplicate them. On the other hand, it is also very likely that users do not deploy CAPM3 using kustomize and kubectl. If they do, they probably already know how to handle these things. If they rather deploy using `clusterctl init` or the CAPI operator, then these docs are useless regardless.

Maybe it is best to just leave this out? It is then up to the users to figure out how to do it based on the deployment method they use. If we want to give some guidance, I would suggest linking to these:
- CAPI opedator way: https://cluster-api-operator.sigs.k8s.io/topics/configuration/provider-spec-configuration#provider-spec
- clusterctl config overrides: https://cluster-api.sigs.k8s.io/clusterctl/configuration#overrides-layer
- clusterctl generate provider: https://cluster-api.sigs.k8s.io/clusterctl/commands/generate-provider
- the kustomize examples here below

vishalanarase:
> This looks good to me now! Please squash the commits and check the markdown linter errors. It is just complaining about some indents so should be easy to fix.
> 
> I think we should mark this as ⚠️ instead of 🐛 since it can really make CAPM3 unschedulable. What do others think?

Done

lentzi90 (commented):
/approve

tuminoid:
/test metal3-ubuntu-e2e-capi-md-test-main
/test metal3-ubuntu-e2e-kubernetes-n3-upgrade-main
/test metal3-e2e-in-place-upgrade-test-main
/test metal3-e2e-conformance-test-main

Let's run few more to be sure the CI jobs are properly for this.

tuminoid:
/hold
To allow extra tests to finish.

tuminoid:
/override metal3-e2e-in-place-upgrade-test-main
In-place is broken in CI actually, ignore that.

config/manager/manager.yaml
```diff
@@ -67,3 +67,32 @@ spec:
...
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
+            - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
```
fmuyassarov: considering the fact that this label is no more active and deprecated, do we still want to add affinity for that?

config/manager/manager.yaml
```diff
@@ -67,3 +67,24 @@ spec:
...
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
+            - matchExpressions:
+              - key: node-role.kubernetes.io/infra
+                operator: Exists
+          preferredDuringSchedulingIgnoredDuringExecution:
```
fmuyassarov: Is this some well-known label? I feel like this is a custom label, if so I would not add it here and instead let the users define it on their deployments.

fmuyassarov: The only reference i could find is in OpenShift docs but not in the k/k.

lentzi90: `node-role.kubernetes.io/*` is well known: https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io. There are at least some k8s distributions that use specifically `/infra` (e.g. openshift). I would say it is well known enough, and it is convenient to have some affinity like this by default. Of course, the user can choose something else and modify the manifests accordingly. I do think it is better to have it here by default than to only have the control-plane affinity though. It gives users an easy way to deploy on non-control-plane nodes without changing the manifests.

tuminoid:
> /test metal3-ubuntu-e2e-capi-md-test-main /test metal3-centos-e2e-feature-test-main-features /test metal3-ubuntu-e2e-kubernetes-n3-upgrade-main /test metal3-e2e-conformance-test-main /test metal3-centos-e2e-integration-test-main metal3-ubuntu-e2e-integration-test-main

Just for storing a reference, all of these have passed, in case we push some minor tweaks.

lentzi90:
We seem to have some disagreement about if/how this should actually be done. Let's discuss a bit on the issue first.
/approve cancel

config/manager/manager.yaml
fmuyassarov: Copying what I said on the issue here: https://github.com/metal3-io/cluster-api-provider-metal3/issues/940

I think this request is trying to lock the capm3 controller to control plane nodes, filtered via well-known labels like `node-role.kubernetes.io/control-plane`. While I understand the intention of keeping the controller pods on control plane nodes, I believe this kind of restriction should be handled on the consumer side, not here.

Environments differ, and an affinity rule that works for one setup may not work for another. In my opinion, we should provide the bare minimum YAML that works across environments and let consumers/admins decide where to place the pods. In other words, admins should define their own affinity rules and patch the manifests before deployment.

Unless there’s a strong reason to enforce such restrictions by default, I think we should aim for provider/environment-agnostic manifests.

fmuyassarov: and my original thinking wasn't about `node-role.kubernetes.io/control-plane` but rather `node-role.kubernetes.io/infra` which doesn't exist on the k/k documentation.

tuminoid (commented):
/lgtm

Looks good to me. Thank you and sorry for the pivot!
/cc @lentzi90

lentzi90:
/hold cancel
Cancelling the hold since this does not touch manifests that are used in the tests anyway.

vishalanarase:
> LGTM
> 
> Could you please add this to book as well? https://github.com/metal3-io/metal3-docs/tree/main/docs/user-guide

I’ll open a follow‑up PR against metal3-docs and add a “CAPM3 controller pod placement” section to the user guide with the same explanation and links.

PR #3081, opened 2026-01-13
🐛 Fix Metal3Cluster webhook blocking cloudProvider toggle updates

MaxRink:
The validating webhook incorrectly rejected updates when toggling the cloudProviderEnabled field from true to false or vice versa. The cross-field validation was comparing new values against old values of the opposite field (e.g., newM3C.CloudProviderEnabled vs oldM3C.NoCloudProvider), which would always fail when legitimately changing the cloudProvider setting.

Remove the incorrect cross-field validation between old and new objects. The internal consistency check (ensuring cloudProviderEnabled and noCloudProvider are opposite values in the new object) is sufficient, especially since the defaulting webhook keeps both fields in sync.

Add test cases to verify:
- Enabling cloudProvider from disabled state works
- Disabling cloudProvider from enabled state works
- Partial updates with only one field changing work
- Internal conflicts (both fields set to same value) are still caught

### DIFF
--- a/internal/webhooks/v1beta1/metal3cluster_webhook.go
+++ b/internal/webhooks/v1beta1/metal3cluster_webhook.go
@@ -97,7 +97,7 @@ func (webhook *Metal3Cluster) ValidateDelete(_ context.Context, _ runtime.Object
 	return nil, nil
 }
 
-func (webhook *Metal3Cluster) validate(oldM3C, newM3C *infrav1.Metal3Cluster) error {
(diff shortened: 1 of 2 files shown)


internal/webhooks/v1beta1/metal3cluster_webhook.go
adilGhaffarDev:
```suggestion
func (webhook *Metal3Cluster) validate(_, newM3C *infrav1.Metal3Cluster) error {
```

kashifest:
Failure is unrelated to change, trying again 

Issue #2671, opened 2025-07-30
Tasks for v1beta2 API


adilGhaffarDev:
- [x] Create v1beta2 types & use them in controllers/webhooks
  - [x] Add CAPM3 v1beta2 types (inlc. conversion + imports) ref: https://github.com/kubernetes-sigs/cluster-api/pull/12037
    - [x] PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3110  
  - [x] Promote v1Beta2Conditions. ref: https://github.com/kubernetes-sigs/cluster-api/pull/12066
    - [x] Promote conditions
      - [x] PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3133 
    - [x] Rename condition constants (example AssociateBareMetalHostV1Beta2Condition -> AssociateBareMetalHostCondition) ref: https://github.com/kubernetes-sigs/cluster-api/pull/12091 
      - [x] PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3215 
    - [ ] Use new conditions / audit usage of old conditions. ref: https://github.com/kubernetes-sigs/cluster-api/pull/12109/changes 
  - [x] Ensure v1beta1 API types are not using clusterv1beta2 types (& vice versa)
  - [x] Stop using v1beta1 status in controllers (of CAPI & CAPM3 CRDs)
  - [x] Deprecate v1beta1 packages
    - [x] PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3138 
  - [x] Drop DefaulterRemoveUnknownOrOmitableFields option
    - [x] PR: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3139 

- [x] Implement v1beta2 contract: https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.10-to-v1.11#cluster-api-contract-changes 
  - [x] Implement v1beta2 contract for Metal3Machine and Metal3Cluster
    - [x] Metal3Cluster
      - [x] control plane endpoint (use copy of APIEndpoint struct)
      - [x] failure domains (check if anything needs to be updted there)
      - [x] initialization completed
      - [x] terminal failures (move failureReason/Message to deprecated)
    - [x] Metal3Machine
      - [x] Ensure we implemented addresses correctly (correct field type)
      - [x] provider ID, *string to string.
      - [x] initialization completed
      - [x] terminal failures (move failureReason/Message to deprecated)
[...]

PR #3110, opened 2026-01-23
✨ Add v1beta2 API type

adilGhaffarDev:
Add v1beta2 API type

**Checklist:**


adilGhaffarDev:
@kashifest @lentzi90

adilGhaffarDev:
NOTE:
We have to fix FromPool by making APIGroup and Kind not mandatory and set default value for them, APIGroup: "ipam.metal3.io" and kind: "IPPool"
When conversion was added it started failing because we were not setting APIGroup and kind in IPAddressesFromPool and PrefixesFromPool. Controller code is defaulting to APIGroup: "ipam.metal3.io" and kind: "IPPool" but on api level we are not setting it. We should set these defaults on api level.

I have added this to this issue: https://github.com/metal3-io/cluster-api-provider-metal3/issues/2671

api/v1beta2/api_suite_test.go
lentzi90, line 1: This should be renamed. We could change it to `v1beta2_suite_test.go` or perhaps even drop the v1betaX from the name, since it is anyway in the folder. Perhaps `api_suite_test.go`?

adilGhaffarDev:
> How can the tests work with the `""` API group and kind? Is it because the controllers see the empty string and use the default instead? Other than that I only have one comment below. The rest all looks good.

yes when API group and kind are set to `""` controllers defaults to APIGroup: "ipam.metal3.io" and kind: "IPPool" .

adilGhaffarDev:
> > How can the tests work with the `""` API group and kind? Is it because the controllers see the empty string and use the default instead? Other than that I only have one comment below. The rest all looks good.
> 
> yes when API group and kind are set to `""` controllers defaults to APIGroup: "ipam.metal3.io" and kind: "IPPool" .

even when they are not set(nil) controller defaults to APIGroup: "ipam.metal3.io" and kind: "IPPool" .

lentzi90:
Tip for other reviewers: Checkout the PR locally so you can diff the API folders. There are basically no differences between v1beta1 and v1beta2 for now and this way of comparing makes it obvious.

api/v1beta2/api_suite_test.go
adilGhaffarDev: done

PR #3134, opened 2026-02-03
:seedling: bump github actions to new major versions

tuminoid:
Dependabot isn't allowed to bump to new major versions automatically in case there is a breaking change. In these updates, we don't have any breaking changes right now as the runner version is new enough to support them.


lentzi90 (commented):
/approve

tuminoid:
/override metal3-ubuntu-e2e-integration-test-main metal3-centos-e2e-integration-test-main

smoshiur1237 (commented):
/lgtm

Issue #3167, opened 2026-02-19
Node Reuse Flakiness


adilGhaffarDev:
We’re observing flakiness in the node_reuse test, it is coming from CAPI. The feature is flaky not the test itself.

CAPI is expected to set the following annotations on Metal3Machine objects:
 * cluster.x-k8s.io/cloned-from-groupkind: Metal3MachineTemplate.infrastructure.cluster.x-k8s.io
 * cluster.x-k8s.io/cloned-from-name: test1-workers

However, these annotations are flaky on the CAPI side, in some cases they are missing on the Metal3Machine.
Our node-reuse logic is heavily dependent on these annotations.
 * We check for these annotations before setting the node-reuse label on the BareMetalHost (BMH).
 * If the annotations are missing, the label is not set.
 * If the label is not set, node reuse does not happen.

This leads to test instability in CI. The issue does not happen frequently enough to fail consistently, but it does occur intermittently. Examples:
* https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-centos-e2e-feature-test-main-pivoting/500/
* https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-centos-e2e-feature-test-main-pivoting/495/

There is also another edge case:
In some scenarios, node reuse does not happen (i.e., the BMH label is never set), but the randomly selected BMH happens to be the same one we expected. This can mask the failure and make the test appear successful.
Example:
* https://jenkins.nordix.org/view/Metal3/job/metal3-centos-e2e-feature-test-main-pivoting/110/consoleFull

**What steps did you take and what happened:**
To see the missing annotation you can run make of metal3-dev-env again and again, soon you will see the missing annotations on or more metal3machines. When you see missing annotation do node reuse it will fail.

**What did you expect to happen:**
Node reuse feature should work.

**Anything else you would like to add:**
Proposed fix:
1. Make Template Resolution Deterministic:
 Instead of relying on the cloned-from-* annotations (see current check here: https://github.com/metal3-io/cluster-api-provider-metal3/blob/15ef6e7b9d8d46ad5bc665f234184ff6a2f79589/baremetal/metal3machine_manager.go#L608), we could fetch the Metal3MachineTemplate through a more deterministic path — for example via the MachineDeployment. This would require a bit more code, but it would remove reliance on flaky annotations and make the behavior deterministic.
 
2. Improve Observability of Node-Reuse Success:
[...]

PR #3168, opened 2026-02-22
:warning: Make node-reuse more deterministic

adilGhaffarDev:

Fixes https://github.com/metal3-io/cluster-api-provider-metal3/issues/3167

**Checklist:**

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/selection"
@@ -93,7 +94,7 @@ type MachineManagerInterface interface {
 	IsBaremetalHostProvisioned(context.Context) bool
 	IsBootstrapReady() bool
 	MachineHasNodeRef() bool
-	Associate(context.Context) error
+	Associate(context.Context) (string, error)
 	Delete(context.Context) error
 	Update(context.Context) error
 	HasAnnotation() bool
@@ -340,7 +341,7 @@ func (m *MachineManager) SetPauseAnnotation(ctx context.Context) error {
 }
 
 // Associate associates a machine and is invoked by the Machine Controller.
-func (m *MachineManager) Associate(ctx context.Context) error {
+func (m *MachineManager) Associate(ctx context.Context) (string, error) {
 	// Parallel attempts to associate is problematic since the same BMH
 	// could be selected for multiple M3Ms. Therefore we use a mutex lock here.
 	associateBMHMutex.Lock()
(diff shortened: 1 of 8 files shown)


adilGhaffarDev:

cc @lentzi90 @kashifest @Rozzii

baremetal/metal3machine_manager.go
```diff
@@ -93,7 +93,7 @@ type MachineManagerInterface interface {
 	IsBaremetalHostProvisioned(context.Context) bool
 	IsBootstrapReady() bool
 	MachineHasNodeRef() bool
-	Associate(context.Context) error
+	Associate(context.Context) (string, error)
```
lentzi90, line 97: I think this should be marked as a breaking change. Is it necessary, or could we get the information another way?

baremetal/metal3machine_manager.go
```diff
@@ -821,12 +819,12 @@ func getHost(ctx context.Context, m3Machine *infrav1.Metal3Machine, cl client.Cl
 // chooseHost iterates through known hosts and returns one that can be
 // associated with the metal3 machine. It searches all hosts in case one already has an
 // association with this metal3 machine.
-func (m *MachineManager) chooseHost(ctx context.Context) (*bmov1alpha1.BareMetalHost, *patch.Helper, error) {
+func (m *MachineManager) chooseHost(ctx context.Context) (*bmov1alpha1.BareMetalHost, *patch.Helper, string, error) {
```
lentzi90, line 843: We should add to the doc string above what the return values are.

baremetal/metal3machine_manager.go
```diff
@@ -1975,6 +1970,67 @@ func (m *MachineManager) getMachineSet(ctx context.Context) (*clusterv1.MachineS
...
+	// Check if this is a control plane machine.
+	for _, mOwnerRef := range m.Machine.ObjectMeta.OwnerReferences {
+		aGV, err := schema.ParseGroupVersion(mOwnerRef.APIVersion)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse the group and version: %w", err)
+		}
+		if aGV.Group == controlplanev1.GroupVersion.Group {
+			// This is a KubeadmControlPlane machine.
+			kcp := &controlplanev1.KubeadmControlPlane{}
+			key := client.ObjectKey{
```
lentzi90: We should support other control plane providers also. Maybe we can use "unstructured"? The same way CAPI handles all infra providers.

adilGhaffarDev:
> I did not quite realize the implications of this earlier, sorry about that. I am not sure we can go this way. We would be directly tide to KubeadmControlPlane, and no other control-plane providers would work. Maybe there is a way to do this without restricting to only KCP, I am not sure. Perhaps the better solution would be to go upstream and see if we can make the annotations more predictable in CAPI?

good point regarding restricting to KubeadmControlPlane, for now in this PR I will use unstructured.
For later I will try to collect CAPI controller logs when this issue happens and report the issue in the CAPI. If I get time I will try to reproduce with CAPD too.

baremetal/metal3machine_manager.go
adilGhaffarDev: This function is used by controller internally, not sure how this is breaking change?

baremetal/metal3machine_manager.go
adilGhaffarDev: done

baremetal/metal3machine_manager.go
adilGhaffarDev: fixed with unstructured please check.

baremetal/metal3machine_manager.go
lentzi90: Well, if anyone is using CAPM3 as a go module, then they are probably using this interface and then this is breaking for them. This code is available as a public module at least.
https://pkg.go.dev/github.com/metal3-io/cluster-api-provider-metal3@v1.12.2/baremetal

adilGhaffarDev: i have marked it as breaking change

adilGhaffarDev:
@kashifest @Sunnatillo @Rozzii please check

baremetal/metal3machine_manager.go
kashifest: interesting I also thought this Associate function should be internal to the controller but looks like its not

Rozzii:
Unrelated dev-env bootstrapping issue.
/lgtm.

Looks about right .

PR #3188, opened 2026-02-26
📖 Update providerID documentation and add legacy format deprecation notice

kashifest:
Adds ProviderID workflow reference and deprecate legacy format

**Checklist:**

### DIFF
--- a/api/v1beta1/metal3machine_types.go
+++ b/api/v1beta1/metal3machine_types.go
@@ -115,7 +115,9 @@ const (
 // Metal3MachineSpec defines the desired state of Metal3Machine.
 type Metal3MachineSpec struct {
 	// ProviderID will be the Metal3 machine in ProviderID format
-	// (metal3://<bmh-uuid>)
+	// (metal3://<namespace>/<bmh-name>/<m3m-name>).
+	// The legacy format (metal3://<bmh-uuid>) will be deprecated in CAPM3 v1.13
+	// and removed in CAPM3 v1.14.
 	// +optional
 	ProviderID *string `json:"providerID,omitempty"`
 
--- a/api/v1beta2/metal3machine_types.go
+++ b/api/v1beta2/metal3machine_types.go
@@ -119,7 +119,9 @@ const (
 // Metal3MachineSpec defines the desired state of Metal3Machine.
 type Metal3MachineSpec struct {
 	// ProviderID will be the Metal3 machine in ProviderID format
-	// (metal3://<bmh-uuid>)
+	// (metal3://<namespace>/<bmh-name>/<m3m-name>).
+	// The legacy format (metal3://<bmh-uuid>) will be deprecated in CAPM3 v1.13
+	// and removed in CAPM3 v1.14.
 	// +optional
 	ProviderID *string `json:"providerID,omitempty"`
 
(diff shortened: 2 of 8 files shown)


docs/providerid-workflow.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Step 2 — Cloud provider path
+
+This step is taken when `Metal3Cluster.spec.cloudProviderEnabled` is `true`, or
+`Metal3Cluster.spec.noCloudProvider` is explicitly `false`.
+
+When a cloud provider is present, it is responsible for setting `spec.providerID`
+on the workload-cluster node (using the BMH UID). CAPM3 then finds that node by
+its ProviderID, copies the value to `Metal3Machine.spec.providerID`, and marks
+the machine ready.
```
lentzi90: This seems backwards, but I guess this is how we actually do it. IMO CAPM3 should set the `Metal3Machine.spec.providerID` directly after associating and not care anything about the node. Then IF it acts as cloud provider also, it should set the provider ID on the node to the same thing it set on the M3M.

Is the assumption that an external cloud provider always uses the UID? Then how can we remove this provider ID format?

docs/providerid-workflow.md
```diff
@@ -0,0 +1,194 @@
...
+   not finished setting the label.
+1. If more than one node is found with the same label, a hard error is returned
+   — this is a misconfiguration.
+1. If exactly one node is found:
+   - If the node has no `spec.providerID` yet: sets
+     `Metal3Machine.spec.providerID` to the **new format**, marks the machine
+     ready, then patches `node.spec.providerID` on the workload cluster.
+   - If the node already has a ProviderID matching the new format: copies it to
+     `Metal3Machine.spec.providerID` and marks the machine ready (covers the
+     CAPI pivot / move scenario).
```
lentzi90: I am not understanding why this is needed. Only the status goes away after move, so the providerID should still be there, right? I think this was why we switched the format. With the UID it could change and then this was needed. But with the new format it should be fine to skip this.

kashifest: Yes, thats what I meant, by adding the new format we cover the pivot scenario when UUID changes

docs/providerid-workflow.md
kashifest: Good catch, double checked the code, Code actually accepts both format currently. Updated the text accordingly. So, once we deprecate we need also tighten this part of the code to not accept any other format than the valid one from Cloud providers.

docs/providerid-workflow.md
lentzi90: Ok, so this point is only for switching from UID to new format. Got it :+1:

docs/providerid-workflow.md
lentzi90, line 1: What do you think about moving this whole file to metal3-docs instead?

kashifest: I was thinking the same but its not a Feature feature so in that sense kept it here. I am open to suggestions

lentzi90: I think it would be better in the book than here. People will not find it otherwise.
For CAPM3 we don't really have any docs like these yet, but for BMO there is the state machine page. I think we should just make a similar page for CAPM3 to collect things like this.

kashifest: Pushed a PR here https://github.com/metal3-io/metal3-docs/pull/663

kashifest:
The link checker shall pass once the metal3-docs PR merges.
/hold

kashifest:
/hold cancel
Mostly doc change 

PR #3193, opened 2026-03-02
:seedling: Bump golang for image build to v1.25.8

kashifest:
- Bumps golang for image build to v1.25.8
- bump osv-scanner to v2.3.3
- golangci-lint older than v2.4.0 was compiled with Go 1.24.4. When linting the
test/ module with Go 1.25.8 installed, golangci-lint panics:

  panic: file requires newer Go version go1.25 (application built with go1.24)

This is triggered because golang.org/x/crypto@v0.45.0 (pulled in via
helm/v3 and docker/docker in test/go.mod) contains files with
//go:build go1.25 constraints. With Go 1.25 installed these files are
included in the build graph, but golangci-lint's embedded go/types
checker (compiled with Go 1.24) cannot handle them.

v2.4.0 is the minimum version compiled with Go 1.25.0, so it handles
these files correctly.

- Also fix 13 noctx violations

Fixes #

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] E2E tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #3191, opened 2026-02-27
🌱 Bump golang to v1.25 

kashifest:
Also bumping osv scanner since it is required by go 1.25 bump

Fixes #

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] E2E tests have been added, if necessary.
- [ ] Integration tests have been added, if necessary.

PR #3196, opened 2026-03-02
:seedling: Enable Kube-API-linter

peppi-lotta:
Enable Kube-API-linter to lint contents of api directory. I've disabled most of the linters and I will create separate PRs enabling each. Some fixes will be breaking or so large there is no point in having them in one PR. 

I've set similar rules as for linting as is set in CAPI project. Check CAPI config [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/.golangci-kal.yml)

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/kube-api-lint.yml
+++ b/.github/workflows/kube-api-lint.yml
@@ -1,0 +1,32 @@
+name: kube-api-lint
+
+on:
+  pull_request:
+    types: [ opened, edited, reopened, synchronize, ready_for_review ]
+
+permissions: {}
+
+jobs:
+  kubeapilint:
+    name: api-lint
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        working-directory:
+        - api
+
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+    - name: Calculate go version
+      id: vars
+      run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -35,3 +35,5 @@ jobs:
         version: v2.8.0
         working-directory: ${{matrix.working-directory}}
         args: --timeout=15m
+    - name: Run kube-api-linter
```
tuminoid: Can it and should it be separate job?

peppi-lotta:
/hold 
untill https://github.com/metal3-io/cluster-api-provider-metal3/pull/3191 and https://github.com/metal3-io/cluster-api-provider-metal3/pull/3193 are merged.

.github/workflows/golangci-lint.yml
peppi-lotta: This can be another job. I felt this is logical to have these together as kube-api-linter a golanci linter. What would be the reason to separate this into another job?

tuminoid: Golang linting vs api linting is two vastly different things, even if its implemented by a plugin to language linter.

tuminoid: So if its implicitly built into golangci-lint, fine, no need to build extra scaffolding, but here it is clearly a separate make target, so I'd think it would be beneficial to have it separately as well. Other opinions are welcome too.

peppi-lotta:
> /hold untill #3191 and #3193 are merged.

I thought it would be nice to get these merged before this PR but it's not necessary

/hold cancel

adilGhaffarDev:

PR #3216, opened 2026-03-11
🌱 Update e2e tests to use v1beta2

adilGhaffarDev:
Migrate the e2e test suite from the v1beta1 to the v1beta2 CAPM3 API, following the upstream CAPI v1beta2 migration. This also includes a backward-compatibility mechanism to still run tests against v1beta1 cluster templates, this can be done by setting `CAPM3_API_VERSION=v1beta1` (we will do it jjb in cicd repo).

##### E2E test suite (e2e)

- Replace all infrav1beta1 imports with infrav1 (pointing to v1beta2)
- Remove the duplicate `infrav1beta1.AddToScheme` registration; register only `infrav1.AddToScheme`
- Add `capm3APIVersion` global variable (defaults to `v1beta2`, overridable via `CAPM3_API_VERSION` env var) and a `flavorSuffix()` helper that appends `-v1beta1` to flavor names when `CAPM3_API_VERSION=v1beta1`
- Update all references to `Metal3Machine.Spec.ProviderID` from pointer dereference to direct string access (v1beta2 field change)
- Update all `Metal3Machine.Status.Ready` checks to `ptr.Deref(m3m.Status.Initialization.Provisioned, false)` (v1beta2 status change)

##### Cluster templates

- Add a new `main-v1beta1` template directory with a full set of v1beta1-compatible cluster templates (cluster, clusterclass, kubeadm configs, ippool, remediation, taints)
- `test/e2e/data/infrastructure-metal3/main` contains v1beta2 templates which will be used by default.
- Update main cluster templates and clusterclass manifests for v1beta2 compatibility
- For v1.10, v1.11, and v1.12 v1beta1 is hardcoded 

**Checklist:**


adilGhaffarDev:
@kashifest @Sunnatillo @peppi-lotta

test/e2e/node_deletion_remediation.go
```diff
@@ -88,7 +89,7 @@ func NodeRemediation(ctx context.Context, inputGetter func() NodeRemediationInpu
 
 	By("Creating a Metal3Remediation resource")
 	timeout := metav1.Duration{Duration: 30 * time.Minute}
```
peppi-lotta:
```suggestion

```

I think it would be better to remove the `timeout` because it doesn't seem to be used anywhere else accept on line 111

test/e2e/node_deletion_remediation.go
```diff
@@ -103,14 +104,14 @@ func NodeRemediation(ctx context.Context, inputGetter func() NodeRemediationInpu
...
-		Spec: infrav1beta1.Metal3RemediationSpec{
-			Strategy: &infrav1beta1.RemediationStrategy{
-				Type:       "Reboot",
-				RetryLimit: 1,
-				Timeout:    &timeout,
+		Spec: infrav1.Metal3RemediationSpec{
+			Strategy: &infrav1.RemediationStrategy{
+				Type:           "Reboot",
+				RetryLimit:     1,
+				TimeoutSeconds: ptr.To(int32(timeout.Seconds())),
```
peppi-lotta:
```suggestion
				TimeoutSeconds: &thirtyMinutes,
```

This is unnecessarily complicated. Don't use timeout or create a new variable like `thirtyMinutes := int32(1800)` and use that. This would be much cleaner.

adilGhaffarDev: done

test/e2e/node_deletion_remediation.go
adilGhaffarDev: done

test/e2e/common.go
```diff
@@ -404,18 +404,18 @@ func ListBareMetalHosts(ctx context.Context, c client.Client, opts ...client.Lis
...
 // Similar to kubectl get metal3machines.
 func ListMetal3Machines(ctx context.Context, c client.Client, opts ...client.ListOption) {
-	metal3Machines := infrav1beta1.Metal3MachineList{}
+	metal3Machines := infrav1.Metal3MachineList{}
 	Expect(c.List(ctx, &metal3Machines, opts...)).To(Succeed())
 
 	rows := make([][]string, len(metal3Machines.Items)+1)
 	// Add column names
-	rows[0] = []string{"Name:", "Ready:", "Provider ID:"}
+	rows[0] = []string{"Name:", "Provisioned:", "Provider ID:"}
```
kashifest, line 412: whats this change for?

test/e2e/common.go
```diff
@@ -503,11 +503,11 @@ func WaitForNumBmhInState(ctx context.Context, state bmov1alpha1.ProvisioningSta
 func WaitForNumMetal3MachinesReady(ctx context.Context, input WaitForNumInput) {
 	Logf("Waiting for %d Metal3Machines to be ready", input.Replicas)
 	Eventually(func(g Gomega) {
-		m3mList := infrav1beta1.Metal3MachineList{}
+		m3mList := infrav1.Metal3MachineList{}
 		g.Expect(input.Client.List(ctx, &m3mList, input.Options...)).To(Succeed())
 		numReady := 0
 		for _, m3m := range m3mList.Items {
-			if m3m.Status.Ready {
+			if ptr.Deref(m3m.Status.Initialization.Provisioned, false) {
```
kashifest, line 510: same

test/e2e/common.go
adilGhaffarDev: contract for infra machine is changed, its no longer ready, now its `metal3machine.Status.Initialization.Provisioned`

test/e2e/common.go
adilGhaffarDev: same answer as above, because of new contract.

adilGhaffarDev:
> So once this merges, we wont be testing v1beta1 API any more in main, is that what we want or do we need two tests now?

Yes we will not be using v1beta1 on main anymore, but we are testing upgrades from v1beta1 to v1beta2 in clusterctl upgrade tests.
If we want to keep testing v1beta1 too it will be lot of tests. One thing we can do is we can keep testing v1beta1 in dev-env tests.

adilGhaffarDev:
> > So once this merges, we wont be testing v1beta1 API any more in main, is that what we want or do we need two tests now?
> 
> Yes we will not be using v1beta1 on main anymore, but we are testing upgrades from v1beta1 to v1beta2 in clusterctl upgrade tests. If we want to keep testing v1beta1 too it will be lot of tests. One thing we can do is we can keep testing v1beta1 in dev-env tests.

I will add separate test for v1beta1.

adilGhaffarDev:
/hold
I need to make changes in jjbs

test/e2e/data/infrastructure-metal3/main/bases/clusterclass/clusterclass.yaml
```diff
@@ -43,7 +43,7 @@ spec:
     description: Overrides controlPlaneEndpoint data of Metal3ClusterTemplate used by the cluster
     definitions:
     - selector:
-        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+        apiVersion: infrastructure.cluster.x-k8s.io/${CAPM3_VERSION}
```
lentzi90: Wouldn't it be better to just hard code these to v1beta2? I see for the older releases we hard code to v1beta1. I am thinking that this will most likely fail if we try to use anything other than v1beta2 with this ClusterClass.

adilGhaffarDev: this is kept as variable because we want to keep testing v1beta1 until it is fully removed.
This will be set from the jjb, I have added test that will set this to v1beta1.

adilGhaffarDev:
jjb is merged test is running lets see
https://jenkins.nordix.org/view/Metal3/job/metal3-centos-e2e-integration-test-main/2512/console

test/e2e/data/infrastructure-metal3/main/bases/clusterclass/clusterclass.yaml
lentzi90: But how can it work with v1beta1? I mean we are planning to change some fields, right?

adilGhaffarDev: true test failed, we need separate yaml for v1beta1.
I need to figure how to test both v1beta1 and v1beta2 on main, need to do some trick in e2e config file.

lentzi90: I would suggest to keep it simple. Have separate "hard coded" manifests for the different APIs. In reality I would also say that v1beta1 tests are not worth much when the controllers run v1beta2. These tests will test the exact same thing as v1beta2, just going through the conversion webhook one time when applying the manifests.

adilGhaffarDev: done, now api version in yamls is hardcoded and I have separate folder for v1beta1.
v1beta2 is default and v1beta1 yamls will be used if we set CAPM3_API_VERSION=v1beta1, which we will set from jjb in separate v1beta1 tests.

PR #3230, opened 2026-03-17
:seedling: remove docker push stuff from Makefile

tuminoid:
We don't manually ever push to quay, we have Github actions for releasing and on each merge push as well. Cleaning this off to make multi-arch building and maintenance simpler.


tuminoid:
/hold
Syncing all PRs.

PR #3231, opened 2026-03-17
⚠️ Remove corev1 ObjectReference and use local struct

adilGhaffarDev:
Remove corev1 ObjectReference and use local struct Metal3ObjectRef

Some info regarding why we are removing this:

> ObjectReference contains enough information to let you inspect or modify the referred object. --- New uses of this type are discouraged because of difficulty describing its usage when embedded in APIs.
> 

>  - Ignored fields. It includes many fields which are not generally honored. For instance, ResourceVersion and FieldPath are both very rarely valid in actual usage.

>  - Invalid usage help. It is impossible to add specific help for individual usage. In most embedded usages, there are particular restrictions like, "must refer only to types A and B" or "UID not honored" or "name must be restricted". Those cannot be well described when embedded.
>   - Inconsistent validation. Because the usages are different, the validation rules are different by usage, which makes it hard for users to predict what will happen.
>   - The fields are both imprecise and overly precise. Kind is not a precise mapping to a URL. This can produce ambiguity during interpretation and require a REST mapping. In most cases, the dependency is on the group,resource tuple and the version of the actual struct is irrelevant.
>   - We cannot easily change it. Because this type is embedded in many locations, updates to this type will affect numerous schemas. Don't make new APIs embed an underspecified API type they do not control.
> 
> Instead of using this type, create a locally provided and used type that is well-focused on your reference. For example, ServiceReferences for admission registration: https://github.com/kubernetes/api/blob/release-1.17/admissionregistration/v1/types.go#L533 . +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object +structType=atomic 

https://pkg.go.dev/k8s.io/api@v0.34.6/core/v1#ObjectReference 

**Checklist:**


adilGhaffarDev:

@peppi-lotta @Sunnatillo

internal/webhooks/v1beta2/metal3data_webhook.go
```diff
@@ -116,14 +116,6 @@ func (webhook *Metal3Data) ValidateUpdate(_ context.Context, oldObj, newObj runt
 				"cannot be modified",
 			),
 		)
-	} else if newMetal3Data.Spec.Template.Kind != oldMetal3Data.Spec.Template.Kind {
```
peppi-lotta, line 119: Why is this Template.Kind check removed?

adilGhaffarDev: new struct that we are using in place of ObjectRefrence looks like this:

```
// Metal3ObjectRef is a reference to a Metal3 resource by name and namespace.
type Metal3ObjectRef struct {
	// name of the resource.
	// +optional
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
	Name string `json:"name,omitempty"`

	// namespace of the resource.
	// +optional
	Namespace string `json:"namespace,omitempty"`
}
```

we dont have kind anymore.

peppi-lotta: Oh right and `Kind` in not needed for `Metal3ObjectRef` because it's a well structured object.

adilGhaffarDev:
> Changes look fine to me, but could you add explanation why we are changing from corev1.ObjectReference to infrav1.Metal3ObjectRef. covev1.ObjectReference is much more sophisticated from a first glance.

added in the pr description.

api/v1beta1/conversion.go
```diff
@@ -192,6 +263,20 @@ func (src *Metal3DataClaim) ConvertTo(dstRaw conversion.Hub) error {
 		return err
 	}
 
+	if !reflect.DeepEqual(src.Spec.Template, corev1.ObjectReference{}) {
+		dst.Spec.Template = infrav1.Metal3ObjectRef{
+		Name:      src.Spec.Template.Name,
+			Namespace: src.Spec.Template.Namespace,
+		}
```
fmuyassarov: nit, indentation...

api/v1beta2/common_types.go
```diff
@@ -100,6 +100,27 @@ type CustomDeploy struct {
...
+}
+
+// IsDefined returns true if the Metal3ObjectRef is set.
+func (r *Metal3ObjectRef) IsDefined() bool {
+	if r == nil {
+		return false
+	}
+	return r.Name != "" && r.Namespace != ""
+}
+
```
fmuyassarov: Not a blocker, but I want to understand it better. Should we require name and namespace or can it be okay to skip or have an empty value for the namespace? Because, I can see from the tests below that we don't provide a namespace - but only the name and according to this - we would get `False` (?) in those cases.
```Go
DataTemplate: &infrav1.Metal3ObjectRef{Name: "abcd"},
```

fmuyassarov:
/hold
there are some open comments

api/v1beta1/conversion.go
```diff
@@ -672,3 +777,25 @@ func Convert_v1beta2_IPPoolReference_To_v1_TypedLocalObjectReference(in *infrav1
...
+
+// Convert_v1_ObjectReference_To_v1beta2_Metal3ObjectRef is a no-op stub;
+// actual conversion is performed in ConvertTo/ConvertFrom.
+func Convert_v1_ObjectReference_To_v1beta2_Metal3ObjectRef(_ *corev1.ObjectReference, _ *infrav1.Metal3ObjectRef, _ apimachineryconversion.Scope) error {
+	return nil
+}
+
+// Convert_v1beta2_Metal3ObjectRef_To_v1_ObjectReference is a no-op stub;
+// actual conversion is performed in ConvertTo/ConvertFrom.
+func Convert_v1beta2_Metal3ObjectRef_To_v1_ObjectReference(_ *infrav1.Metal3ObjectRef, _ *corev1.ObjectReference, _ apimachineryconversion.Scope) error {
```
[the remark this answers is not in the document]
adilGhaffarDev: removed conversion form ConvertTo/ConvertFrom and added here as recommended.

config/crd/bases/infrastructure.cluster.x-k8s.io_metal3machines.yaml
```diff
@@ -631,46 +631,15 @@ spec:
...
-                      Specific resourceVersion to which this reference is made, if any.
-                      More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency
-                    type: string
-                  uid:
-                    description: |-
-                      UID of the referent.
-                      More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
+                    description: namespace of the resource.
                     type: string
                 type: object
```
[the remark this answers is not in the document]
adilGhaffarDev: makes sense, added `// +structType=atomic` on top of `Metal3ObjectRef` struct

api/v1beta1/conversion.go
```diff
@@ -192,6 +263,20 @@ func (src *Metal3DataClaim) ConvertTo(dstRaw conversion.Hub) error {
 		return err
 	}
 
+	if !reflect.DeepEqual(src.Spec.Template, corev1.ObjectReference{}) {
+		dst.Spec.Template = infrav1.Metal3ObjectRef{
+		Name:      src.Spec.Template.Name,
```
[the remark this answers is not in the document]
adilGhaffarDev: fixed

api/v1beta1/conversion.go
```diff
@@ -422,6 +521,14 @@ func Convert_v1beta2_Metal3MachineStatus_To_v1beta1_Metal3MachineStatus(in *infr
 	}
 	out.V1Beta2 = &Metal3MachineV1Beta2Status{}
 	out.V1Beta2.Conditions = in.Conditions
+
+	if in.RenderedData.IsDefined() {
+		out.RenderedData = &corev1.ObjectReference{
+			Name:      in.RenderedData.Name,
+			Namespace: in.RenderedData.Namespace,
+		}
+	}
```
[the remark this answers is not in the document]
adilGhaffarDev: removed this function.

api/v1beta1/conversion.go
```diff
@@ -455,6 +562,11 @@ func Convert_v1beta1_Metal3MachineStatus_To_v1beta2_Metal3MachineStatus(in *Meta
 	}
 	out.Deprecated.V1Beta1.FailureReason = in.FailureReason
 	out.Deprecated.V1Beta1.FailureMessage = in.FailureMessage
+
+	if in.RenderedData != nil && !reflect.DeepEqual(in.RenderedData, &corev1.ObjectReference{}) {
```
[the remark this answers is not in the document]
adilGhaffarDev: fixed

api/v1beta2/common_types.go
adilGhaffarDev: is was used only at one, I have removed this function and doing only nil check where this function was previously.
Ref: 
```
	if in.RenderedData != nil {
		out.RenderedData = &corev1.ObjectReference{
			Name:      in.RenderedData.Name,
			Namespace: in.RenderedData.Namespace,
		}
	}
```
https://github.com/metal3-io/cluster-api-provider-metal3/pull/3231/changes#diff-010a3c3002f2104a2dc44482ef65502bfe78bac5eef87e8356d588966b919359R427

api/v1beta1/conversion.go
adilGhaffarDev: fixed

adilGhaffarDev:
pr also adding exclusion godot for kubebuilder markers
```
- godot
      source: '// \+'
```

adilGhaffarDev:
all comments addressed, unholding
/unhold

PR #3232, opened 2026-03-17
:seedling: Do not run go commands if GO is not set

terror96:

If GO env variable is not set, make will try to run invalid shell commands. Do not run the commands if the value is not set.

Fixes #

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,11 @@ SHELL:=/usr/bin/env bash
 GO_VERSION ?= 1.25.8
 GO := $(shell type -P go)
 # Use GOPROXY environment variable if set
-GOPROXY := $(shell $(GO) env GOPROXY)
+ifneq ($(GO),)
+	GOPROXY := $(shell $(GO) env GOPROXY)
+endif
 ifeq ($(GOPROXY),)
-GOPROXY := https://proxy.golang.org
+	GOPROXY := https://proxy.golang.org
 endif
 export GOPROXY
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,9 +47,11 @@ FAKE_APISERVER_DIR := hack/fake-apiserver
 MAKE_ROOT_DIR := $(CURDIR)
 
 # Set --output-base for conversion-gen if we are not within GOPATH
+ifneq ($(GO),)
 ifneq ($(abspath $(ROOT_DIR)),$(shell $(GO) env GOPATH)/src/github.com/metal3-io/cluster-api-provider-metal3)
 	CONVERSION_GEN_OUTPUT_BASE := --output-base=$(ROOT_DIR)
 endif
```
tuminoid, line 53: Nit: indent this.

Makefile
```diff
@@ -72,7 +76,9 @@ GINKGO_PKG := github.com/onsi/ginkgo/v2/ginkgo
 
 # Helper function to get dependency version from go.mod
 get_go_version = $(shell $(GO) list -m $1 | awk '{print $$2}')
-GINGKO_VER := $(call get_go_version,github.com/onsi/ginkgo/v2)
+ifneq ($(GO),)
+	GINGKO_VER := $(call get_go_version,github.com/onsi/ginkgo/v2)
```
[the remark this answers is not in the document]
terror96: Fixing this as part of this PR too.

Makefile
terror96: Actually. We found this:

"The only cases where it wouldn't work:

GNU Make < 3.82 (very old)
Non-GNU Make implementations (BSD make, etc.)"

Mac seems to use Make 3.81. Let's not indent.

tuminoid: OK, Mac's make 3.81 from 2006 does not support it. Ignore.

PR #3284, opened 2026-04-14
:bug: Set default providerid 

knfoo:

When the workload cluster API is not yet available (e.g. during initial
provisioning), SetProviderIDFromNodeLabel and SetNodeProviderIDByHostname
now fall through instead of returning transient errors that block the
reconciliation loop. This allows the controller to set a default
ProviderID via SetDefaultProviderID so that CAPI can proceed with its
lifecycle (e.g. populating Machine.Status.Addresses) before nodes join.

Fixes # https://github.com/metal3-io/cluster-api-provider-metal3/issues/3274

**Checklist:**


lentzi90 (commented):
This looks ok to me. I see it is still in draft mode. Do you expect to make more changes?

knfoo:
No, I am good, will move it to regular PR.

lentzi90 (commented):
/lgtm
/cc @adilGhaffarDev

Sunnatillo (commented):
LGTM
/cc @adilGhaffarDev

Sunnatillo (commented):
The changes look fine to me. 

Could you explain more about your use case. 
Lennart already mentioned about "Thanks for creating the issue! I think this is a valid use-case and something we have simply missed because we always use kubeadm-based bootstrap and control-plane providers."

But I want to know how are you using CAPM3, so you faced this issue, if possible of course

Sunnatillo (commented):
/

knfoo:
> The changes look fine to me.
> 
> Could you explain more about your use case. Lennart already mentioned about "Thanks for creating the issue! I think this is a valid use-case and something we have simply missed because we always use kubeadm-based bootstrap and control-plane providers."
> 
> But I want to know how are you using CAPM3, so you faced this issue, if possible of course

@Sunnatillo our use case is capi and having talos as control plane and bootstrap provider. 
https://github.com/metal3-io/cluster-api-provider-metal3/issues/3286 is related to this one.

PR #3293, opened 2026-04-17
🌱 Do not run code related linter tests PR title change

Sunnatillo:

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -2,7 +2,7 @@ name: golangci-lint
 
 on:
   pull_request:
-    types: [ opened, edited, reopened, synchronize, ready_for_review ]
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
This is intended to get them to rerun in case they fail. Otherwise a push is required.

Yes, its sometimes causing extra linter runs, but its just CI running for couple minutes, who cares, when it solves an issue for humans.

/hold

Sunnatillo:
> This is intended to get them to rerun in case they fail. Otherwise a push is required.
> 
> Yes, its sometimes causing extra linter runs, but its just CI running for couple minutes, who cares, when it solves an issue for humans.
> 
> /hold

I see. So we are using to rerun the tests this way, 

/close

PR #3317, opened 2026-04-22
🐛 Fix logical operator in ConsumerRef validation

lentzi90:

Changed AND to OR in the condition that validates BareMetalHost ConsumerRef Kind and Group. The controller should reject a ConsumerRef if either the Kind is not Metal3Machine OR the Group does not match infrav1, not only when both conditions are true.

Add test cases for both wrong Kind and wrong Group scenarios.

Fixes #

**Checklist:**


smoshiur1237 (commented):
/approve

Sunnatillo (commented):
nice catch
/lgtm

PR #3318, opened 2026-04-22
🌱 Switch from docker to moby dependencies

kashifest:
OSV scan is failing due to vulnerabilities in github.com/docker/docker. Bump to v29 is not trivial and wont fix the issue either. Following CAPI's lead on switching to moby instead https://github.com/kubernetes-sigs/cluster-api/pull/13545


smoshiur1237 (commented):
/approve

test/go.mod
```diff
@@ -191,7 +188,7 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/apiserver v0.35.4 // indirect
 	k8s.io/cli-runtime v0.35.4 // indirect
-	k8s.io/cluster-bootstrap v0.35.4 // indirect
+	k8s.io/cluster-bootstrap v0.35.3 // indirect
```
Sunnatillo: Hmm, is some dependency downgrading this?

lentzi90: Looks a bit strange actually. But this is just the test module anyway.

lentzi90 (commented):
Not sure we should scan the test module to begin with, but this switch does make sense.
/lgtm
Hold for discussing the downgrade that makes one k8s dependency out of sync with the rest
/hold

test/go.mod
kashifest: Yes, seems like it, I didn't downgrade it deliberately. I can manually put it back if thats expected although I don't usually touch indirect dependencies.

kashifest:
> Not sure we should scan the test module to begin with, but this switch does make sense.

I didn't get this. We only have this dependancy in the test module in CAPM3

test/go.mod
kashifest: Done

lentzi90 (commented):
/lgtm

lentzi90:
> > Not sure we should scan the test module to begin with, but this switch does make sense.
> 
> I didn't get this. We only have this dependancy in the test module in CAPM3
> 

I mean that we should not have OSV scanner care about test modules. If there is a vulnerability in the test code, it does not affect users. It is not part of the images and other artifacts we ship.

/hold cancel

kashifest:
> I mean that we should not have OSV scanner care about test modules. If there is a vulnerability in the test code, it does not affect users. It is not part of the images and other artifacts we ship.

Thats a valid point, we should silent the scanner for test modules. In that note, I don't think this needs to be backported.

PR #3326, opened 2026-04-23
:bug: Fix incorrect logical operator in ConsumerRef validation

kashifest:
Use `||` instead of `&&` when checking ConsumerRef Kind and Group, so a partial match is properly rejected.


kashifest:
it seems like this was fixed
/close

PR #3327, opened 2026-04-23
:bug: Restrict namespace secret references in Metal3Machine

kashifest:
Validate that userData, metaData, and networkData secret references do not point to namespaces other than the Metal3Machine's own namespace.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -389,7 +389,10 @@ func (m *MachineManager) Associate(ctx context.Context) (string, error) {
 
 	// A machine bootstrap not ready case is caught in the controller
 	// ReconcileNormal function
-	m.getUserDataSecretName(ctx)
+	err = m.getUserDataSecretName(ctx)
+	if err != nil {
+		return "", err
+	}
 
 	m.setHostLabel(ctx, host)
 
@@ -431,16 +434,20 @@ func (m *MachineManager) Associate(ctx context.Context) (string, error) {
 	return chosenHostReason, nil
 }
 
-// getUserDataSecretName gets the UserDataSecretName from the machine and exposes it as a secret
-// for the BareMetalHost through Metal3Machine. The UserDataSecretName might already be in a secret with
-// CABPK v0.3.0+, but if it is in a different namespace than the BareMetalHost,
-// then we need to create the secret.
-func (m *MachineManager) getUserDataSecretName(_ context.Context) {
+// getUserDataSecretName resolves the userData secret reference for the BareMetalHost.
+// If Metal3Machine.Spec.UserData is set, it is used directly but must reference the
+// same namespace as the Metal3Machine (cross-namespace references are rejected).
+// Otherwise, the bootstrap DataSecretName or ConfigRef from the CAPI Machine is used.
+func (m *MachineManager) getUserDataSecretName(_ context.Context) error {
 	if m.Metal3Machine.Status.UserData != nil {
-		return
+		return nil
 	}
 
 	if m.Metal3Machine.Spec.UserData != nil {
+		if m.Metal3Machine.Spec.UserData.Namespace != "" && m.Metal3Machine.Spec.UserData.Namespace != m.Metal3Machine.Namespace {
+			return fmt.Errorf("cross-namespace secret reference is not allowed for userData (namespace %q does not match %q)",
+				m.Metal3Machine.Spec.UserData.Namespace, m.Metal3Machine.Namespace)
+		}
 		m.Metal3Machine.Status.UserData = m.Metal3Machine.Spec.UserData
 	}
 
@@ -450,13 +457,14 @@ func (m *MachineManager) getUserDataSecretName(_ context.Context) {
 			Name:      *m.Machine.Spec.Bootstrap.DataSecretName,
 			Namespace: m.Machine.Namespace,
 		}
-		return
+		return nil
 	} else if m.Machine.Spec.Bootstrap.ConfigRef.IsDefined() {
 		m.Metal3Machine.Status.UserData = &corev1.SecretReference{
 			Name:      m.Machine.Spec.Bootstrap.ConfigRef.Name,
 			Namespace: m.Machine.Namespace,
 		}
 	}
+	return nil
 }
 
 // Delete deletes a metal3 machine and is invoked by the Machine Controller.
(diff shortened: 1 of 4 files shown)


internal/webhooks/v1beta2/metal3machine_webhook.go
```diff
@@ -58,8 +59,26 @@ func (webhook *Metal3Machine) validate(newM3M *infrav1.Metal3Machine) error {
 		allErrs = append(allErrs, newM3M.Spec.Image.Validate(*field.NewPath("Spec", "Image"))...)
 	}
 
+	allErrs = append(allErrs, validateSecretNamespace(newM3M.Spec.UserData, newM3M.Namespace, field.NewPath("spec", "userData"))...)
+	allErrs = append(allErrs, validateSecretNamespace(newM3M.Spec.MetaData, newM3M.Namespace, field.NewPath("spec", "metaData"))...)
+	allErrs = append(allErrs, validateSecretNamespace(newM3M.Spec.NetworkData, newM3M.Namespace, field.NewPath("spec", "networkData"))...)
```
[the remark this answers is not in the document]
kashifest: The existing v1beta2 webhooks predominantly use lowercase JSON names — metal3cluster_webhook.go uses `spec`, `controlPlaneEndpoint`, metal3remediation_webhook.go  uses `spec`, `strategy` etc which matches the dominant convention and aligns with what the Kubernetes API server returns in validation errors (JSON field names from json: struct tags).

internal/webhooks/v1beta2/metal3machine_webhook.go
```diff
@@ -58,8 +59,26 @@ func (webhook *Metal3Machine) validate(newM3M *infrav1.Metal3Machine) error {
...
+// validateSecretNamespace rejects secret references that point to a namespace
+// other than the Metal3Machine's own namespace, preventing cross-namespace
+// secret disclosure.
+func validateSecretNamespace(ref *corev1.SecretReference, m3mNamespace string, fldPath *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	if ref != nil && ref.Namespace != "" && ref.Namespace != m3mNamespace {
+		errs = append(errs, field.Forbidden(
+			fldPath.Child("namespace"),
+			"cross-namespace secret references are not allowed",
+		))
```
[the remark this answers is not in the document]
kashifest: This is also nit picking IMO, as mentioned in the previous comment the parent path is using lower case convention

baremetal/metal3machine_manager.go
```diff
@@ -435,12 +438,16 @@ func (m *MachineManager) Associate(ctx context.Context) (string, error) {
 // for the BareMetalHost through Metal3Machine. The UserDataSecretName might already be in a secret with
 // CABPK v0.3.0+, but if it is in a different namespace than the BareMetalHost,
 // then we need to create the secret.
-func (m *MachineManager) getUserDataSecretName(_ context.Context) {
+func (m *MachineManager) getUserDataSecretName(_ context.Context) error {
```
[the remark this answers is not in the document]
kashifest: Done

baremetal/metal3machine_manager_test.go
```diff
@@ -2798,7 +2804,7 @@ var _ = Describe("Metal3Machine manager", func() {
 			)
 			Expect(err).NotTo(HaveOccurred())
 
-			machineMgr.getUserDataSecretName(context.TODO())
+			Expect(machineMgr.getUserDataSecretName(context.TODO())).To(Succeed())
```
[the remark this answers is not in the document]
kashifest: Done

internal/webhooks/v1beta2/metal3machine_webhook_test.go
```diff
@@ -65,6 +66,21 @@ func TestMetal3MachineValidation(t *testing.T) {
...
+
+	crossNsNetworkData := valid.DeepCopy()
+	crossNsNetworkData.Spec.NetworkData = &corev1.SecretReference{Name: "secret", Namespace: "other-ns"}
+
+	sameNsUserData := valid.DeepCopy()
+	sameNsUserData.Spec.UserData = &corev1.SecretReference{Name: "secret", Namespace: "foo"}
+
+	noNsUserData := valid.DeepCopy()
+	noNsUserData.Spec.UserData = &corev1.SecretReference{Name: "secret"}
+
```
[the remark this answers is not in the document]
kashifest: Done

smoshiur1237 (commented):
/approve

baremetal/metal3machine_manager.go
```diff
@@ -431,16 +434,20 @@ func (m *MachineManager) Associate(ctx context.Context) (string, error) {
...
+// same namespace as the Metal3Machine (cross-namespace references are rejected).
+// Otherwise, the bootstrap DataSecretName or ConfigRef from the CAPI Machine is used.
+func (m *MachineManager) getUserDataSecretName(_ context.Context) error {
 	if m.Metal3Machine.Status.UserData != nil {
-		return
+		return nil
 	}
 
 	if m.Metal3Machine.Spec.UserData != nil {
+		if m.Metal3Machine.Spec.UserData.Namespace != "" && m.Metal3Machine.Spec.UserData.Namespace != m.Metal3Machine.Namespace {
```
Rozzii, line 447: TBH the proper fix IMO would be to allow only specifying a secret name not even a namespace, not even a secret reference, just a name and the controller would look up the name in the current namespace of the M3M.

The second best option would be to  make this a webhook validation for admission, so that user would get an immediate rejection and the object would not change. This is what you are doing correctly in this PR so I am not sure why this condition is needed. 

IMO handling this scenario here in the controller's "getter" is not needed if you are already handling it in the webhook.

baremetal/metal3machine_manager.go
```diff
@@ -1725,11 +1745,20 @@ func findOwnerRefFromList(refList []metav1.OwnerReference, objType metav1.TypeMe
 func (m *MachineManager) AssociateM3Metadata(ctx context.Context) error {
 	m.Log.V(VerbosityLevelTrace).Info("Associating M3Metadata with Metal3Machine",
 		LogFieldMetal3Machine, m.Metal3Machine.Name)
-	// If the secrets were provided by the user, use them.
+	// If the secrets were provided by the user, use them after enforcing
+	// same-namespace constraint to prevent cross-namespace secret disclosure.
 	if m.Metal3Machine.Spec.MetaData != nil {
+		if m.Metal3Machine.Spec.MetaData.Namespace != "" && m.Metal3Machine.Spec.MetaData.Namespace != m.Metal3Machine.Namespace {
```
Rozzii, line 1751: Sam comment as above the webhook should have already rejected any problematic secret by this point.

Rozzii (commented):
I think the check in the webhook is enough, if the controller changes were done to facilitate migration between versions I would say mark this as a breaking change. I am afraid that after a few releases these unnecessary checks will stay in place and keep doing unnecessary validation.

kashifest:
> I think the check in the webhook is enough, if the controller changes were done to facilitate migration between versions I would say mark this as a breaking change. I am afraid that after a few releases these unnecessary checks will stay in place and keep doing unnecessary validation.

so breaking change in a sense cross name space reference is not allowed anymore? I think the controller check is second line of defense, specially for existing objects , I don't know if any conversion hook will take care of it. I do buy your argument that its an extra validation but if its unnecessary, I am not sure if I agree still.

baremetal/metal3machine_manager.go
kashifest: > TBH the proper fix IMO would be to allow only specifying a secret name not even a namespace, not even a secret reference, just a name and the controller would look up the name in the current namespace of the M3M.
> 

Well the Namespace is there in the m3m spec, we just do a check if the namespace is not empty and if it matches m3m namespace, isn't that simpler than going through all the secret in the m3m namespace and match for the name?

adilGhaffarDev: I agree that there should not be namespace if we dont allow user to change it, it will always be m3m namespace. In CAPI its also just name: m.Machine.Spec.Bootstrap.DataSecretName , probably because of the same reason. 

But if its a big change, just webhook and this check also fine. Its better to have checks in both places webhook and controller. Webhook might be misconfigured or disabled.

kashifest: I will keep the machine_manager change unless @Rozzii feels strongly against it.

kashifest: @adilGhaffarDev changing it to secret name only is an API change and then we can't backport the fix as is now in the release branch, this way atleast we can

adilGhaffarDev (approved):

tuminoid (commented):
I'm fine with the fix itself, but I'd love to see tests that actually test `setHostSpec` and `AssociateM3Metadata` cannot access the secret, ie. negative test to prevent silent regressions later.

kashifest:
> I'm fine with the fix itself, but I'd love to see tests that actually test `setHostSpec` and `AssociateM3Metadata` cannot access the secret, ie. negative test to prevent silent regressions later.

`setHostSpec `and `getUserDataSecretName` are already covered, check the "User data has explicit alternate namespace" test cases for both the functions, `AssociateM3Metadata`  seems to be missing those.

kashifest:
>  `AssociateM3Metadata` seems to be missing those.

This is now done

tuminoid (commented):
/lgtm

Thanks @kashifest . After merging, please handle backporting as well.

kashifest:
I will do manual backporting

PR #3375, opened 2026-05-08
🌱 Fix missmatch of CAPI version in test

smoshiur1237:
There was a miss match of CAPI version in gomod file. This needs to be aligned before doing the minor release.


lentzi90 (commented):
/approve

PR #3379, opened 2026-05-12
:seedling: Do not delete bootstrap user-data secrets

terror96:
Do not remove secrets created by the bootstrap provider (KubeadmConfig). Rely on OwnerReferences and Kubernetes garbage collection to clean up.

Eliminate code and unit test for no longer used `deleteSecret()`.

We should not delete an object which is not created by capm3.

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -672,21 +672,6 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 
 		host.Spec.ConsumerRef = nil
 
-		// Delete created secret, if data was set without DataSecretName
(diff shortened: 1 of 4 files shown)


lentzi90:
The failed test is broken on main also: https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-centos-e2e-feature-test-main-scalability/

lentzi90 (commented):
The logic looks good to me. Tests are passing. I think it is better to merge this and find out if it breaks something, than to leave the code just in case.
/approve

adilGhaffarDev (approved):

adilGhaffarDev (commented):
/lgtm

PR #3420, opened 2026-05-26
🌱 Fix capm3, capi and ipam versions for main branch

adilGhaffarDev:

Fixes the CAPM3, CAPI, and IPAM versions. Updated the CAPM3 and IPAM versions to 14.99, which points to local resources, and updated the CAPI version to the latest stable 1.13 release instead of the RC version.

Fixes #

**Checklist:**

### DIFF
--- a/clusterctl-settings.json
+++ b/clusterctl-settings.json
@@ -2,6 +2,6 @@
   "name": "infrastructure-metal3",
   "config": {
     "componentsFile": "infrastructure-components.yaml",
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/approve
Could you add something in the description also? Looks a bit strange when it is empty.

adilGhaffarDev:
> /approve
> Could you add something in the description also? Looks a bit strange when it is empty.

Done

PR #3439, opened 2026-06-05
🌱 Set cooldown days to 3 for dependabot

migiFi:
Changed the dependabot cooldown days from 7 to 3 and removed zizmor: ignored from file.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,13 +20,13 @@ updates:
   - dependency-name: "*"
     update-types: [ "version-update:semver-major" ]
   cooldown:
-    default-days: 7
+    default-days: 3
   commit-message:
     prefix: ":seedling:"
   labels:
   - "ok-to-test"
   # Go
-- package-ecosystem: "gomod" # zizmor: ignore[dependabot-cooldown]
+- package-ecosystem: "gomod"
   directories:
   - "/"
   - "/api"
@@ -57,6 +57,8 @@ updates:
   - dependency-name: "github.com/metal3-io/ip-address-manager/api"
   commit-message:
     prefix: ":seedling:"
+  cooldown:
+    default-days: 3
   labels:
   - "ok-to-test"
 ## main branch config ends here
@@ -78,13 +80,13 @@ updates:
   - dependency-name: "*"
     update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
   cooldown:
-    default-days: 7
+    default-days: 3
   commit-message:
     prefix: ":seedling:"
   labels:
   - "ok-to-test"
   # Go
-- package-ecosystem: "gomod" # zizmor: ignore[dependabot-cooldown]
+- package-ecosystem: "gomod"
   directories:
   - "/"
   - "/api"
@@ -113,6 +115,8 @@ updates:
     update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
   commit-message:
     prefix: ":seedling:"
+  cooldown:
+    default-days: 3
   labels:
   - "ok-to-test"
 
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -57,6 +57,8 @@ updates:
   - dependency-name: "github.com/metal3-io/ip-address-manager/api"
   commit-message:
     prefix: ":seedling:"
+  cooldown:
+    default-days: 7
```
tuminoid: Community consensus is 3 days, not 7.

Delaying too much exposes you to hacks if something is found and fixed in them, while updating too fast exposes you to supply chain.

tuminoid (commented):
That title is horrendous :) 

Also, lets use 3 days, not 7.

/hold

.github/dependabot.yml
migiFi: ok I will make the changes to 3, should we do the same for CAPO ?

migiFi: before making the changes, are we only changing gomod to 3 and leaving gh-actions to 7? or all of them to 3?

tuminoid: Both of them to 3 days in Metal3 repos. For CAPO, @lentzi90 as CAPO maintainer can answer that. They don't need to be same necessarily.

migiFi: It seems like zizmor is not allowing a config with less than 7 days as cooldown. This is coming from the zizmor rule config upstream. I will change it back to 7 for now. 

@tuminoid could we make the change the zizmor rule from 7 to 3? 

```yaml
178 |     default-days: 3
    |     ^^^^^^^^^^^^^^^ insufficient default-days configured (less than 7)
```

Sunnatillo:
/retitle 🌱 Set cooldown days for Zizmor

.github/dependabot.yml
tuminoid: There is probably zizmor ignore for this, the minimum cooldown 7 is likely hardcoded into their checks.

migiFi:
Added `zizmor.yml` at the repo root to override zizmor's default `cooldown default-days` from 7 to 3. The config is explicitly referenced in `.github/workflows/zizmor.yml` via the `config: input`.

lentzi90 (commented):
/lgtm

lentzi90:
Please squash the commits also

lentzi90:
Could you fix the commit message? It still has the original description

smoshiur1237 (commented):
Please fix the commit message. otherwise
/approve

.github/workflows/zizmor.yml
```diff
@@ -28,9 +28,12 @@ jobs:
     - name: Run zizmor (SARIF)
       if: github.event_name == 'push'
       uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
+      with:
+        config: zizmor.yml
```
tuminoid: Name it `.zizmor.yml` so it doesn't pollute the root dir.

tuminoid (commented):
Just rename the config file, since it is configurable to not pollute repo root more than is needed, and then this is fine, getting rid of some ignores and setting 3 days (in addition to our "it runs once a week" delay).

lentzi90 (commented):
Thank you!
/lgtm

lentzi90:
/override metal3-ubuntu-e2e-integration-test-main metal3-ubuntu-e2e-integration-test-main
Overriding irrelevant tests

lentzi90:
Ah not possible to do that for multiple on the same line it seems...
/override metal3-ubuntu-e2e-integration-test-main

lentzi90:
Err... my bad.

PR #3454, opened 2026-06-10
🌱 Removed zizmor inline ignore

migiFi:
Removed the `zizmor: ignore[artipacked]` inline suppression comment from `.github/workflows/release.yaml`. The zizmor warning is handled by the `.zyzmor.yml` config at the repo root instead.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,7 +20,7 @@ jobs:
     if: github.repository == 'metal3-io/cluster-api-provider-metal3'
     steps:
     - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 # zizmor: ignore[artipacked]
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -20,9 +20,10 @@ jobs:
     if: github.repository == 'metal3-io/cluster-api-provider-metal3'
     steps:
     - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 # zizmor: ignore[artipacked]
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         fetch-depth: 0
+        persist-credentials: false
```
tuminoid: This is wrong fix, the credentials are needed later in the workflow.

tuminoid (commented):
This is wrong fix, the credentials are needed later in the workflow.

/hold

smoshiur1237 (commented):
/approve

tuminoid (commented):
/unhold

PR #3455, opened 2026-06-10
🌱 Removed zizmor inline ignore

migiFi:
Removed the `zizmor: ignore[artipacked]` inline suppression comment from `.github/workflows/release.yaml`. The zizmor warning is handled by the `.zyzmor.yml` config at the repo root instead.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,13 +20,13 @@ updates:
   - dependency-name: "*"
     update-types: [ "version-update:semver-major" ]
   cooldown:
-    default-days: 7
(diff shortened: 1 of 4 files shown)


tuminoid (commented):
I don't see zizmor.yaml config change though?

migiFi:
Added the `.zizmor.yml` config file, 3 day cooldown and `ignore: artipacked` in place.

Sunnatillo (commented):
LGTM

tuminoid (commented):
/approve

PR #3458, opened 2026-06-13
✨ feat: add configurable BareMetalHost allocation policy

synthe102:

Add an opt-in `bareMetalHostAllocationPolicy` field to Metal3MachineSpec (v1beta2) allowing BareMetalHosts to be allocated either at random (the default, preserving existing behavior) or in alphabetical order by host name ("ordered").

Previously `pickHost` always selected an available host using crypto/rand, so a MachineDeployment scaling up filled its pool non-deterministically. With `ordered`, the available host whose name sorts first is chosen, including within a failure domain.

The field is set on the Metal3MachineTemplate at
`spec.template.spec.bareMetalHostAllocationPolicy` and inherited by every generated Metal3Machine. As it is hub-only, it is restored on v1beta1 conversion. Generated deepcopy is unaffected (plain string field).

Fixes #

**Checklist:**

### DIFF
--- a/api/v1beta2/metal3machine_types.go
+++ b/api/v1beta2/metal3machine_types.go
@@ -31,6 +31,10 @@ const (
 	CleaningModeMetadata = "metadata"
 	ClonedFromGroupKind  = "Metal3MachineTemplate.infrastructure.cluster.x-k8s.io"
 	LiveIsoDiskFormat    = "live-iso"
+
+	// AllocationPolicyOrdered selects the available BareMetalHost whose name sorts
+	// first alphabetically.
+	AllocationPolicyOrdered = "ordered"
 )
 
 // Metal3Machine's Ready condition and corresponding reasons that will be used in v1Beta2 API version.
@@ -176,6 +180,14 @@ type Metal3MachineSpec struct {
 	// +kubebuilder:validation:Enum:=metadata;disabled
 	// +optional
 	AutomatedCleaningMode string `json:"automatedCleaningMode,omitempty,omitzero"`
+
+	// bareMetalHostAllocationPolicy controls how a BareMetalHost is selected from the
+	// set of available hosts matching the hostSelector. "random" (the default) picks a
+	// host at random; "ordered" picks the available host whose name sorts first
+	// alphabetically.
+	// +kubebuilder:validation:Enum:=random;ordered
+	// +optional
+	BareMetalHostAllocationPolicy string `json:"bareMetalHostAllocationPolicy,omitempty,omitzero"`
 
 	// failureDomain is the failure domain unique identifier this machine should be attached to, as defined in Cluster API.
 	// +optional
--- a/api/v1beta1/conversion.go
+++ b/api/v1beta1/conversion.go
@@ -111,6 +111,10 @@ func (src *Metal3Machine) ConvertTo(dstRaw conversion.Hub) error {
 	if ok && restored.Spec.Image.Checksum == nil {
 		dst.Spec.Image.Checksum = nil
 	}
+	// Restore hub-only BareMetalHostAllocationPolicy (no equivalent in v1beta1).
+	if ok {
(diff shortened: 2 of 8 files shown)


api/v1beta2/metal3machine_types.go
```diff
@@ -31,6 +31,13 @@ const (
 	CleaningModeMetadata = "metadata"
 	ClonedFromGroupKind  = "Metal3MachineTemplate.infrastructure.cluster.x-k8s.io"
 	LiveIsoDiskFormat    = "live-iso"
+
+	// AllocationPolicyRandom selects a BareMetalHost at random from the available
+	// hosts matching the hostSelector. This is the default behavior.
+	AllocationPolicyRandom = "random"
```
fmuyassarov: seems a dead code?

fmuyassarov (commented):
Thanks for the changes. Could you please share what is the use case where you get a benefit from sorted hosts?

api/v1beta2/metal3machine_types.go
synthe102: Indeed, I'll remove.

synthe102:
> Thanks for the changes. Could you please share what is the use case where you get a benefit from sorted hosts?

BMH naming reflects network topology on the systems I manage. BMH close together on the network have the same prefix with an added prefix (eg: `foo-1`, `foo-2`, `bar-1`, `bar-2`).
Sorted hosts would make similar BMH provision first, enabling my end users to leverage best case networking during large clusters bringup (which can take several hours) and start running their applications while other BMH are still provisioning.

Without this, there's no guarantee that our HPC jobs can be scheduled since they are dependent on the network topology, even if we have enough random nodes provisioned.

fmuyassarov:
> > Thanks for the changes. Could you please share what is the use case where you get a benefit from sorted hosts?
> 
> BMH naming reflects network topology on the systems I manage. BMH close together on the network have the same prefix with an added prefix (eg: `foo-1`, `foo-2`, `bar-1`, `bar-2`). Sorted hosts would make similar BMH provision first, enabling my end users to leverage best case networking during large clusters bringup (which can take several hours) and start running their applications while other BMH are still provisioning.
> 
> Without this, there's no guarantee that our HPC jobs can be scheduled since they are dependent on the network topology, even if we have enough random nodes provisioned.

Okay, so the idea is to pack nodes into network adjacent groups first, which makes sense.

fmuyassarov:
I was wondering whether you have considered using annotation similar to the pause in BMO. It seems like a simpler approach and might be sufficient. We would lose schema validation, but that doesn’t strike me as a major issue.

synthe102:
That's an option indeed. Lets discuss it during the next community meeting !

dtantsur:
For the sake of preserving history, this was my comment on slack:

> you need to be aware of https://github.com/metal3-io/metal3-docs/blob/main/design/hostclaim-multitenancy.md happening, which is going to change where the allocation logic even lives. Given that, it could be helpful to start with a design proposal to iron out details.

fmuyassarov:
@synthe102 Do you plane to move with the design proposal on the metal3-docs first? If so, should we close this for now?

synthe102:
> @synthe102 Do you plane to move with the design proposal on the metal3-docs first? If so, should we close this for now?

Yes, I'm still working on it. I'm closing this PR.

Sunnatillo:
> > Thanks for the changes. Could you please share what is the use case where you get a benefit from sorted hosts?
> 
> BMH naming reflects network topology on the systems I manage. BMH close together on the network have the same prefix with an added prefix (eg: `foo-1`, `foo-2`, `bar-1`, `bar-2`). Sorted hosts would make similar BMH provision first, enabling my end users to leverage best case networking during large clusters bringup (which can take several hours) and start running their applications while other BMH are still provisioning.
> 
> Without this, there's no guarantee that our HPC jobs can be scheduled since they are dependent on the network topology, even if we have enough random nodes provisioned.

I like the idea, looking forward to see a proposal soon

PR #3473, opened 2026-06-19
:bug: Node reuse: transient error treatment

### LINKED ISSUE #3385: Node reuse and roles when control-plane is not kubeadm
Updating a cluster that does not use kubeadm as control-plane may fail especially if node-reuse is set.

Following https://github.com/metal3-io/cluster-api-provider-metal3/pull/3168, it should be explicitly stated that for deployments of capm3 associated to a control-plane provider that is not kubeadm, the role of the capm3 controller must be modified to add read rights over the control plane resource. 

In most case it "works" because, the fact that the search for the metal3machinetemplate fails is considered as normal. The check for error does not distinguish IsNotFound from real errors https://github.com/metal3-io/cluster-api-provider-metal3/blob/a8a8b6041faf2fefc1fa52015323984597f39c3f/baremetal/metal3machine_manager.go#L626
This may have unintended consequences in case of transient errors.

The only fix with the current algorithm is to modify the roles. Only the end-user can do that but he must be warned in metal3 documentation. I would also prefer that the problem is not silently hidden and getMetal3MachineTemplate correctly distinguish errors from the resource not being accessible as it may make node-reuse very brittle in production.

**Environment:**

- Cluster-api version:
- Cluster-api-provider-metal3 version: only on main / 1.13-rc
- Environment (metal3-dev-env or other):
- Kubernetes version: (use `kubectl version`):

/kind bug

pierrecregut:
While looking for Metal3 templates, there are two kinds of potential problems:

* A transient error occurred while fetching a resource
* A resource no longer exists

In the second case, it is reasonable to consider that there is no node reuse, but in the first case, a retry must be done.

Fixes: #3385

### DIFF
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -710,19 +710,16 @@ func (m *MachineManager) Delete(ctx context.Context) error {
 				var m3mt *infrav1.Metal3MachineTemplate
 				m3mt, err = m.getMetal3MachineTemplate(ctx)
 				if err != nil {
+					return fmt.Errorf("cannot get Metal3MachineTemplate: %w", err)
+				}
+				if m3mt == nil {
 					// we are here, because while normal deprovisioning, Metal3MachineTemplate will be deleted first
 					// and we can't get it even though Metal3Machine has reference to it. We consider it nil and move
 					// forward with normal deprovisioning.
-					m3mt = nil
 					m.Log.Info("Metal3MachineTemplate associated with Metal3Machine is deleted",
-						LogFieldHost, host.Name,
-						LogFieldError, err.Error())
+						LogFieldHost, host.Name)
 				} else {
-					// in case of upgrading, Metal3MachineTemplate will not be deleted and we can fetch it,
-					// in order to check for node reuse feature in the next step.
 					m.Log.Info("Found Metal3machineTemplate", "metal3machineTemplate", m3mt.Name)
-				}
-				if m3mt != nil {
 					if ptr.Deref(m3mt.Spec.NodeReuse, false) {
 						if host.Labels == nil {
 							host.Labels = make(map[string]string)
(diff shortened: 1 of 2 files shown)


baremetal/metal3machine_manager_test.go
```diff
@@ -4551,7 +4553,7 @@ var _ = Describe("Metal3Machine manager", func() {
 			Machine:            nil,
 			expectedMachineSet: nil,
 			MachineSetList:     nil,
-			expectError:        true,
+			expectError:        false,
```
Rozzii, line 5136: Please align the Entry test with the actual test expectations.

baremetal/metal3machine_manager.go
```diff
@@ -1947,85 +1944,102 @@ func (m *MachineManager) getControlPlaneName(_ context.Context) (string, error)
...
-		return "md-" + msOwnerRef.Name, nil
+		key := client.ObjectKey{Name: mOwnerRef.Name, Namespace: m.Machine.Namespace}
+		mSet := &clusterv1.MachineSet{}
+		if err := m.client.Get(ctx, key, mSet); err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil, nil //nolint:nilnil
+			}
+			return nil, err
+		}
+		return mSet, nil
```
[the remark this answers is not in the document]
Rozzii: Please check this, I am also concerned about this issue, this could cause a security issue.

baremetal/metal3machine_manager_test.go
```diff
@@ -4564,7 +4566,7 @@ var _ = Describe("Metal3Machine manager", func() {
 			},
 			expectedMachineSet: nil,
```
Rozzii, line 5147: Consider adding unit tests or rather fixture based integration tests that would check these scenarios:
- whether deletion proceeds
- whether a node-reuse label is or is not written
- whether transient errors are returned

Rozzii (commented):
Me and the AI have a few comments but in general I think this PR is going in the right direction.

baremetal/metal3machine_manager.go
```diff
@@ -1947,85 +1944,102 @@ func (m *MachineManager) getControlPlaneName(_ context.Context) (string, error)
...
-			machineSetError = fmt.Sprintf("MachineSet %s does not match Machine %s owner reference %s", machineset.Name, m.Machine.Name, mOwnerRef.Name)
+	for _, mOwnerRef := range m.Machine.ObjectMeta.OwnerReferences {
+		aGV, err := schema.ParseGroupVersion(mOwnerRef.APIVersion)
+		if err != nil {
+			m.Log.Error(err, "Found an unparsable APIVersion in owners", "apiVersion", mOwnerRef.APIVersion)
+			continue
+		}
+		if aGV.Group != controlplanev1.GroupVersion.Group {
+			continue
+		}
```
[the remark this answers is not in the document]
pierrecregut: Short answer: no. We are not fetching object whose type we do not know relying on rbac to prevent it... Most providers are in cluster.x-k8s.io anyway even if it is not a strict requirement. Another approach is to get the APIGroup from the Cluster (controlPlaneRef.apiGroup).

baremetal/metal3machine_manager_test.go
pierrecregut: I modified Delete test to add transient failures. Control-plane test has been fixed (only the worker example was fixed in the previous PR). I made the logic around testing errors similar to other tests in the same file (expectError + expectRequeue rather than expectedResult which was misleading as it was in fact equivalent to expectRequeue: true).

Rozzii (commented):
/lgtm

smoshiur1237 (commented):
@Sunnatillo @adilGhaffarDev @lentzi90 @Rozzii 
Code freeze for Release-1.14 will start at the end of this week. Could we take this in beforehand

baremetal/metal3machine_manager_test.go
```diff
@@ -2279,13 +2278,16 @@ var _ = Describe("Metal3Machine manager", func() {
...
+		Host                *bmov1alpha1.BareMetalHost
+		Secret              *corev1.Secret
+		Machine             *clusterv1.Machine
+		M3Machine           *infrav1.Metal3Machine
+		ControlPlane        *controlplanev1.KubeadmControlPlane
+		BMCSecret           *corev1.Secret
+		ExpectedConsumerRef *corev1.ObjectReference
+		ExpectError         bool
+		ExpectRequeue       bool
+		//		ExpectedResult                  error
```
lentzi90, line 2290: nit: commented code could be removed.

baremetal/metal3machine_manager_test.go
```diff
@@ -2510,8 +2524,9 @@ var _ = Describe("Metal3Machine manager", func() {
 				m3mObjectMetaWithValidAnnotations(),
 			),
 			ExpectedConsumerRef: consumerRef(),
-			ExpectedResult:      ReconcileError{nil, TransientErrorType, time.Second * 30},
-			Secret:              newSecret(),
+			ExpectRequeue:       true,
+			// ExpectedResult:      ReconcileError{nil, TransientErrorType, time.Second * 30},
```
lentzi90, line 2528: nit: commented code

lentzi90 (commented):
I'm going to trigger some more tests on this, and putting hold for them.
Two nits below, not going to block on them.
/approve
/hold
/test ?

lentzi90:
[NodeReuse is part of pivoting](https://github.com/metal3-io/cluster-api-provider-metal3/blob/71ac84b029cdf4b03e74a35d974223b4b79dd930/test/e2e/pivoting_based_feature_test.go#L129), so focusing on those tests.

Issue #1358, opened 2023-12-12
Unnecessary owner references


lentzi90:
**What steps did you take and what happened:**

1. Create a Cluster using Metal3DataTemplate and IPPool
2. Check the owner references on the IPAddress, IPClaim, Metal3Data, Metal3DataClaim and BareMetalHost
3. The Metal3Data owns the IPAddress and IPClaim (and the IPClaim owns the IPAddress). The Metal3Machine owns the BareMetalHost, Metal3DataClaim and Metal3Data (and the Metal3DataClaim owns the Metal3Data).

This owner structure seems quite strange and unnecessary. Why should the Metal3Machine own the Metal3Data for example when it already owns the Metal3DataClaim? And why should it own the BareMetalHost that is most likely created by the user?

These relations makes it harder to reason about behaviors and understanding the structure.
The owner references are usually used to delete owned objects when the owner is deleted. Owning both the parent and child, like is the case with the IPClaim and IPAddress, is not necessary and can lead to surprising behaviors.

**What did you expect to happen:**

The owner chain should not have situations where the grand parent owns both parent and child.
We should not put owner references on objects lightly. In general they belong on objects created from or managed by the owner, that should be deleted when the owner is deleted.

**Anything else you would like to add:**

There may be reasons for the structure that we have, but I have not been able to find them.
The main goal with this issue is to think critically about the structure we have, and think about if/how we can improve it.

**Environment:**

- Cluster-api version: v1.5.3
- Cluster-api-provider-metal3 version: v1.5.2
- Environment (metal3-dev-env or other): other
- Kubernetes version: (use `kubectl version`): v1.28.3

/kind bug

Rozzii:
I agree with you @lentzi90 as we have discussed this off GitHub, I will bring this up on triaging session on Wednesday, I would like a wide discussion over this involving as many of the contributors as possible.

zaneb:
Not an expert on CAPM3, but in general there's 2 motivations:

1. Ensure stuff gets deleted when the thing owning is deleted
2. Trigger reconciliation of the owner object when the owned object is modified

(Ideally these line up, but they don't always.)

> The Metal3Data owns the IPAddress and IPClaim (and the IPClaim owns the IPAddress).

Depending on what the IPAddress represents, this sounds potentially suspect.

> And why should \[the Metal3Machine\] own the BareMetalHost that is most likely created by the user?

[...]

PR #3520, opened 2026-07-01
⚠️ Deprecate Metal3DataTemplate spec.clusterName

adilGhaffarDev:
Historically a `Metal3DataTemplate` was tied to one `Cluster` via `spec.clusterName`, and the controller set a `Cluster` owner reference on it. This PR deprecates that field and decouples the template from any single Cluster, so one template can be shared across multiple clusters. Removing the coupling exposed a few behaviors that assumed a single owning Cluster, which this PR also fixes.

### Changes

1. **Deprecate `spec.clusterName`.** A validating webhook warns when the field is set. The controller no longer derives an owning Cluster from it or sets a Cluster owner reference on the template.
2. **Object-level pause.** Uses the pause annotation set directly on the template (`HasPaused(template)`) instead of the cluster-derived `IsPaused(cluster, obj)`, since a template may belong to many clusters.
3. **Secret(metaData and networkData) cluster-name label.** Derives the cluster name from the consuming machine (`machine.Spec.ClusterName`) instead of the template's own label.
4. **`Metal3Data` GC leak.** With the template no longer Cluster-owned, its leftover owner reference kept `Metal3Data` alive after cluster deletion. Removed it and made `Metal3DataClaim` the controlling owner, so `Metal3Data` is now in the Cluster's GC chain.
5. **Pivot (`clusterctl move`).** Added the `clusterctl.cluster.x-k8s.io/move` label to the CRD so the template is relocated during pivot even without a Cluster owner reference.
6. **Index allocation.** Made template-scoped, so a shared template hands out non-colliding indexes across clusters.
7. **E2E cleanup.** Teardown no longer waits for `Metal3DataTemplate` deletion on cluster removal, since it's intentionally no longer GC'd with the Cluster.

### Tests

**Unit** (baremetal, webhooks, controllers):
- **Shared-template index allocation** — a template already holding indexes `0` and `1` from two clusters gives a third cluster's claim index `2`, proving collision-free sharing.
- **`Metal3Data` owner references** — asserts the two-owner layout (controlling `Metal3DataClaim` + `Metal3Machine`, no `Metal3DataTemplate`), locking in the GC fix.
- **Webhook deprecation warning** — warns when `spec.clusterName` is set, not when empty.
- **Controller pause** — reconciliation is skipped via the template's own pause annotation.

**E2E** (e2e):
- **Pivot** — verifies the template is moved to the target cluster via the new move label.
- **Cleanup** — waits for `Metal3Data`/`Metal3DataClaim` GC on cluster deletion, no longer for `Metal3DataTemplate`.

Part of https://github.com/metal3-io/cluster-api-provider-metal3/issues/3521
Also related to https://github.com/metal3-io/cluster-api-provider-metal3/issues/1358

**Checklist:**

### DIFF
--- a/api/v1beta2/metal3datatemplate_types.go
+++ b/api/v1beta2/metal3datatemplate_types.go
@@ -888,8 +888,10 @@ type NetworkData struct {
 // Metal3DataTemplateSpec defines the desired state of Metal3DataTemplate.
 type Metal3DataTemplateSpec struct {
 	// clusterName is the name of the Cluster this object belongs to.
-	// +required
-	// +kubebuilder:validation:MinLength=1
+	//
+	// Deprecated: This field is deprecated and will be removed in a future release.
+	// It is no longer used by the controllers.
+	// +optional
 	// +kubebuilder:validation:MaxLength=512
 	ClusterName string `json:"clusterName,omitempty"`
 
@@ -923,6 +925,7 @@ type Metal3DataTemplateStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:subresource:status
 // +kubebuilder:object:root=true
+// +kubebuilder:metadata:labels="clusterctl.cluster.x-k8s.io/move="
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels.cluster\\.x-k8s\\.io/cluster-name",description="Cluster to which this template belongs"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Time duration since creation of Metal3DataTemplate"
 
--- a/baremetal/metal3datatemplate_manager.go
+++ b/baremetal/metal3datatemplate_manager.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
-	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -41,7 +40,6 @@ import (
 type DataTemplateManagerInterface interface {
 	SetFinalizer()
 	UnsetFinalizer()
-	SetClusterOwnerRef(*clusterv1.Cluster) error
 	// UpdateDatas handles the Metal3DataClaims and creates or deletes Metal3Data accordingly.
 	// It returns if there are still Data object and undeleted DataClaims objects.
 	UpdateDatas(context.Context) (bool, bool, error)
@@ -78,34 +76,6 @@ func (m *DataTemplateManager) UnsetFinalizer() {
 	// Remove the finalizer.
 	m.Log.V(VerbosityLevelTrace).Info("Removing finalizer from Metal3DataTemplate")
 	controllerutil.RemoveFinalizer(m.DataTemplate, infrav1.DataTemplateFinalizer)
-}
-
-// SetClusterOwnerRef sets ownerRef.
-func (m *DataTemplateManager) SetClusterOwnerRef(cluster *clusterv1.Cluster) error {
-	m.Log.V(VerbosityLevelTrace).Info("Setting cluster owner reference on Metal3DataTemplate")
-	// Verify that the owner reference is there, if not add it and update object,
-	// if error requeue.
-	if cluster == nil {
-		return errors.New("missing cluster")
-	}
-	_, err := findOwnerRefFromList(m.DataTemplate.OwnerReferences,
-		cluster.TypeMeta, cluster.ObjectMeta)
-	if err != nil {
-		if ok := errors.As(err, &errNotFound); !ok {
-			return err
-		}
-		m.Log.V(VerbosityLevelDebug).Info("Adding cluster owner reference",
-			LogFieldMetal3DataTemplate, m.DataTemplate.Name,
-			LogFieldCluster, cluster.Name)
-		m.DataTemplate.OwnerReferences, err = setOwnerRefInList(
-			m.DataTemplate.OwnerReferences, false, cluster.TypeMeta,
-			cluster.ObjectMeta,
-		)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
 }
 
 // RecreateStatus recreates the status if empty.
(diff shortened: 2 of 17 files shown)


fmuyassarov (commented):
LGTM. 

How about this though: https://github.com/metal3-io/cluster-api-provider-metal3/blob/d30ed2b62a13ffd7fecef529c4c3d33dbffe3bd4/test/e2e/remediation.go#L266

adilGhaffarDev:
> LGTM.
> 
> How about this though:
> 
> https://github.com/metal3-io/cluster-api-provider-metal3/blob/d30ed2b62a13ffd7fecef529c4c3d33dbffe3bd4/test/e2e/remediation.go#L266

good catch, I have removed it

fmuyassarov (approved):
/approve

fmuyassarov:
@adilGhaffarDev 
I wonder if we could/should actually add admission webhook warning when `clusterName` is set, since the field is now deprecated and unused by controllers. Just wondering if this could act as a better signal to the end user.

adilGhaffarDev:
> @adilGhaffarDev 
> 
> I wonder if we could/should actually add admission webhook warning when `clusterName` is set, since the field is now deprecated and unused by controllers. Just wondering if this could act as a better signal to the end user.
> 
> 

Yes i agree, I will do that

adilGhaffarDev:
/hold
there are some more changes needed.

api/v1beta2/metal3datatemplate_types.go
```diff
@@ -888,8 +888,10 @@ type NetworkData struct {
 // Metal3DataTemplateSpec defines the desired state of Metal3DataTemplate.
 type Metal3DataTemplateSpec struct {
 	// clusterName is the name of the Cluster this object belongs to.
-	// +required
-	// +kubebuilder:validation:MinLength=1
+	//
+	// Deprecated: This field is deprecated and will be removed in a future release.
+	// It is no longer used by the controllers.
+	// +optional
 	// +kubebuilder:validation:MaxLength=512
```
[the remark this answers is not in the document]
adilGhaffarDev: fixed

api/v1beta2/metal3datatemplate_types.go
```diff
@@ -923,6 +925,7 @@ type Metal3DataTemplateStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:subresource:status
 // +kubebuilder:object:root=true
+// +kubebuilder:metadata:labels="clusterctl.cluster.x-k8s.io/move="
```
kashifest, line 928: do we need to add the label here? seem like the label is missing the ending quote or is it intentional?

baremetal/metal3machine_manager.go
```diff
@@ -1667,54 +1666,29 @@ func (m *MachineManager) SetMetal3DataReadyConditionTrue(reason string) {
 	})
 }
 
-// SetOwnerRef adds an ownerreference to this Metal3Machine.
-func setOwnerRefInList(refList []metav1.OwnerReference, controller bool,
```
kashifest, line 1671: why are we touching metal3Machine ownerRefs?

api/v1beta2/metal3datatemplate_types.go
adilGhaffarDev: This is how we were setting labels on BMH and HardwareData crds so I followed same here. I need to check the CAPI documentation to confirm whether a value is recommended here. But not setting the value does not effect the functionality. 

Some context on why this label is needed: previously, `Metal3DataTemplate` was moved automatically during pivoting via the cluster `OwnerReference` chain. We removed cluster ownerref from `Metal3DataTemplate`, the label must now be set explicitly to ensure the object is included in a pivot.

For the transition strategy, I'm leaning towards setting the label by default for now and deprecating that behaviour in a future release. This gives users time to adapt without breaking existing workflows. The rough plan would be:

- **Now:** set the label by default; add a deprecation notice in the release notes informing users that `Metal3DataTemplate` will no longer have this label set automatically in a future release.
- **Next release:** remove the default label-setting logic; users must set `clusterv1.ClusterNameLabel` on their `Metal3DataTemplate` CRs manually to ensure correct pivoting.

Or should we keep this label forever for the ease of user.

kashifest: we need to check if setting the label by default is the best practice, otherwise users should set them explicitly. You are deprecating the spec.clusterName but already removing the ownerRef so isn't that change of behavior already? shall we do it one step at a time?

baremetal/metal3machine_manager.go
adilGhaffarDev: This is misleading, this was not used to set owner on `m3m`, it was used only one place and there it was being used to set owner on `m3mdt`, here https://github.com/metal3-io/cluster-api-provider-metal3/blob/982445d89f99de338e488b281223ddd4c87c1d9d/baremetal/metal3datatemplate_manager.go#L100

api/v1beta2/metal3datatemplate_types.go
adilGhaffarDev: yes probably, i will check that.

adilGhaffarDev: the ownerRef was being set using spec.clusterName. We can try to set set ownerRef by some other way but that is going to be extra code that we will need remove anyway.

smoshiur1237 (commented):
@Sunnatillo @adilGhaffarDev @lentzi90 @Rozzii 
Code freeze for Release-1.14 will start at the end of this week. Could we take this in beforehand

smoshiur1237 (commented):
/approve

baremetal/metal3datatemplate_manager.go
```diff
@@ -319,8 +289,15 @@ func (m *DataTemplateManager) createData(ctx context.Context,
 
 	m.Log.Info("Index assigned", LogFieldMetal3DataClaim, dataClaim.Name, LogFieldIndex, claimIndex)
 
-	// Create the Metal3Data object, with an Owner ref to the Metal3Machine
-	// (curOwnerRef) and to the Metal3DataTemplate. Also add a finalizer.
+	// Create the Metal3Data object with the Metal3DataClaim as its controlling
+	// owner and the Metal3Machine as an additional owner. The Metal3DataClaim is
+	// deliberately the controlling owner (instead of the Metal3DataTemplate) so
+	// that the Metal3Data participates in the Cluster garbage-collection chain
```
lentzi90, line 295: Good! I think we can also remove the Metal3Machine. It would anyway be in the chain. We can do that as a separate PR though.

controllers/metal3datatemplate_controller.go
```diff
@@ -109,22 +105,12 @@ func (r *Metal3DataTemplateReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-			log.V(baremetal.VerbosityLevelDebug).Info("Error fetching cluster, may not exist yet")
-			log.Info("Error fetching cluster. It might not exist yet, Requeuing")
-			return ctrl.Result{}, nil
-		}
+	// A Metal3DataTemplate is not bound to a single Cluster; multiple Clusters
+	// may reference the same template. Honor a pause requested directly on the
+	// object via the pause annotation instead of deriving an owning Cluster.
+	if annotations.HasPaused(metal3DataTemplate) {
+		log.Info("Metal3DataTemplate is currently paused. Remove pause annotation to continue reconciliation.")
+		return ctrl.Result{Requeue: true, RequeueAfter: requeueAfter}, nil
```
lentzi90, line 113: Does the user have to add the pause annotation before doing `clusterctl move` or is this somehow automated/not needed?

lentzi90 (commented):
Great work!
I have one question about the paused annotation, otherwise this is LGTM.

controllers/metal3datatemplate_controller.go
adilGhaffarDev: Pausing is not automated, and I don't think it's needed here. During `clusterctl move`, the `Metal3Data` and `Metal3DataClaim` objects are effectively paused, so the template's rendered data is not being mutated while the move is in progress. The `Metal3DataTemplate` itself keeps reconciling, but its reconcile won't create new data for claims that are already rendered.

The only way this becomes a problem is if, during the pivot, another cluster starts consuming the `Metal3DataTemplate` that is being moved. But that is already a breaking scenario, initializing a new cluster (or performing other mutating operations) while a `clusterctl move` is in progress will break things even without this PR.

lentzi90: Thanks for checking @adilGhaffarDev!

lentzi90 (commented):
/lgtm

PR #3522, opened 2026-07-01
 🐛 Allow OCI image URLs without checksum

### LINKED ISSUE #3503: OCI Image unusable due to checksum validation conflict with BMO

Specifying an `oci://` URL for `image.url` in the `Metal3MachineTemplate` results in a validation conflict between CAPM3 and BMO, making it unusable. If the `checksum` is left empty, CAPM3 rejects it with the error `Image.Checksum: Required value: cannot be empty`; if a value is provided, BMO rejects it with the error `checksum must be empty for OCI images`. I believe this is likely due to changes introduced in BMO `PR#3186`(https://github.com/metal3-io/baremetal-operator/pull/3186).

I would like CAPM3 to not require a checksum when an `oci://` URL is used, and for the checksum to be omitted when populating the BMH object. 

I believe this can be resolved simply by adding an exception for `oci://` URLs to the checksum validation logic in `Image.Validate` and to the image copying logic (to the BMH) in `metal3machine_manager.go`.

**Environment:**

- Cluster-api version: v1.12.5
- Cluster-api-provider-metal3 version: v1.13.1

/kind bug

daigo-suhara:

OCI URLs (`oci://`) embed the content digest in the URL itself, so a separate `Image.Checksum` is not meaningful. BMO's `BareMetalHost` webhook explicitly forbids it (see [baremetal-operator#3186](https://github.com/metal3-io/baremetal-operator/pull/3186)), but CAPM3 still required a non-empty checksum and silently propagated whatever the user set to the `BareMetalHost`, causing the BMH patch to fail with `checksum must be empty for OCI images`.

This PR:

- Relaxes `Image.Validate()` so that an OCI URL is valid without a checksum, and rejects a non-empty checksum for OCI URLs with a clear error.
- Makes `setHostSpec()` refuse the same combination instead of forwarding a meaningless checksum to the `BareMetalHost`.
- Updates the generated CRDs so that `checksum` is no longer a required field (it remains required at runtime for non-OCI, non-live-iso images via `Image.Validate()`).
- Adds unit tests covering the OCI valid / invalid cases in `api/v1beta2/common_types_test.go`.

Fixes #3503

**Checklist:**

### DIFF
--- a/api/v1beta2/common_types.go
+++ b/api/v1beta2/common_types.go
@@ -98,8 +98,8 @@ type Image struct {
 	URL string `json:"url,omitempty"`
 
 	// checksum is a md5sum, sha256sum or sha512sum value or a URL to retrieve one.
-	// When diskFormat is live-iso, empty string is also a valid value.
-	// +required
+	// Optional for live-iso and oci:// URLs; required otherwise.
+	// +optional
 	// +kubebuilder:validation:MinLength=0
 	// +kubebuilder:validation:MaxLength=512
 	Checksum *string `json:"checksum,omitempty"`
@@ -143,6 +143,11 @@ type Metal3ObjectRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
+// IsOCI reports whether the image URL is an OCI reference.
+func (i *Image) IsOCI() bool {
+	return i != nil && strings.HasPrefix(strings.ToLower(i.URL), "oci://")
+}
+
 // Validate performs validation on [Image], returning a list of field errors using the provided base path.
 // It is intended to be used in the validation webhooks of resources containing [Image].
 func (i *Image) Validate(base field.Path) field.ErrorList {
@@ -161,6 +166,18 @@ func (i *Image) Validate(base field.Path) field.ErrorList {
 			errors = append(errors, field.Invalid(base.Child("URL"), i.URL, "not a valid URL"))
 		}
 	}
+
+	// Checksum is not required for OCI.
+	if i.IsOCI() {
+		if i.Checksum != nil && *i.Checksum != "" {
+			errors = append(errors, field.Forbidden(base.Child("Checksum"), "must be empty for OCI images"))
+		}
+		if i.ChecksumType != "" {
+			errors = append(errors, field.Forbidden(base.Child("ChecksumType"), "must be empty for OCI images"))
+		}
+		return errors
+	}
+
 	// Checksum is not required for live-iso.
 	if i.DiskFormat != LiveISODiskFormat {
 		if i.Checksum == nil || *i.Checksum == "" {
--- a/api/v1beta2/common_types_test.go
+++ b/api/v1beta2/common_types_test.go
@@ -98,6 +98,37 @@ func TestImageValidate(t *testing.T) {
 			ErrorExpected: false,
 			Name:          "Valid spec with live-iso diskFormat",
 		},
+		{
+			Image: Image{
+				URL: "oci://example.com/image:latest",
+			},
+			ErrorExpected: false,
+			Name:          "OCI image without checksum",
+		},
+		{
+			Image: Image{
+				URL:      "oci://example.com/image:latest",
+				Checksum: ptr.To(""),
+			},
+			ErrorExpected: false,
+			Name:          "OCI image with empty checksum",
+		},
+		{
+			Image: Image{
+				URL:      "oci://example.com/image:latest",
+				Checksum: ptr.To("sha256hash"),
+			},
+			ErrorExpected: true,
+			Name:          "OCI image with checksum",
+		},
+		{
+			Image: Image{
+				URL:          "oci://example.com/image:latest",
+				ChecksumType: "sha256",
+			},
+			ErrorExpected: true,
+			Name:          "OCI image with checksumType",
+		},
 	}
 
 	for _, tc := range cases {
--- a/baremetal/metal3machine_manager.go
+++ b/baremetal/metal3machine_manager.go
@@ -1155,15 +1155,11 @@ func (m *MachineManager) setHostSpec(_ context.Context, host *bmov1alpha1.BareMe
 			return fmt.Errorf("Metal3Machine %s has neither Image.URL nor CustomDeploy.Method configured", m.Metal3Machine.Name)
 		}
 
-		checksumType := ""
-		if m.Metal3Machine.Spec.Image.ChecksumType != "" {
-			checksumType = m.Metal3Machine.Spec.Image.ChecksumType
-		}
 		if imageConfigured {
 			host.Spec.Image = &bmov1alpha1.Image{
 				URL:          m.Metal3Machine.Spec.Image.URL,
 				Checksum:     ptr.Deref(m.Metal3Machine.Spec.Image.Checksum, ""),
-				ChecksumType: bmov1alpha1.ChecksumType(checksumType),
+				ChecksumType: bmov1alpha1.ChecksumType(m.Metal3Machine.Spec.Image.ChecksumType),
 				DiskFormat:   &m.Metal3Machine.Spec.Image.DiskFormat,
 			}
 		}
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -65,7 +65,9 @@ type Image struct {
 	URL string `json:"url"`
 
 	// Checksum is a md5sum, sha256sum or sha512sum value or a URL to retrieve one.
-	Checksum string `json:"checksum"`
+	// Optional for live-iso and oci:// URLs; required otherwise.
+	// +optional
+	Checksum string `json:"checksum,omitempty"`
 
 	// ChecksumType is the checksum algorithm for the image.
 	// e.g md5, sha256, sha512
@@ -87,6 +89,11 @@ type CustomDeploy struct {
 	Method string `json:"method"`
 }
 
+// IsOCI reports whether the image URL is an OCI reference.
+func (i *Image) IsOCI() bool {
+	return i != nil && strings.HasPrefix(strings.ToLower(i.URL), "oci://")
+}
+
 // Validate performs validation on [Image], returning a list of field errors using the provided base path.
 // It is intended to be used in the validation webhooks of resources containing [Image].
 func (i *Image) Validate(base field.Path) field.ErrorList {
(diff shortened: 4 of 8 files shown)


api/v1beta2/common_types.go
```diff
@@ -161,6 +161,16 @@ func (i *Image) Validate(base field.Path) field.ErrorList {
 			errors = append(errors, field.Invalid(base.Child("URL"), i.URL, "not a valid URL"))
 		}
 	}
```
mabulgu, line 168: Consider adding a reusable `IsOCI()` method on `Image` instead of inlining this check. The same logic is duplicated in `metal3machine_manager.go`, and BMO already defines an equivalent method. Something like:

```go
func (i *Image) IsOCI() bool {
	return i != nil && strings.HasPrefix(strings.ToLower(i.URL), "oci://")
}
```

Then both the webhook and manager can use `i.IsOCI()` instead of repeating the prefix check.

Also, the same checksum validation issue exists in `api/v1beta1/common_types.go` -- its `Image.Validate()` also requires checksum unconditionally (except for live-iso). Users on v1beta1 would hit the same bug. Please update v1beta1 as well, along with its tests.

api/v1beta2/common_types.go
```diff
@@ -161,6 +161,16 @@ func (i *Image) Validate(base field.Path) field.ErrorList {
...
+
+	// Checksum is not required for OCI.
+	isOCI := strings.HasPrefix(strings.ToLower(i.URL), "oci://")
```
mabulgu: Should CAPM3 reject non-empty checksum for OCI images? BMO already enforces this constraint. Duplicating it here creates tight coupling -- if BMO ever relaxes this rule, CAPM3 would also need an update.

An alternative is to simply skip the checksum requirement for OCI (like we do for `live-iso`) and let BMO handle its own validation. This also makes the webhook behavior consistent: live-iso allows any checksum value (including non-empty), while OCI would forbid it.

Either approach is defensible, but worth discussing.

baremetal/metal3machine_manager.go
```diff
@@ -1160,9 +1160,13 @@ func (m *MachineManager) setHostSpec(_ context.Context, host *bmov1alpha1.BareMe
 			checksumType = m.Metal3Machine.Spec.Image.ChecksumType
 		}
 		if imageConfigured {
+			checksum := ptr.Deref(m.Metal3Machine.Spec.Image.Checksum, "")
+			if strings.HasPrefix(strings.ToLower(m.Metal3Machine.Spec.Image.URL), "oci://") && checksum != "" {
```
mabulgu: Two concerns here:

1. This error should never be reached if the webhook is working correctly -- the `Validate()` webhook already rejects OCI + non-empty checksum. If we keep both, a comment explaining the belt-and-suspenders intent would help.

2. More importantly, even when checksum is empty, the code still sets `Checksum: ""` and `ChecksumType` on the BMH for OCI images. It would be cleaner to omit these fields entirely for OCI:

```go
if imageConfigured {
	host.Spec.Image = &bmov1alpha1.Image{
		URL:        m.Metal3Machine.Spec.Image.URL,
		DiskFormat: &m.Metal3Machine.Spec.Image.DiskFormat,
	}
	if !m.Metal3Machine.Spec.Image.IsOCI() {
		host.Spec.Image.Checksum = ptr.Deref(m.Metal3Machine.Spec.Image.Checksum, "")
		if m.Metal3Machine.Spec.Image.ChecksumType != "" {
			host.Spec.Image.ChecksumType = bmov1alpha1.ChecksumType(m.Metal3Machine.Spec.Image.ChecksumType)
		}
	}
}
```

mabulgu (changes requested):
Thanks for the contribution @daigo-suhara! The overall approach is correct. I have a few suggestions below -- mostly around code reuse, missing v1beta1 coverage, and how checksum is passed to the BMH.

api/v1beta2/common_types.go
daigo-suhara: Indeed. To prevent a recurrence of a bug similar to this one, it would be best to leave it to BMO.

api/v1beta2/common_types.go
daigo-suhara: Done in c9c7246d. Added `IsOCI()` and applied the same fix to `api/v1beta1`

baremetal/metal3machine_manager.go
daigo-suhara: Done in c9c7246d — omitted `Checksum` / `ChecksumType` on the BMH `Image` for OCI images, and dropped the runtime error

mabulgu:
@daigo-suhara You should also DCO sign-off your commits. see here: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3522/checks?check_run_id=84599438085

daigo-suhara:
> @daigo-suhara You should also DCO sign-off your commits. see here: https://github.com/metal3-io/cluster-api-provider-metal3/pull/3522/checks?check_run_id=84599438085

I've amended the commits to include the DCO sign-off and force-filed them.

baremetal/metal3machine_manager.go
```diff
@@ -1155,16 +1155,14 @@ func (m *MachineManager) setHostSpec(_ context.Context, host *bmov1alpha1.BareMe
...
-				ChecksumType: bmov1alpha1.ChecksumType(checksumType),
-				DiskFormat:   &m.Metal3Machine.Spec.Image.DiskFormat,
+				URL:        m.Metal3Machine.Spec.Image.URL,
+				DiskFormat: &m.Metal3Machine.Spec.Image.DiskFormat,
+			}
+			if !m.Metal3Machine.Spec.Image.IsOCI() {
+				host.Spec.Image.Checksum = ptr.Deref(m.Metal3Machine.Spec.Image.Checksum, "")
+				host.Spec.Image.ChecksumType = bmov1alpha1.ChecksumType(m.Metal3Machine.Spec.Image.ChecksumType)
 			}
 		}
```
mabulgu, line 1165: Nit: the old code had a guard `if m.Metal3Machine.Spec.Image.ChecksumType != ""` before setting `ChecksumType`. Without that guard, an empty string gets assigned as `bmov1alpha1.ChecksumType("")`. Not a bug, but it is a needless assignment. Consider restoring the guard:

```go
if !m.Metal3Machine.Spec.Image.IsOCI() {
	host.Spec.Image.Checksum = ptr.Deref(m.Metal3Machine.Spec.Image.Checksum, "")
	if m.Metal3Machine.Spec.Image.ChecksumType != "" {
		host.Spec.Image.ChecksumType = bmov1alpha1.ChecksumType(m.Metal3Machine.Spec.Image.ChecksumType)
	}
}
```

baremetal/metal3machine_manager.go
```diff
@@ -1155,16 +1155,14 @@ func (m *MachineManager) setHostSpec(_ context.Context, host *bmov1alpha1.BareMe
...
-		}
 		if imageConfigured {
 			host.Spec.Image = &bmov1alpha1.Image{
-				URL:          m.Metal3Machine.Spec.Image.URL,
-				Checksum:     ptr.Deref(m.Metal3Machine.Spec.Image.Checksum, ""),
-				ChecksumType: bmov1alpha1.ChecksumType(checksumType),
-				DiskFormat:   &m.Metal3Machine.Spec.Image.DiskFormat,
+				URL:        m.Metal3Machine.Spec.Image.URL,
+				DiskFormat: &m.Metal3Machine.Spec.Image.DiskFormat,
+			}
```
mabulgu: Could you also add a unit test for this `setHostSpec` change? Specifically, a test that verifies an OCI image does not populate `Checksum` or `ChecksumType` on the resulting BMH, and a test that verifies non-OCI images still populate them as before.

mabulgu (commented):
Thanks for addressing the previous review, @daigo-suhara! The changes look much better. Two small items below.

baremetal/metal3machine_manager.go
daigo-suhara: Done in 15ad5839. Restored the `if != ""` guard so `ChecksumType("")` is no longer assigned when unset.

baremetal/metal3machine_manager.go
daigo-suhara: Done in 15ad5839. Added two entries to the `Test SetHostSpec` DescribeTable:
 
  - `OCI image does not populate Checksum or ChecksumType` — verifies the BMH Image only has `URL` and `DiskFormat` for `oci://` URLs.
  - `Non-OCI image populates Checksum and ChecksumType` — verifies the existing behavior with an explicit `ChecksumType: "sha256"`.

mabulgu (approved):
LGTM, thanks for addressing all the feedback @daigo-suhara! The `ChecksumType` guard and the new `setHostSpec` tests look good.

peppi-lotta:
@daigo-suhara Thank you for this fix! This is good work. Excellent reviews from @mabulgu. 
/copilot-review

api/v1beta2/common_types_test.go
```diff
@@ -98,6 +98,29 @@ func TestImageValidate(t *testing.T) {
...
+			ErrorExpected: false,
+			Name:          "Valid OCI URL with empty checksum",
+		},
+		{
+			Image: Image{
+				URL:      "oci://quay.io/example/rhcos:latest",
+				Checksum: ptr.To("f7600f7a274d974a236c4da5161265859c32da93a7c8de6a77d560378a1384ef"),
+			},
+			ErrorExpected: false,
+			Name:          "OCI URL with non-empty checksum is accepted (BMO will reject)",
```
peppi-lotta: `setHostSpec` strips the checksum before sending to BMO does it not? so this comment is not accurate.

api/v1beta2/common_types.go
peppi-lotta: Currently doesn't the `setHostSpec` strip the checksum when image is OCI? CAPM3 now "silently" drops the values and never sends them to BMO.  I think it might be better to have the same constraint in CAPM3 and BMO because now if BMO relaxes the constraints, CAPM3 would require the `setHostSpec` to be changed but  it would be hard to remember/catch the inconsistency because CAPM3 accepts the checksum value and never sends it forward. The flow of these values from CAPM3 to BMO is never tested when image is OCI and tests are only repo specific.

daigo-suhara:
@peppi-lotta Thank you for the review. 
I've reworked the change to keep CAPM3 and BMO aligned:

**Changes**

  - `Image.Validate()` now rejects a non-empty `Checksum` and `ChecksumType` with `field.Forbidden` when the URL is `oci://`.
  - Removed the `if !IsOCI()` branch in `setHostSpec()` and restored the pass-through: `Checksum` and `ChecksumType` are copied to the BareMetalHost as-is.

peppi-lotta:
> @peppi-lotta Thank you for the review. I've reworked the change to keep CAPM3 and BMO aligned:
> 
> **Changes**
> 
>     * `Image.Validate()` now rejects a non-empty `Checksum` and `ChecksumType` with `field.Forbidden` when the URL is `oci://`.
> 
>     * Removed the `if !IsOCI()` branch in `setHostSpec()` and restored the pass-through: `Checksum` and `ChecksumType` are copied to the BareMetalHost as-is.

@daigo-suhara Thank you! I think this is clearer for user because values are kept aligned between the repos. Validation gives fast failure so user is fast to mitigate. I'm happy to give this an LGTM but before that please squash your commits  into one.

daigo-suhara:
> > @peppi-lotta Thank you for the review. I've reworked the change to keep CAPM3 and BMO aligned:
> > **Changes**
> > ```
> > * `Image.Validate()` now rejects a non-empty `Checksum` and `ChecksumType` with `field.Forbidden` when the URL is `oci://`.
> > 
> > * Removed the `if !IsOCI()` branch in `setHostSpec()` and restored the pass-through: `Checksum` and `ChecksumType` are copied to the BareMetalHost as-is.
> > ```
> 
> @daigo-suhara Thank you! I think this is clearer for user because values are kept aligned between the repos. Validation gives fast failure so user is fast to mitigate. I'm happy to give this an LGTM but before that please squash your commits into one.

@peppi-lotta Squashed into a single commit (`d075fd5b`) — force-pushed.

adilGhaffarDev (approved):
/approve

PR #3583, opened 2026-07-30
🌱 Harden hack scripts against WORKDIR volume mount injection

migiFi:
Affects the following scripts under the hack/ directory:

- codegen.sh
- gomod.sh
- unit.sh
- manifestlint.sh
- markdownlint.sh
- shellcheck.sh
- build.sh

These scripts read WORKDIR from the env and insert it directly into the container's --volume mount specification. Because the volume spec is colon-delimited `(host:target:options)`, a WORKDIR value containing a colon could alter how the mount is parsed.

This change removes the WORKDIR var and hardcodes the container path to /workdir, which matches the pattern already used in several other metal3 repos. Since /workdir was already the default, there is no change in behavior for normal use, but the injection surface is eliminated.

**NOTE**: codegen.sh keeps its `rw, z` mount because it runs `make generate` in place, so only the path is changed there, not the mount mode.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -5,7 +5,6 @@ set -eux
 
 IS_CONTAINER="${IS_CONTAINER:-false}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
-WORKDIR="${WORKDIR:-/workdir}"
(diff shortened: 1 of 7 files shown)


Rozzii (commented):
/lgtm

smoshiur1237 (commented):
/approve

PR #3595, opened 2026-08-04
🌱 Add warning about setting high values for retrylimit

Sunnatillo:
Document that a large spec.strategy.retryLimit keeps the host in a reboot loop for a long time, and regenerate the CRDs.

Fixes #

**Checklist:**

### DIFF
--- a/api/v1beta2/metal3remediation_types.go
+++ b/api/v1beta2/metal3remediation_types.go
@@ -62,6 +62,10 @@ type RemediationStrategy struct {
 
 	// retryLimit sets maximum number of remediation retries.
 	// retryLimit must be a positive integer. The default is 1.
+	// Setting a very high value is discouraged: the controller waits at
+	// least timeoutSeconds (minimum 100s) between attempts, so a large
+	// retryLimit keeps the host in a reboot loop for a very long time
+	// before remediation is allowed to fail.
 	// +optional
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:default=1
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_metal3remediations.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_metal3remediations.yaml
@@ -161,6 +161,10 @@ spec:
                     description: |-
                       retryLimit sets maximum number of remediation retries.
                       retryLimit must be a positive integer. The default is 1.
+                      Setting a very high value is discouraged: the controller waits at
+                      least timeoutSeconds (minimum 100s) between attempts, so a large
+                      retryLimit keeps the host in a reboot loop for a very long time
+                      before remediation is allowed to fail.
                     format: int32
                     minimum: 1
                     type: integer
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_metal3remediationtemplates.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_metal3remediationtemplates.yaml
@@ -152,6 +152,10 @@ spec:
                             description: |-
                               retryLimit sets maximum number of remediation retries.
                               retryLimit must be a positive integer. The default is 1.
+                              Setting a very high value is discouraged: the controller waits at
+                              least timeoutSeconds (minimum 100s) between attempts, so a large
+                              retryLimit keeps the host in a reboot loop for a very long time
+                              before remediation is allowed to fail.
                             format: int32
                             minimum: 1
                             type: integer


Sunnatillo:

Documentation change, one testing is enough

smoshiur1237 (commented):
/hold
is it a duplicate of #3596

api/v1beta2/metal3remediation_types.go
```diff
@@ -62,6 +62,10 @@ type RemediationStrategy struct {
 
 	// retryLimit sets maximum number of remediation retries.
 	// retryLimit must be a positive integer. The default is 1.
+	// Setting a very high value is discouraged: the controller waits at
+	// least timeoutSeconds (minimum 100s) between attempts, so a large
+	// retryLimit keeps the host in a reboot loop for a very long time
+	// before remediation is allowed to fail.
```
fmuyassarov, line 68: A go doc comment only helps devs reading the source. If the intent is to signal to users, shouldn't we rather add a warning via the existing validating webhook instead for example?

example
```Go
if newM3rt.Spec.Template.Spec.Strategy.RetryLimit > highRetryLimitWarningThreshold {
		timeout := newM3rt.Spec.Template.Spec.Strategy.TimeoutSeconds
		if timeout == 0 {
				timeout = defaultTimeout
		}
		warnings = append(warnings, fmt.Sprintf(
				"spec.template.spec.strategy.retryLimit=%d is very high: the controller waits at least timeoutSeconds=%d "+
						"between attempts, so the host may stay in a reboot loop for a long time before remediation is allowed to fail",
				newM3rt.Spec.Template.Spec.Strategy.RetryLimit, timeout,
		))
}
```

Rozzii (commented):
I am conflicted if we need this change either as a comment or as a log message.
I think the spec field description of the CR would be the correct place for this information.

fmuyassarov:
> I am conflicted if we need this change either as a comment or as a log message.
> 
> I think the spec field description of the CR would be the correct place for this information.
> 
> 
> 
> 

or as enforcement https://github.com/metal3-io/cluster-api-provider-metal3/pull/3595#discussion_r3718546238

api/v1beta2/metal3remediation_types.go
Sunnatillo: CRD description is generated from this comments, it is serving its purpose correctly. 

For webhook validation we could not agree what to set threshold. Is it 10 or is it 100? We do not know. It depends on the use case.

Sunnatillo:
> /hold is it a duplicate of #3596

it is not duplicate. 
/unhold

Sunnatillo:
> I am conflicted if we need this change either as a comment or as a log message. I think the spec field description of the CR would be the correct place for this information.

it is already in CRD description

api/v1beta2/metal3remediation_types.go
fmuyassarov: > CRD description is generated from this comments, it is serving its purpose correctly.

If your intent is to really give a singal to user, then I doubt that consumers would read the CRD description. 
If your intent is just to keep this for devs, well then maybe yes. But I really doubt this will be helpful to the people who are deploying it on their clusters. It just gets buried.

Rozzii (commented):
/approve

Rozzii:
ahh I have only lgtm rights here
/lgtm

api/v1beta2/metal3remediation_types.go
Sunnatillo: We decided to keep it like this, this does not worth to do webhook changes. We cannot decide which value too high, this is up to user.

fmuyassarov: What's the benefit then?

fmuyassarov: You can decide however you prefer, but I believe that this won't help or benefit the user in this form. It will just get buried.

fmuyassarov: To be clear, I am not blocking it. I just want to understand why are we doing it or for whom. If the intention is to help user from ending up in unexpected situation then I don't think this is visible to them. But if the intent is for the sake of documentation I don't issue issues here

Sunnatillo:
@adilGhaffarDev @fmuyassarov I made the changes and answered the questions. PTAL again.

adilGhaffarDev:
I am ok with this for now. If we decide to some high value for which we can show warning, that can done in separate PR.
But documentation purposes this is ok.
/approve

Sunnatillo:
/retitle 🌱 Add warning about setting high values for retrylimit

PR #3597, opened 2026-08-04
🌱 Block syncing of reserved kubernetes.io labels

Sunnatillo:
Filter out labels under the reserved kubernetes.io and k8s.io domains (and their subdomains) in buildLabelSyncSet, so they are never synced from a BareMetalHost to a Node even if an admin lists such a prefix in the label-sync annotation. Prevents scheduling issues from labels like kubernetes.io/os or kubernetes.io/arch.

Fixes #

**Checklist:**

### DIFF
--- a/baremetal/labelsync.go
+++ b/baremetal/labelsync.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2025 The Metal3 Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package baremetal
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+const (
+	// PrefixAnnotationKey is the annotation key on Metal3Cluster that lists the label prefixes to sync.
+	PrefixAnnotationKey = "metal3.io/metal3-label-sync-prefixes"
+)
+
+// reservedLabelDomains are Kubernetes-reserved label domains that must never be used as sync prefixes.
+var reservedLabelDomains = []string{
+	"kubernetes.io",
+	"k8s.io",
+}
+
+// IsReservedLabelPrefix reports whether the given label prefix belongs to a
+// Kubernetes-reserved domain (kubernetes.io, k8s.io) or any of their subdomains.
+func IsReservedLabelPrefix(prefix string) bool {
+	for _, domain := range reservedLabelDomains {
+		if prefix == domain || strings.HasSuffix(prefix, "."+domain) {
+			return true
+		}
+	}
+	return false
+}
+
+// ParsePrefixAnnotation parses a comma-separated list of DNS-1123 subdomain prefixes.
+func ParsePrefixAnnotation(prefixStr string) (map[string]struct{}, error) {
+	entries := strings.Split(prefixStr, ",")
+	prefixSet := make(map[string]struct{})
+	for _, prefix := range entries {
+		prefix = strings.TrimSpace(prefix)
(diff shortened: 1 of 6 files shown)


smoshiur1237 (commented):
/approve

fmuyassarov:
@Sunnatillo would you mind to elaborate what's the issue of inheriting those labels?

fmuyassarov:
> @Sunnatillo would you mind to elaborate what's the issue of inheriting those labels?

Based on what we discussed during the offline call, it sounds like there are some security concerns around preventing unintended labels from being set on the cluster.

My preference would be to first explore whether it's possible to filter out the labels within BMO rather than handling it here. If BMO needs to rely on those labels when it's used without CAPM3, I'd be interested in understanding that use case better. My understanding is that these labels aren't intended to be set by users or admins, regardless of whether BMO is used with or without CAPM3.

fmuyassarov:
> > @Sunnatillo would you mind to elaborate what's the issue of inheriting those labels?
> 
> Based on what we discussed during the offline call, it sounds like there are some security concerns around preventing unintended labels from being set on the cluster.
> 
> My preference would be to first explore whether it's possible to filter out the labels within BMO rather than handling it here. If BMO needs to rely on those labels when it's used without CAPM3, I'd be interested in understanding that use case better. My understanding is that these labels aren't intended to be set by users or admins, regardless of whether BMO is used with or without CAPM3.

It seems I misread the description. The labels are actually passed from CAPM3 to BMO, not the other way around. In that case, it makes sense to handle the filtering here. What initially threw me off was that the description made me think we were blocking labels in BMO, whereas we're actually preventing them from being propagated by CAPM3, which is the correct place to do it.

controllers/metal3labelsync_controller.go
adilGhaffarDev, line 222: maybe we can add some restriction for "prefix set" in webhooks?
dont allow user to set "prefix set" that k8s related labels?

Rozzii (commented):
/hold until both the AI and @adilGhaffarDev 's comment have been addressed. Otherwise looks good.

adilGhaffarDev:
@Sunnatillo can you address the comments?

smoshiur1237:
@Sunnatillo address the comments and update. We would like to take this in.

controllers/metal3labelsync_controller.go
Sunnatillo: That would be the best approach. Thank you for suggesting that.

controllers/metal3labelsync_controller.go
```diff
@@ -303,6 +326,11 @@ func buildLabelSyncSet(prefixSet map[string]struct{}, labels map[string]string)
 		if _, ok := prefixSet[p]; !ok {
 			continue
 		}
+		// Never sync labels under Kubernetes-reserved domains, even if an
+		// admin explicitly added such a prefix to the sync annotation.
+		if isReservedLabelPrefix(p) {
+			continue
```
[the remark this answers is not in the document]
Sunnatillo: not relevant anymore

Sunnatillo:
In the current implementation webhook does not allow kubernetes.io and k8s.io labels

Sunnatillo:
@Rozzii Could you unhold this PR? Comments have been addressed

baremetal/labelsync.go
```diff
@@ -0,0 +1,94 @@
...
+	return nil
+}
+
+// regexError returns a string explanation of a regex validation failure.
+func regexError(msg string, format string, examples ...string) string {
+	if len(examples) == 0 {
+		return msg + " (regex used for validation is '" + format + "')"
+	}
+	return msg + " (e.g. '" + strings.Join(examples, "' or '") + "', regex used for validation is '" + format + "')"
+}
```
Rozzii: As I understood, it is a standard practice to parse / filter / block these labels related to k8s reserved domains, so isn't there some K8s go library that we could import instead of implementing the pattern matching and format filtering from scratch ?

Rozzii (commented):
In general I am okay with this but it seems odd that we have to write this labelsync filter logic from scratch .

baremetal/labelsync.go
Sunnatillo: Agree,
DNS-1123 part is now imported from "k8s.io/apimachinery/pkg/util/validation"

Rozzii (commented):
/lgtm

adilGhaffarDev (commented):
/lgtm

PR #3598, opened 2026-08-04
 ⚠️  Add length validation to v1beta1 Metal3Machine string fields

Sunnatillo:
v1beta1 omitted length limits that v1beta2 already enforces. Since v1beta1 stays served, add maxLength bounds and regenerate CRDs:
- providerID:          maxLength 512
- failureDomain:       maxLength 256
- customDeploy.method: minLength 1, maxLength 512

minLength is applied only to customDeploy.method, where an empty value is meaningless; failureDomain and providerID keep empty as a valid (unset) state to avoid breaking existing objects.

Fixes #

**Checklist:**

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -84,6 +84,8 @@ type CustomDeploy struct {
 	// Custom deploy method name.
 	// This name is specific to the deploy ramdisk used. If you don't have
 	// a custom deploy ramdisk, you shouldn't use CustomDeploy.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
 	Method string `json:"method"`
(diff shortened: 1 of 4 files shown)


adilGhaffarDev:
I dont think we need to add this in v1beta1.

adilGhaffarDev:
> I dont think we need to add this in v1beta1.

We will be removing v1beta1 soon anyway and this is not something critical that we need to fix.

Rozzii (commented):
Strongly agree with @adilGhaffarDev , we don't need this as v1beta1 is being phased out and this change would just complicate backward compatibility at this point. Cancel the PR please @Sunnatillo  if you agree with us.
/hold

Sunnatillo:
Agreed

PR #3603, opened 2026-08-10
🐛 Fix Metal3DataTemplate racing with clusterctl move

adilGhaffarDev:
After #3520 the Metal3DataTemplate controller no longer derives its pause state from the owning Cluster; it only checks its own paused annotation. During `clusterctl move` the Cluster is paused but the template is not, so on the target cluster the controller keeps reconciling and re-creates Metal3Data objects with a freshly-computed claim/index mapping. When clusterctl then moves the original Metal3Data, the update is rejected by the webhook because spec.claim and spec.index are immutable ("cannot be modified"), which breaks (re)pivoting.

Skip rendering a claim whose owning Cluster is paused in UpdateDatas, mirroring the util.GetClusterFromMetadata + annotations.IsPaused pattern already used by the Metal3Data controller. This restores the pre-#3520 freeze during clusterctl move while keeping the template decoupled from a single Cluster (the check is per-claim via the ClusterNameLabel).

Added unit test verifying no Metal3Data is created for a claim whose Cluster is paused.

Fixes #

**Checklist:**

### DIFF
--- a/baremetal/metal3datatemplate_manager.go
+++ b/baremetal/metal3datatemplate_manager.go
@@ -31,6 +31,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/patch"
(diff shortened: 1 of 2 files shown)


smoshiur1237 (commented):
/approve

lentzi90 (commented):
/lgtm

Rozzii (commented):
/lgtm

adilGhaffarDev:
ipa flake running test again.
/retest

PR #3636, opened 2026-08-24
✨Add RootFSUUID support for whole-disk images with software RAID

mnasiadka:

Propagates BareMetalHost.spec.image.rootFSUUID (added in https://github.com/metal3-io/baremetal-operator/pull/3554) through Metal3MachineTemplate/Metal3Machine, so it can be set declaratively on CAPI-managed hosts instead of only on directly-managed BareMetalHosts.
Ironic requires this field when deploying a whole-disk image onto a host with software RAID configured, to locate the root filesystem inside the assembled RAID array and reinstall a RAID-aware bootloader.

Adds RootFSUUID to both the v1beta1 and v1beta2 Image types (with manual conversion between them, matching the existing pattern for Checksum/ChecksumType/DiskFormat), and wires it into metal3machine_manager.go where Metal3Machine.Spec.Image already gets copied into BareMetalHost.Spec.Image.

**In draft mode and using go.mod from my BMO fork until 3554 merges.**

**Checklist:**

- [X] Documentation has been updated, if necessary.
- [X] Unit tests have been added, if necessary.
- [X] E2E tests have been added, if necessary.
- [X] Integration tests have been added, if necessary.


adilGhaffarDev:
Please this PR to draft until this https://github.com/metal3-io/baremetal-operator/pull/3554 is merged.
Few things:
- No need to add this in v1beta1
- Sign your commits to satisfy DCO

PR #3687, opened 2026-09-04
✨ Pass ociAuthSecretName through to the BareMetalHost

akselleirv:
BMO gained per-host OCI registry credentials in metal3-io/baremetal-operator#2745: BareMetalHost.spec.image.ociAuthSecretName names a Docker-config secret, which BMO resolves and forwards to Ironic as instance_info[image_pull_secret].

CAPM3 owns spec.image on a consumed host, and setHostSpec only writes it while it is nil, so there is no way for a user to set the field: a value placed on the BareMetalHost by hand is dropped the next time the host is deprovisioned and reprovisioned, and pre-setting spec.image to carry it makes setHostSpec skip the whole block, including UserData, MetaData and NetworkData.

Add OCIAuthSecretName to the CAPM3 Image type in v1beta1 and v1beta2 and copy it in setHostSpec, so it can be declared on a Metal3MachineTemplate like the other image fields. Image.Validate rejects it for non-OCI URLs, matching the existing treatment of Checksum and ChecksumType.

**Checklist:**


api/v1beta1/common_types.go
```diff
@@ -79,6 +79,13 @@ type Image struct {
 	// +kubebuilder:validation:Enum=raw;qcow2;vdi;vmdk;live-iso
 	// +optional
 	DiskFormat *string `json:"format,omitempty"`
+
+	// OCIAuthSecretName optionally names a Docker-config secret containing
+	// registry credentials for oci:// images. Must be in the same namespace
+	// as the BareMetalHost. Allowed types: kubernetes.io/dockerconfigjson|dockercfg.
+	// Only used when URL has the oci:// scheme.
+	// +optional
+	OCIAuthSecretName *string `json:"ociAuthSecretName,omitempty"`
```
adilGhaffarDev: no need to add it in v1beta1

adilGhaffarDev (commented):
thanks for the PR.
we dont need to add this in v1beta1, we will remove v1beta1 in next cycle.

PR #3747, opened 2026-09-15
🌱 Build locally from source for ci and PR tests

adilGhaffarDev:
This will make sure to build capm3 container images locally.
This was missed in vbmctl PR.

Fixes #

**Checklist:**


lentzi90 (commented):
I think we can simplify this quite a lot, but let's take it in as is first before we break something.
/approve

test/e2e/data/infrastructure-metal3/overlays/main/kustomization.yaml
```diff
@@ -4,6 +4,10 @@ kind: Kustomization
 resources:
 - ../../../../../../config/default
 
+# The CAPM3 image is injected at e2e runtime by scripts/ci-e2e.sh
+# (kustomize edit set image -> ${REGISTRY}/localimages/cluster-api-provider-metal3:e2e),
+# so the run tests the code in this tree. Standalone kustomize build keeps the
+# config/default :main image.
```
lentzi90, line 10: Is there a reason to do it this way? I think it would be safer and easier to understand if we just hard-code the image for e2e here.

hack/build-e2e-image.sh
```diff
@@ -0,0 +1,51 @@
...
+
+REPO_ROOT=$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/..)
+
+PROVISIONING_URL_HOST="${PROVISIONING_URL_HOST:-172.22.0.1}"
+REGISTRY_PORT="${REGISTRY_PORT:-5000}"
+REGISTRY="${REGISTRY:-${PROVISIONING_URL_HOST}:${REGISTRY_PORT}}"
+E2E_TAG="${E2E_TAG:-e2e}"
+REGISTRY_PORT="${REGISTRY##*:}"
+CAPM3_LOCAL_REPO="${REGISTRY}/localimages/cluster-api-provider-metal3"
+CAPM3_E2E_IMAGE="${CAPM3_E2E_IMAGE:-${CAPM3_LOCAL_REPO}:${E2E_TAG}}"
```
lentzi90, line 31: I think we could basically get rid of all these variables. We do not need to change the port, certainly not the tag. The CAPM3_LOCAL_REPO is also not necessary.

Sunnatillo (commented):
LGTM

PR #3748, opened 2026-09-16
🌱 Simplify e2e image build script

adilGhaffarDev:
Removed unnecessary variables and minor comment improvement.

Fixes #

**Checklist:**


hack/build-e2e-image.sh
```diff
@@ -41,11 +36,10 @@ if [[ -z "$(docker ps -q -f name="^registry$")" ]]; then
 fi
 
 make -C "${REPO_ROOT}" docker-build-e2e \
```
nuhakala, line 38: This make target feels little unnecessary if running tests executes this script anyway? And what would be a case where user wants to build the e2e image without running the tests? Or is it just good practice to include it in the makefile?

